PR #33, opened 2014-11-27
[CXF-6119] WS-Transfer Implementation

dudaerich:
Document, where I discuss implementation issues: https://developer.jboss.org/docs/DOC-52429?uniqueTitle=false
Example program: https://github.com/dudaerich/WS-TransferExample

I have also implemented WS-Fragment, because it is required by WS-Transfer. Except in a few special cases, which must be discussed, the implementation is complete. In tests I cover examples given in the specification, which proves correctness of the implementation for these cases.
Document, where I discuss unclear parts in the specification: https://developer.jboss.org/docs/DOC-52941?uniqueTitle=false


asoldano:
I've had a brief chat with Dan Kulp on this and currently the plan is to merge on master as soon as it's move to 3.2, which should happen just after 3.1.2 release.

tomitribe-dev:

PR #43, opened 2014-12-17
Fix for [CXF-6161]

escabe:
Instead of using a local filename based on the original URL of an
wsdl:import use a numbered name: importNN.wsdl. Similar to schema
imports. This should resolve issue [CXF-6161] in which an invalid
filename may be generated.


tomitribe-dev:

PR #55, opened 2015-03-01
CXF-6268 Make cxf-codegen-plugin toolchains aware during maven build

kops:
Support toolchains configuration during java code generation.

When _fork_ parameter is enabled, 
1. use provided javaExecutable path if available
2. use toolchains configuration if available
3. use default `${java.home}/bin/java`

I added a maven-invoker-plugin to run a simple test using toolchains.


dkulp:
This patch needs to be redone based on master.   2.6.x is unsupported.   All patches flow from master -> 3.0.x -> 2.7.x and thus needs to apply cleanly against master.

kops:
I will review this patch and submit a new pull request.

PR #59, opened 2015-03-19
Store the protocol headers if they need initialized

skitt:
Currently, if SamlHeaderOutInterceptor finds no protocol headers in the message, it still proceeds to create the assertion, but it stores it in a new map which is not stored anywhere. The assertion doesn't appear in the HTTP headers output by CXF...

Simply storing the created map as the message's protocol headers fixes this.


sberyozkin:
Patch has been applied, thanks

PR #64, opened 2015-04-10
CXF-6343 Support and parse EncryptedHeader

spark404:
SOAP Message Security 1.1 paragraph 9.3 mentions that EncryptedHeader is recommended to be used when a header part is encrypted. Recent .NET implementations do this and for interop CXF needs to be able to handle this. WSS4J already does, this fix will make CXF pass the right information on to WSS4J


coheigea:
Did you see my comment on the JIRA?

Colm.

PR #66, opened 2015-04-16
CXF-6343 Support and parse EncryptedHeader

spark404:
Pull request for the master branch.


spark404:
Thanks for merging @coheigea 

I don't know the policy in CXF but in general is good to keep the original author the author of the commit to trace contributions with regards to the Apache License. And of course it is more inviting for new contributors when they see their commits popup in the code base.

coheigea:
I just downloaded the patch and applied it via "git apply". I wasn't aware that this removed the author. I guess I should have used "git am" instead...apologies.

spark404:
@coheigea No worries :-)   The important part is the fix and i'm happy it's in!

PR #73, opened 2015-05-23
[CXF-6421] Slim Exchange map down

asoldano:
- Avoid storing Bus, Service, Endpoint, Binding, BindingOperationInfo into Exchange's map as there're direct accessors for them

- Avoid storing ServiceInfo, InterfaceInfo, BindingInfo into Exchange's map as they're easily retrieved through other stuff in the map
- misc minor cleanups


asoldano:
This is not going to be merged (too much changes for 3.0.x). See https://github.com/apache/cxf/pull/74 for changes to master (3.1.x).

PR #81, opened 2015-07-09
Fixed CXF-6492 - Don't assume only one empty space after 'Basic' in A…

sagara-gunathunga:
…uthorization header


sberyozkin:
Hi - did not see your pull request before applying my own fix - hope you are ok with the fix applied, thanks

PR #84, opened 2015-09-04
Implement optional hostname canonicalization for kerberos (SPNEGO)

dmansfield:
This commit implements an additional property,
useCanonicalHostname, which is supported by the AbstractSpnegoAuthSupplier,
and therefore available to SpnegoAuthSupplier and
KerberosAuthOutInterceptor.

The default behavior is not modified.  If useCanonicalHostname is set,
then the service principal will be created using the canonical name of the
remote host.

For example, if there is a CNAME DNS record for host.example.com which
refers to an A record host-a-b-c.example.com, the default behavior would
be to generate a service principal HTTP/host.example.com. When
useCanonicalHostname is set to true, the service principal would be HTTP/
host-a-b-c.example.com. Most browsers and many other HTTP user agents
perform this kind of canonicalization.

### DIFF
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/auth/AbstractSpnegoAuthSupplier.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/auth/AbstractSpnegoAuthSupplier.java
@@ -18,6 +18,7 @@
  */
 package org.apache.cxf.transport.http.auth;
 
+import java.net.InetAddress;
 import java.net.URI;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
@@ -62,6 +63,7 @@ public abstract class AbstractSpnegoAuthSupplier {
     private boolean credDelegation;
     private Configuration loginConfig;
     private Oid serviceNameType;
+    private boolean useCanonicalHostname;
     
     public String getAuthorization(AuthorizationPolicy authPolicy,
                                    URI currentURI,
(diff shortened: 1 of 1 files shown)


dmansfield:
This patch was discussed briefly on the list back in May.  Subject at that time was "[PATCH] canonicalized host name for kerberos (SPNEGO)".

dmansfield:
There should probably be a documentation patch as well.  Can you point me to it?  Basically this:

When determining the service principal name to be used for kerberos / SPNEGO, hostname canonicalization can be configured in XML as follows:
```
     <bean id="myAuthSupplier"
              class="org.apache.cxf.transport.http.auth.SpnegoAuthSupplier">
         <property name="useCanonicalHostname" value="true" />
     </bean>
     <http-conf:conduit name="...">
         <http-conf:authorization>
             <sec:AuthorizationType>Negotiate</sec:AuthorizationType>
             <sec:Authorization>CXFClient</sec:Authorization>
         </http-conf:authorization>
         <http-conf:authSupplier bean="myAuthSupplier" />
     </http-conf:conduit>
```

Or similarly for KerberosAuthOutInterceptor.

coheigea:
Patch applied, thanks. Documentation can be added here: http://cxf.apache.org/docs/jaxrs-kerberos.html

Colm.

dmansfield:
I don't have permission to edit that page.  My login on that system is apache [at] dm.cobite.com.

What I would do is change http://cxf.apache.org/docs/jaxrs-kerberos.html#JAXRSKerberos-Configuringtheserviceprincipalname by adding a new third sentence:

When the "servicePrincipalName" is not specified, the target host from the provided endpoint URL is used to construct one as-is.  To perform canonicalization of this hostname (e.g. if a CNAME record host.example.com points to an A record host-x.example.com, then use "host-x.example.com" when constructing the servicePrincipalName), the "useCanonicalHostname" property can be set to "true".

coheigea:
This is now done - thanks for your input.

PR #98, opened 2015-10-26
Cxf 6621 integration

rsearls:
integration of changes from cxf-2.7.x-fixes for cxf-6621


rsearls:
submitted wrong branch

PR #101, opened 2015-12-11
Tentative pom to create an OBR index for CXF runtime

mrulli:
This is just tentative and I would like to collect some feedbacks for further improvements. In particular I don't like the fact that so many dependencies must be specified (and maintained) by hand in the indexer pom...


cschneider:
I added a repository some time ago. Do you think we can close this PR?

mrulli:
Yes,  thanks!

Matteo
Il 19/apr/2016 01:31 PM, "Christian Schneider" <notifications@github.com>
ha scritto:

cschneider:
Hi Matteo. You need to close the PR yourself. I could only do it through a commit message.

PR #102, opened 2015-12-12
CXF-6706 hibernate help

benson-basis:
I'm hoping to get some collaborative attention via this PR, as to why the test I added (and @Ignored) fails so spectacularly.


reta:
@benson-basis I'll try to help you out, @cschneider may have better idea about the issue.

reta:
@benson-basis Interesting, I pulled out your PR / branch, run tests and all of them are passed:

test

Any clues?

benson-basis:
Oh, it works. it's just really ugly :-)

Look at the horrible proxy code.

cschneider:
Why do you use the hibernate bundle from servicemix?
It looks like hibernate already provides a validator bundle:
http://search.maven.org/#artifactdetails|org.hibernate|hibernate-validator-osgi|5.2.2.Final|pom

benson-basis:
I didn't know. I assumed that the SM stuff was there for a reason. That
will be my next port of call.

benson-basis:
Do you think I can then avoid the 'specs' bundle, too?

benson-basis:
I think that what's here now should be digestible. No 'helper', no proxy, just a bit of TCCL juggling and import management required by the creator of the BeanValidationProvider instance. People who want to do that from Blueprint will be sad, I guess. Do people want some sort of factory?

If something better emerges from Hibernate later, great.

cschneider:
I think we should not create our own feature for hibernate. There is already one:

benson-basis:
Yes, and it's busted. It uses wrap, and karaf 4 refuses to start up if you
provision it as a bootFeature. For a 3.1.x change, I doubt that we want to
remove the feature. I have unwould the extra feature I added.

cschneider:
Ah yes I see there are two wrap entries ... still I think this does not belong into cxf. 
I am especially concerned with the tight binding of the cxf validation to hibernate validator. I think CXF should be able to work with any validation provider and let the user choose which to install.

How about adding a correct hibernate 5 feature without wrap into the upcoming karaf 4.0.4?
I will take care of providing a fix to hibernate so their own feature is better in the next release.

I propose we only have the cxf-bean-validation feature in CXF. We can load the validation api using dependency=true. This way it should nicely interoperate with the hibernate validator feature.

benson-basis:
I am entirely with you for 3.2.0. For 3.1.5 I just wanted to patch things
into working condition. What do you think of my smaller maneuver in
beanvalidationprovider?

cschneider:
The change in BeanValidator looks good to me. I think it is just a workaround though. If hibernate needs el to work internally it should take care of importing it by itself. Maybe we should file this as a bug.
On the other hand providing a special classloader makes sense anyway as hibernate validator needs access to the user classes.

cschneider:
So anyway .. +1 for this change in CXF 3.1.x. I sometimes concentrate too much on the long term effects. For the bugfix release your changes make sense.

tomitribe-dev:

tomitribe-dev:
Build started sha1 is original commit.

tomitribe-dev:
Build finished. 697 tests run, 2 skipped, 0 failed.

deki:
CXF-6706 is resolved so this should be closed as well, shouldn't it?

PR #105, opened 2016-01-07
[CXF-6738] Replace synchronized blocks with DCL to reduce contention

asoldano:
Sending this as PR for getting explicit review. Please note I had to add a //NOPMD comment to please the PMD checks (which keep on saying that DCL is broken on Java, which is not exactly the case nowadays...)
The DCL idiom here works because the AbstractPropertiesHolder's propertyMap is an AtomicReference, whose internal volatile value should always grant the happens-first relationship.


asoldano:
I've applied the changes directly, closing the PR.

PR #111, opened 2016-02-19
enhancing readability #CXF-5688

azitabh:


sberyozkin:
Thanks for the patch, however I believe I clarified why the code is written the way it is now in https://issues.apache.org/jira/secure/EditComment!default.jspa?id=12708243&commentId=14163245

PR #112, opened 2016-02-20
[CXF-6792] Some STS tests fails locally and on CI server

oscerd:
Hi all,

This PR is related to:
https://issues.apache.org/jira/browse/CXF-6792

Andrea


oscerd:
The errors in the build seems to be sporadic.

PR #119, opened 2016-03-09
Avoid lock on read (issue: https://issues.apache.org/jira/browse/CXF-6825)

yotammadem:
This is seems to be a bottle neck in our performance tests

### DIFF
--- a/core/src/main/java/org/apache/cxf/BusFactory.java
+++ b/core/src/main/java/org/apache/cxf/BusFactory.java
@@ -113,19 +113,19 @@ public abstract class BusFactory {
             return defaultBus;
         }
     }
-    
+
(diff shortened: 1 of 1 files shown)


dkulp:
You HAVE to have the lock on the get as well as the put.   During the put, the entire map is allowed to be restructured internally (buckets recalculated, etc... if the map grows) which would make anything going on in the get invalid.

yotammadem:
So what do you suggest to do?
Maybe we should use some other data structure (one of the concurrent ones)?

yotammadem:
What do you mean by invalid?
I understand that it can return null because it didn't find the thread in the map
but do you think it is really possible for it to return something else?

(in the case of null my code is treating it well)

yotammadem:
I am closing this, will open other one when fix is ready

PR #120, opened 2016-03-09
Avoid lock on read (issue: https://issues.apache.org/jira/browse/CXF-6825)

yotammadem:
Seems to be a performance bottle neck in out tests


yotammadem:
Closing this - will open another one when fix is ready

PR #124, opened 2016-03-31
 - implemented AutoCloseable Client

bilak:
Implemented AutoCloseable interface which simple calls destroy() method

### DIFF
--- a/core/src/main/java/org/apache/cxf/endpoint/Client.java
+++ b/core/src/main/java/org/apache/cxf/endpoint/Client.java
@@ -31,7 +31,7 @@ import org.apache.cxf.service.model.BindingOperationInfo;
 import org.apache.cxf.transport.Conduit;
 import org.apache.cxf.transport.MessageObserver;
 
-public interface Client extends InterceptorProvider, MessageObserver, ConduitSelectorHolder {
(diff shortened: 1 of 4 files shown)


cschneider:
This code does not seem to build. I got somee checkstyle errors.
I also think setting the classloader on the bus is not correct as the bus may be for a complete
bundle. So setting it when a DynamicClient is created and removing it when it is removed does not seem to be valid for all cases.

Pelase also try to not reformat the code in commits that also change the logic as this makes it a lot harder to review.

Ideally you should create one branch / pull request per jira issue and make sure the code builds on your machine using mvn clean install.

bilak:
Yes I've reformatted the code because I didn't knew about your formatter (then I found it). So probably you can fix the issue better if you know how. Sorry for bundling all commits to one PR, I'm not using git too long.

cschneider:
No problem. I had also had a lot of issue with git when I started. I just updated the getting involved guide to show how you ideally structure your work in git. https://cxf.apache.org/getting-involved.html

It does not show the git commands you need but how it should look like.

tomitribe-dev:

tomitribe-dev:
Build started sha1 is merged.

tomitribe-dev:
Build finished. 0 tests run, 0 skipped, 0 failed.

PR #128, opened 2016-04-12
Remove reduntant blueprint

neykov:
The functionality in the blueprint is already provided by the Activator in same module.

Activator implemented in https://github.com/apache/cxf/commit/dae56b90c1de0dce1c00abe12108e364934eccf7, but blueprint not deleted.

Currently both the Activator & blueprint register the same service, on the same config admin PID.


cschneider:
In your PR you also removed the meta type xml. I kept the meta type in my commit. I hope this is what you intended.

neykov:
Thanks @cschneider. I made the change analogously to https://github.com/apache/cxf/commit/dae56b90c1de0dce1c00abe12108e364934eccf7 which removes the corresponding meta type, but makes sense to keep it.

PR #129, opened 2016-04-12
Fixes a NPE if Bus is not registered

neykov:
Just adding the bundle in an OSGi container causes a NPE in the logs because the Activator is called, but no Bus service is available. Even if there's a bundle providing the Bus service it could be registered after the asyncclient Activator executes so the `HTTPConduitFactory` service will be registered indeterministically. The changes will wait for the Bus service to become available and register the `HTTPConduitFactory` service.

Here's the stack trace of the exception:
```
2016-03-09 18:41:49,298 | ERROR | ort.http.async]) | configadmin                      | 3 - org.apache.felix.configadmin - 1.8.8 | [org.osgi.service.cm.ManagedService, id=146, bundle=84/mvn:org.apache.cxf/cxf-rt-transports-http-hc/3.1.4]: Unexpected problem updating configuration org.apache.cxf.transport.http.async
java.lang.NullPointerException
    at org.apache.cxf.transport.http.asyncclient.AsyncHTTPConduitFactory.addListener(AsyncHTTPConduitFactory.java:285)[84:org.apache.cxf.cxf-rt-transports-http-hc:3.1.4]
    at org.apache.cxf.transport.http.asyncclient.AsyncHTTPConduitFactory.<init>(AsyncHTTPConduitFactory.java:144)[84:org.apache.cxf.cxf-rt-transports-http-hc:3.1.4]
    at org.apache.cxf.transport.http.asyncclient.Activator$ConduitConfigurer.updated(Activator.java:79)[84:org.apache.cxf.cxf-rt-transports-http-hc:3.1.4]
    at org.apache.felix.cm.impl.helper.ManagedServiceTracker.updated(ManagedServiceTracker.java:189)[3:org.apache.felix.configadmin:1.8.8]
    at org.apache.felix.cm.impl.helper.ManagedServiceTracker.updateService(ManagedServiceTracker.java:152)[3:org.apache.felix.configadmin:1.8.8]
    at org.apache.felix.cm.impl.helper.ManagedServiceTracker.provideConfiguration(ManagedServiceTracker.java:85)[3:org.apache.felix.configadmin:1.8.8]
    at org.apache.felix.cm.impl.ConfigurationManager$ManagedServiceUpdate.provide(ConfigurationManager.java:1444)[3:org.apache.felix.configadmin:1.8.8]
    at org.apache.felix.cm.impl.ConfigurationManager$ManagedServiceUpdate.run(ConfigurationManager.java:1400)[3:org.apache.felix.configadmin:1.8.8]
    at org.apache.felix.cm.impl.UpdateThread.run0(UpdateThread.java:143)[3:org.apache.felix.configadmin:1.8.8]
    at org.apache.felix.cm.impl.UpdateThread.run(UpdateThread.java:110)[3:org.apache.felix.configadmin:1.8.8]
    at java.lang.Thread.run(Thread.java:745)[:1.7.0_80]
```

### DIFF
--- a/rt/transports/http-hc/src/main/java/org/apache/cxf/transport/http/asyncclient/Activator.java
+++ b/rt/transports/http-hc/src/main/java/org/apache/cxf/transport/http/asyncclient/Activator.java
@@ -18,6 +18,7 @@
  */
 package org.apache.cxf.transport.http.asyncclient;
 
+import java.io.IOException;
 import java.util.Dictionary;
 import java.util.Enumeration;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


cschneider:
I think this is still not correct. When a config update is received it takes a bus service from the ServiceTracker. The problem is that there is not only one CXF bus service in OSGi. So I think we need a completely different approach for this. I will try to sync with Dan or Sergey to find a good solution.

cschneider:
I checked with Dan Kulp how to solve the multiplicity of the busses. He has a good idea:
"Thus, the activator should be tracking al the bus things.  It should then likely to a bus.getExtension(AsyncHTTPConduit.class) to retrieve the existing conduit and configure it".

I think that should work fine. Do you want to change the PR accordingly? If not I can also tackle the implementation next week.

neykov:
Would be great if you give it a go. Agree current implementation doesn't work for multiple buses. Thanks @cschneider .

cschneider:
I think this is still not correct. When a config update is received it takes a bus service from the ServiceTracker. The problem is that there is not only one CXF bus service in OSGi. So I think we need a completely different approach for this. I will try to sync with Dan or Sergey to find a good solution.

cschneider:
I fixed the issue in https://issues.apache.org/jira/browse/CXF-6887. Would be great if you could validate it works for you.

neykov:
Thanks @cschneider closing this in favour of your changes.

For the case of multiple buses - wondering whether the client service should be able to target a specific bus (by name) instead of using the last seen configuration which is indeterministic due to bundle load order not being predefined.

cschneider:
We indeed have a race condition here. The code I wrote will update the ConduitFactory even when the config comes late but this will not influence endpoints that are already created. 

To target a specific bus would also not help as the bus creation is in control of the bundle that owns it. To tackle this is a correct way we would need a completely different approach.

PR #130, opened 2016-04-18
AccessTokenIntrospectionClient throws NPE

fjollberg:
AccessTokenIntrospectionClient throws NPE when introspection result contains expiration timestamp but "issued at" timestamp.


fjollberg:
Should be applied to 3.1.7.

sberyozkin:
Thanks for the patch. I guess the validation code will need to be tightened a bit to support the case where the expiresAt is less than the current time. I can take care of that.

sberyozkin:
See also http://git-wip-us.apache.org/repos/asf/cxf/commit/2f881786, thanks

PR #138, opened 2016-05-26
Allow netbeans to recognize files generated by cxf-xjc-plugin

CMoH:
Netbeans expects generated sources to be placed in folder `target/generated-sources/some-folder`, and recommends using the plugin-name for the leaf folder (which appears in the IDE's source tree as a hint to which plugin generated those sources).

The real fix here is that after opening the CXF sources in Netbeans, all projects having CXF as maven dependency are no longer linked to the jars in the maven repository, but to the CXF source tree. If generated sources are not recognized by the IDE within CXF's source tree, uses of generated classes (such as AuthenticationPolicy) generate an annoying false error alert.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -277,7 +277,7 @@
                                 <extension>org.apache.cxf.xjcplugins:cxf-xjc-javadoc:${cxf.xjc-utils.version}</extension>
                                 <extension>org.apache.cxf.xjcplugins:cxf-xjc-dv:${cxf.xjc-utils.version}</extension>
                             </extensions>
-                            <sourceRoot>${basedir}/target/generated/src/main/java</sourceRoot>
+                            <sourceRoot>${basedir}/target/generated-sources/cxf-xjc-plugin</sourceRoot>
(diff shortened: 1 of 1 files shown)


sberyozkin:
Hi, should all sourceRoot properties in various CXF modules, not only 'core', be updated too ?

CMoH:
@sberyozkin I ran into this problem when using CXF JAX-RS. It is possible that other users are encountering similar difficulties.

I applied a similar change to cxf-

CMoH:
@sberyozkin I ran into this problem when using CXF JAX-RS. It is possible that other users are encountering similar difficulties - though it seems few netbeans users load CXF in the same workspace with CXF client projects :)

I applied a similar change to cxf-rt-bindings-corba and it helped. Tests also succeed. I can add a commit that changes all sourceRoot entries if you think the change is beneficial.

sberyozkin:
Yes, please add a commit that changes all the entries, to make it consistent across the project, for JAXWS NetBeans users to benefit too. Thanks

PR #144, opened 2016-06-28
CXF-6942 - Fix to resolve problem finding wsdl with classifier

eivinhb:
When using <attachWsdl>true<attachWsdl> when generating a wsdl
an ealier fix added the classifier to the artifact. This
classifier is not used correctly by cxf-codegen-plugin when
<wsdlArtifact> is used to locate the wsdl-file. This fix adds
an extra check to see if an artifact with exists with type
containing the classifier.


eivinhb:
The tests that fail (IssueEncryptedUnitTest) seems to fail only on Jenkins. I can't reproduce it locally.

PR #150, opened 2016-08-03
[CXF-6986] If no application classes found, install a default applica…

johnament:
…tion.


johnament:
Don't merge - discussion only right now.

PR #151, opened 2016-08-03
[CXF-6989] Use Commons StrTokenizer to handle filenames with "=" if t…

osbornk:
…hey are quoted.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -56,7 +56,7 @@
             javax.annotation;version="${cxf.osgi.javax.annotation.version}",
             javax.xml.bind*;version="${cxf.osgi.javax.bind.version}",
             javax.xml.stream*;version="${cxf.osgi.javax.stream.version}",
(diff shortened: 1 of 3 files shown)


sberyozkin:
Thanks for the patch. Would it be possible to avoid bringing in a new dependency only to make it easier to deal with this case ?

osbornk:
Unfortunately, dealing with possibly quoted text is not entirely trivial. That is why I brought in StrTokenizer.  It handles this case really well. There is a decent amount of code involved in tokenization. So, that is why I had to introduce this dependency.

sberyozkin:
Well at the moment the CXF code splits by "=". This can be rewritten to check for the first "=" instead and it will be faster and will work for quoted and unquoted file names like filename="a.txt" or filename=a.txt or filename="a=b.txt". Introducing an extra dependency to process this parsing case seems avoidable, thanks

PR #154, opened 2016-08-07
[CXF-6998] Search for ApplicationPath in super classes since its not …

johnament:
…inherited.

See also: https://java.net/jira/browse/JAX_RS_SPEC-534

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/ResourceUtils.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/ResourceUtils.java
@@ -851,7 +851,7 @@ public final class ResourceUtils {
         JAXRSServerFactoryBean bean = new JAXRSServerFactoryBean();
         String address = "/";
         if (!ignoreAppPath) {
-            ApplicationPath appPath = app.getClass().getAnnotation(ApplicationPath.class);
+            ApplicationPath appPath = locateApplicationPath(app.getClass());
             if (appPath != null) {
                 address = appPath.value();
             }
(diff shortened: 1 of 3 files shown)


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/ResourceUtils.java
```diff
@@ -939,21 +939,29 @@ public static boolean isValidResourceClass(Class<?> c) {
         }
         return true;
     }
-    
+
+    public static ApplicationPath locateApplicationPath(Class<?> appClass) {
+        ApplicationPath appPath = appClass.getAnnotation(ApplicationPath.class);
+        if (appPath == null && appClass.getSuperclass() != Application.class) {
```
reta: I may be wrong here, but solving issue proxies we are going against JAX-RS 2.0 spec:
```
3.6 Annotation Inheritance
... Note that inheritance of class or interface annotations is not supported. ...
```

Does it make sense to limit the scope by identifying the particular use case that CXF dials with proxy (may be using Proxy.isProxyClass or alike) and only then fall back to super class hierarchy?

sberyozkin:
Hi Andriy

That spec statement is many years old, coming form 1.1 or earlier, before ApplicationPath was introduced (AFAIK) thus one may say well that statement does not apply to ApplicationPath as it is really about the resource classes and methods. 

(FYI, CXF checks class/interface level Path itself and never had any issues with various TCKs.)
 
I'd say it should def be valid to have a custom Application extending another custom Application with ApplicationPath. So IMHO the patch is safe enough :-)

Sergey

reta:
👍 , sounds good to me, thanks @sberyozkin and @johnament !

PR #155, opened 2016-08-08
[CXF-7000] Allow logging to be enabled on-the-fly

iweiss:
The purpose of LiveLoggingInInterceptor/LiveLoggingOutInterceptor is to work as a marker and avoid removing any user's LoggingInInterceptor/LoggingOutInterceptor implementations.


iweiss:
The failures in the build do not look related to this PR.

deki:
CXF-7000 and CXF-7047 are resolved so this should be closed as well, shouldn't it?

PR #160, opened 2016-08-21
CXF-6740 Propagate swagger.* properties from endpoint to swagger.

splatch:


reta:
Hi Lukasz,

Thanks a lot for the patch. What would be a use case to use such precise and not really straightforward Swagger configuration in CXF? I think the approach with path-based configuration (which essentially does the same trick by manipulating scanner/config ids) is simpler and does not really expose the internals of Swagger to CXF users. Plus, we could just make it a default behavior so the need to manipulate  scanner/config ids would be a very rare exception if at all. 

What do you think?
Thank you.

Best Regards,
    Andriy redko

splatch:
Indeed, this binds cxf more to swagger than relying only on swagger option, however this one offered consistency with Swagger approach. By this way we could also bind lifecycle and flush swagger objects from servlet context when endpoint goes down. Currently we can't do that with swagger bazed approach, for that we need to control also ServletContext attributes.

splatch:
Closing this PR as agreed in JIRA.

PR #161, opened 2016-08-21
CXF-6904 Customize Swagger scan logic so it doesn't hurt in OSGi.

splatch:
This PR overrides bean config and allows to use most of it's code used for swagger configuration. This keeps default behavior of swagger, with some small difference that when scan on Swagger2Feature is disabled it passes classes which are available on endpoint.


sberyozkin:
Very nice, significantly simpler than trying to create the application (we can still have that option if needed) - will apply next week

deki:
CXF-6904 is resolved so this should be closed as well, shouldn't it?

PR #170, opened 2016-09-15
Include details of validation exceptions in response in ValidationExceptionMapper

slavb18:
This PR contains of
1) addMessageToResponse flag to include validation errors in response
2) messageTemplate to set custom message format


sberyozkin:
Unfortunately we can not introduce Spring dependencies into this code. The simplest option to customize is to register a custom ValidationExceptionMapper

slavb18:
ok, thanks

slavb18:
and how about addMessageToResponse?

sberyozkin:
I've applied the patch with the minor modifications, thanks. I've only made StringBuilder instantiation optional, and kept the original message format. I'm open though to add 'Value of' - how does the message would read with it ? I hesitated because a constraint violation has a method like .getInvalidValue()

PR #174, opened 2016-10-03
CXF-7075: Improve performance by removing unnecessary string.length() and map.containsKey() calls.

andymc12:


rt/transports/http/src/test/java/org/apache/cxf/transport/http/HeadersTest.java
```diff
@@ -98,4 +99,21 @@ public void setHeadersTest() throws Exception {
         assertEquals("unexpected header", protocolHeaders.get("soapAction").get(0), headerValues[2]);
         
     }
+    
+    @Test
+    public void sensitveHeadersTest() {
```
reta: Typo here, sensitiveHeadersTest

reta: Typo here, please rename to sensitiveHeadersTest

deki:
CXF-7070 and CXF-7075 are resolved so this should be closed as well, shouldn't it?

PR #176, opened 2016-10-11
CXF-7084 Add a configurable KeyName value using the properties

spark404:
I've added properties to EncryptionProperties and SignatureProperties and also two new SecurityConstants.

This way a KeyName can be directly configured from the *Properties but also configured per message using a property. The Message configuration has a higher preference if both are set.


spark404:
Note that this ties directly to https://github.com/apache/santuario-java/pull/7 and requires an update to the WSS4J snapshot to reference the xmlsec snapshot.

So compile will fail, but i wanted to have this pull request here to show the intended usage of the santuario fix.

PR #181, opened 2016-10-19
ensuring CDI Bean<?> are always the CDI context ones

rmannibucau:


johnament:
The CDI failures are concerning.  The CC must be provided.  can you revert that part of the change? Or does it break something in OWB?

Probably should also commit under the JIRA ticket.

rmannibucau:
@johnament updated and removed the null

retest this please ?

integration/cdi/src/main/java/org/apache/cxf/cdi/JAXRSCdiResourceExtension.java
```diff
@@ -74,16 +74,16 @@
...
                 && Bus.class.isAssignableFrom(event.getBean().getBeanClass())) {
-            busBean = event.getBean();
+            hasBus = true;
         }
     }
 
     public void load(@Observes final AfterDeploymentValidation event, final BeanManager beanManager) {
+        // no need of creational context, it only works for app scoped instances anyway
+        final Bean<?> busBean = beanManager.resolve(beanManager.getBeans(CdiBusBean.CXF));
         bus = (Bus)beanManager.getReference(
```
reta: @rmannibucau, could you please explain the difference between getting bean from the event (in `collect` method) and resolving the bean by name in the `load` method: `busBean = event.getBean()` vs `beanManager.resolve(beanManager.getBeans(CdiBusBean.CXF))`? I am not sure I understand the subtleties between those two.

rmannibucau: There is nothing in the spec about Bean<?> stability (equals/hashCode). OpenWebBeans wraps all user beans (added through addBean(userBean)). For that reason capturing the bean (collect) can give you the wrong reference. Then if you use this one and not the CDI context one (from resolve) then you get 2 different Bean and therefore 2 different instances.

So concretely you should not use a bean not coming from the bean manager to be safe.

rmannibucau: Yes, that's why this PR doesn't store anymore busBean ref but look it up in the CDI context (bean manager) before getting a reference

reta: Right, my question was more about `applicationBeans`, `serviceBeans`, `providerBeans` and `featureBeans`?

rmannibucau: You are right, this is more rare to have these beans added programmatically since CXF is not able to read metadata from CDI but this is possible. That said not sure we should worry much yet on that case - can be another PR if a user hits it maybe.

rmannibucau: @johnament this is not the iterator since there is no specialization and a single item in the list, issue is really linked to the *instance* (as object instance) of the Bean<?>. If user does event.addBean(new MyJAXRSBean()) then you will have it as well. Surely less impacting cause you will get one instance in CXF and one in the app if the resource is injected...which never happens in practise.

reta: @rmannibucau This is kind of worries me, Bean< ? > is essentially a piece of metadata. There is no hacks or backdoors we use here to get the instance of the Bean< ? >, everything is relying on CDI API. So if there are multiple ways to retrieve Bean< ? > instances, as a developer/user, I would expect to have the same result at the end. It seems like this does not hold true and is implementation specific, as you pointed out. I think the fix should target not `bus` only so we won't get back to this issue again. Does it make sense?

rmannibucau: Concretely there is a single reliable way - resolve(getBeans()). That said the cases where this is an issue for providers and resources are very limit today for JAX-RS (but can create some nice feature if CXF embraces - typically reading metadata like @Path from cdi AnnotatatedMethod instead of doing pure reflection).

For features this is a valid point I think.

So concretely what does it mean? That discovered instances through CDI scanning (collect()) is only valid for @Produces or managed bean. Think it is fair to start with that but happy to try to do another PR fixing it more if you think it makes sense.

johnament: For the case of what we're using, collecting beans vs getting all beans will be a little different.  Specifically for `serviceBeans` there is no marker interface for JAX-RS resources,  and provider beans have the same requirement - they're discovered via Object with a non-qualifier annotation.

rmannibucau: Not sure I follow, collect can capture the type (think for this use case beanClass is enough) + qualifiers and do the lookup with the bean manager.

reta:
Just debugged the issue, it seems we have not much choice here, OpenWebBeans do really wrap the bean into another class, while Weld does not. Looks good to me.

reta:
Merged, thanks @rmannibucau @johnament !

PR #183, opened 2016-10-22
[CXF-7093] Change the spring boot example to use samples parent

johnament:


distribution/src/main/release/samples/jax_rs/spring_boot/README
```diff
@@ -8,7 +8,7 @@ The sample uses Maven. It can be built and run from the command line:
 ----
 $ mvn -Pserver
 or
-$ java -jar target/spring-boot-sample-rs-cxf-3.2.0-SNAPSHOT.jar
+$ java -jar target/spring-boot-sample-rs-cxf-*.jar
```
johnament: I figure this makes it more generic and version in-specific

distribution/src/main/release/samples/jax_rs/spring_boot/README
```diff
@@ -31,10 +31,3 @@ or access it from the CXF Services page:
 
   http://localhost:8080/services/helloservice/services
   and follow a Swagger link.
-
-To run the client from a command line open a new terminal window and run:
```
johnament: This wasn't working for me even before the change.

distribution/src/main/release/samples/pom.xml
```diff
@@ -99,6 +99,7 @@
         <module>wsdl_first_rpclit</module>
         <module>jms_pubsub</module>
         <module>jax_rs/spring_security</module>
+        <module>jax_rs/spring_boot</module>
```
johnament: I only added this one for now.  If you're in agreement, I'll also apply the changes to the auto configure example.

distribution/src/main/release/samples/jax_rs/spring_boot/README
reta: If I am not mistaken, this command may not work, you have to provide the exact runnable: `java -jar `ls target/*.jar`` (but than it won't work for Windows users)

distribution/src/main/release/samples/jax_rs/spring_boot/pom.xml
```diff
@@ -2,24 +2,31 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
-        <groupId>org.springframework.boot</groupId>
-        <artifactId>spring-boot-starter-parent</artifactId>
-        <version>1.3.6.RELEASE</version>
-    </parent> 
+        <groupId>org.apache.cxf.samples</groupId>
```
reta: This is good one, I had it on my least as well, will probably change other samples :-), thanks.

distribution/src/main/release/samples/pom.xml
reta: As for me, makes sense to update other samples, @sberyozkin what do you think? Thanks, @johnament !

distribution/src/main/release/samples/jax_rs/spring_boot/README
johnament: @reta Ok, makes sense.  I'll change the build output to not include version #.

distribution/src/main/release/samples/jax_rs/spring_boot/src/main/java/sample/rs/client/SampleRestClientApplication.java
```diff
@@ -31,10 +31,7 @@
 @SpringBootApplication
 @EnableJaxRsWebClient
 @EnableJaxRsProxyClient
-public final class SampleRestClientApplication {
-    private SampleRestClientApplication() {
-    }
-    
+public class SampleRestClientApplication {
```
johnament: So... this sample wasn't actually working.  The client was not defined properly.  This makes it work now

distribution/src/main/release/samples/jax_rs/spring_boot_scan/application/README
```diff
@@ -19,9 +19,11 @@ $ mvn spring-boot:run
 
 2. Run the server 
 ----
-$ mvn -Pserver
+$ mvn spring-boot:run
```
johnament: Wherever possible, I got rid of the unnecessary profiles.

distribution/src/main/release/samples/jax_rs/spring_boot_scan/client/pom.xml
```diff
@@ -43,22 +52,23 @@
             <artifactId>org.qi4j.library.circuitbreaker</artifactId>
             <version>${cxf.zest.version}</version>
         </dependency>
+        <dependency>
```
johnament: Ok, this dependency was a bit of an annoyance.  Its defined in the samples parent, but it interferes with spring boot's use of logback.

johnament:
Sorry - just realized the commits weren't squashed.  Fixed

reta:
IMO, looks cleaner, thanks @johnament, leaving decision to @sberyozkin as he authored or/and worked on these examples.

sberyozkin:
John, does the demo work without a version qualifier of the final jar, when it is started with a "java -jar" ? Also as a side note I'm not convinced -Pserver does not make sense for the server only applications as it hides the specific plugin options - but I'm fine with getting rid of them whenever possible

johnament:
Well we can make spring boot the default goal.  I'm not sure about your question though.  This just changes the file name and I've been able to run the examples with these instructions....

sberyozkin:
While I appreciate this quality patch I'd like to say that the way the spring_boot_scan server and client applications are started is really off topic for CXF-7093, however as I noted I can def live with asking users to do 'mvn spring-boot:run' again instead of -Pserver or -Pclient. Re "java -jar" - I'll experiment tomorrow....

johnament:
I'm not sure what you mean.  That didn't change other than the way mvn is called - mvn -Pclient/server vs mvn spring-boot:run for all 3 modules.  If you want me to make that work I can.  The problem is that the wrong spring boot version was getting used..

PR #184, opened 2016-10-23
[CXF-6988] Add more systests for CDI, including OpenWebBeans.

johnament:
Ok, so sorry about the large PR.  Introduced extra tests for systests to be able to test additional functionality, like OWB support in addition to Weld, and tests for the no application feature.

Originally when looking at this, I thought to use Arquillian.  It isn't a good choice for the type of testing you're doing.  It's also beneficial to be able to run all tests in a single pass, instead of multiple profiles (the normal Arquillian pattern).


johnament:
I believe the test failures are unrelated to my change.

PR #193, opened 2016-11-10
CXF-7130 Maven plugin to invoke SOAP service

zregvart:
I needed a Maven plugin that would allow me to invoke SOAP service as a part of Maven lifecycle so I implemented one, and I'm hoping that you can include it in CXF.

The plugin has the functionality of calling SOAP service given a WSDL, port, service, operation, request body and optional SOAP headers. Additionally Maven project properties can be set from the response using XPath expressions and additional XPath expression can be used to determine if the request needs to be repeated.

This is a minimal implementation that I needed for my use-case, hope you find it useful and that it can be incorporated in CXF.

This is a new pull request, I squashed the commits to make it a bit easier to follow.


zregvart:
Had a discussion with @ffang and decided that this was not a good fit for CXF. Thanks for the consideration!

PR #195, opened 2016-11-15
[CXF-7132] Use the original class when it comes to determining classe…

johnament:
…s for ResourceProvider usage.

### DIFF
--- a/integration/cdi/src/main/java/org/apache/cxf/cdi/CdiResourceProvider.java
+++ b/integration/cdi/src/main/java/org/apache/cxf/cdi/CdiResourceProvider.java
@@ -24,12 +24,12 @@ import org.apache.cxf.message.Message;
 public class CdiResourceProvider implements ResourceProvider {
     private final Object instance;
     private final Class<?> resourceClass;
-    
+
     CdiResourceProvider(final Class<?> resourceClass, final Object instance) {
         this.resourceClass = resourceClass;
         this.instance = instance;
     }
-    
+
     @Override
     public Object getInstance(Message m) {
         return instance;
--- a/integration/cdi/src/main/java/org/apache/cxf/cdi/DefaultApplication.java
+++ b/integration/cdi/src/main/java/org/apache/cxf/cdi/DefaultApplication.java
@@ -18,9 +18,11 @@
  */
 package org.apache.cxf.cdi;
 
+import javax.enterprise.inject.Vetoed;
 import javax.ws.rs.ApplicationPath;
 import javax.ws.rs.core.Application;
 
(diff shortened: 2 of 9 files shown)


johnament:
Looks like the failures are not related to CDI... https://builds.apache.org/job/CXF-Trunk-PR/200/

reta:
Thanks a lot for the fix, @johnament, will work on merging tonight

rmannibucau:
should loadResourceProviders support not normal-scoped instances too (doesn't ATM it seems)?

johnament:
@rmannibucau you are correct, there is a regression problem with non-normal scoped beans.  The code works.  Specifically for dependent and singleton, since they're not using proxies, the instance you get back is the same.  This essentially makes dependent == singleton.

You can see that the test does include both normal scoped and dependent beans https://github.com/johnament/cxf/tree/81e9215b75c02a6b108e94f474b9aa6011c8f00c/systests/cdi/base/src/main/java/org/apache/cxf/systests/cdi/base - look at BookStore and VersionService as examples.

The code does work, but I would expect a per-request pattern to be used for dependent.  More refinement required to clean up this CDI integration.

rmannibucau:
@johnament think singleton per default is fine (even good) but I don't think it works. Add a @PreDestroy releasing a socket pool (http pool to get some meta config for instance) and you will just leak this since the creational context is not released.

reta:
@rmannibucau @johnament I think the CdiResourceProvider should be involved more here. The creation context could be released when `release` call happens.

johnament:
@rmannibucau yes, that is the other problem.  But that's been a long standing issue with leaking instances, way before my time.  I'll clean that up though.  Raised https://issues.apache.org/jira/browse/CXF-7134

@reta Yes, I agree.  That's where I plan to put the `release`.

johnament:
Ok, I added a couple more commits to this.  I'm not 100% happy with CXF-7134 but don't want to block a potential release.

What I'm seeing is that validation feature doesn't get registered on the JAX-RS resource when I move the creational context to the point of getting the instance.  Need to debug a bit further, but I think this is OK to merge for now.  Just leave 7134 open I'll clean up further.

reta:
Hey @johnament , could you please create a separate PRs for these commits? I am about to close the original one, related to @RequestScoped. Now things may get complicated :-) 
Thanks!

johnament:
I can.  I wasn't sure if you were holding off due to the conversation here.

reta:
I taken the approach we have discussed, moving more responsibilities to CdiResourceProvider, I think I have gotten a bit cleaner implementation.

PR #199, opened 2016-11-16
CXF-7133 Clean up how default app is loaded and avoid extra XML confi…

johnament:
…guration.


integration/cdi/src/main/resources/META-INF/beans.xml
```diff
@@ -2,7 +2,4 @@
 <beans xmlns="http://java.sun.com/xml/ns/javaee"
 	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
 	xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://jboss.org/schema/cdi/beans_1_1.xsd">
-	<scan>
```
johnament: Having this config here makes shaded JARs hard to deal with in CXF.  CDI has this feature built in to ignore the bean, which works better IMHO.

PR #203, opened 2016-11-20
Support token api with not Authentication header handling

teavirdis:
Some apis do not support Authorization header. There is no way to switch to parameter approach through configuration. This parameter will make it possible.


sberyozkin:
Thanks for the patch - note I had to set this new property to 'true' by default - to preserve the original filter behaviour with respect to how it authenticates when acquiring the token

PR #210, opened 2016-12-03
[CXF-7162] Inconsistent reading of formatted xml when validating schema

elrodro83:
When calling `StaxUtils.copy()` with a source reader that is validating schema, the spaces between tags used for formatting the XML are not copied to the output.
This causes, when used, the soap body signature validation to fail when using schema validation.
I have already reported this to woodstox (https://github.com/FasterXML/woodstox/issues/29) but anyway, `StaxUtils.copy()` should be aware of both situations (CHARACTERS AND SPACES).


elrodro83:
Per comment on woodstox issue (https://github.com/FasterXML/woodstox/issues/29#issuecomment-264719777), the reader is working as expected and cxf should process its input accordingly. This pr fixes that.

PR #211, opened 2016-12-08
Streaming fix

michail-nikolaev:


sberyozkin:
Thanks for the patch, note it contains an older fix which I've already applied on your behalf earlier. I did a minor tweak and added a utility method to JAXRSUtils to keep the set of streaming classes there, hope you are ok with that. The checkstyle was failing, may be because you had a wildcard import which is probably not supported by the checkstyle. The commit is going now, thanks

PR #218, opened 2016-12-30
[CXF-7201] Correctly format UserInfo response when no signature or encryption is used.

elfogre:
Create Json UserInfo with JwtUtils.claimsToJson when no signature or encryption is Used.


elfogre:
Jenkins test errors are in unrelated modules.

deki:
CXF-7201 is resolved so this should be closed as well, shouldn't it?

PR #221, opened 2017-01-11
[CXF-7210] Fix StringIndexOutOfBoundsException

sebhoss:
The logic goes as follows:

1) Base- & request-URI are the same, then only add base-URI.
2) If request-URI starts with base-URI, only add the parts from request-URI that are not already included in base-URI
3) Otherwise add both base-URI and request-URI


sebhoss:
There is a single test failure (see https://builds.apache.org/job/CXF-Trunk-PR/org.apache.cxf.systests$cxf-systests-ws-rm/269/testReport/org.apache.cxf.systest.ws.rm/SequenceTest/testOnewayAnonymousAcksSuppressed/), however that seems to be unrelated to the above changes.

PR #229, opened 2017-01-29
[CXF-7235]: MessageSelector is not set in JMSConfiguration due to missing property in JMSEndpoint

mitch-lbw:
added messageSelector attribute to JMSEndpoint and enhanced JMSConfigFactory to map this attribute to JMSConfiguration


sberyozkin:
Can you please update this patch so that only the proposed changes are visible ? Looks like some format-related changes made into it too

PR #231, opened 2017-01-30
[CXF-7235]: MessageSelector is not set in JMSConfiguration due to missing property in JMSEndpoint

mitch-lbw:
Re-Pull request of #229


cschneider:
I have applied with some changes in formatting. Thanks for the contribution. For new pull requests better keep them in a branch instead of master.

mitch-lbw:
Hi Christian,
thanks for the hint - it was my first pull request, apparently there had to be something going wrong ;)
Greets,
Michel

cschneider:
No problem.
Please also remember to close the PR once the changes are applied unless it auto closes.
Btw. If you have changes in your PR you do not need to create a new PR. You can add commits and you can even use git reset to undo several commits, put them into one commit and do a push -f. While this is bad in master it is ok on a branch that only you work on.
Christian

PR #232, opened 2017-01-31
Upgrade to spring-boot 1.5.1.RELEASE

rajadilipkolli:
Spring boot 1.5.1.RELEASE was released yesterday and we can use the same in cxf-spring-boot-starters. so that we can seamlessly integrate in cxf 3.2.0

PR #233, opened 2017-02-01
CXF-7198: Update Spring-Boot to 1.5.1 release

deki:


sberyozkin:
See my comments to #232. The actual version upgrade does not change anything in CXF but a change a new major SprongBoot dep. This is not a bad idea in itself - but will we break 1.4.x SpringBoot users if we start inadvertently utilizing 1.5.x features where 1.4.x is still not exactly old ?

sberyozkin:
In other words, given that it was confirmed CXF starters can now work with 1.5.x, I think that we should only upgrade to 1.5.x if we start using 1.5.x features inside CXF starters - to make sure CXF Spring Boot 1.4.x and 1.5.x users can use the startes

deki:
Ok.

sberyozkin:
I've opened https://issues.apache.org/jira/browse/CXF-7237 to get this issue tracked and be more visible, thanks

PR #236, opened 2017-02-13
CXF-7137: extended unittest to show non working security definitions

deki:
I've extended the unittest to show that the implementation from CXF-7137 isn't working at the moment. While contact is part of the BeanConfig and correctly serialized, the SecurityDefinitions are not shown in the swagger.json as Swagger's BaseApiListingResource handles only the attributes from BeanConfig.

I'm not sure if this can be fixed without changes in Swagger, please have a look.


sberyozkin:
Thanks for the patch

PR #237, opened 2017-02-17
Added tests for DefaultResourceManager, fixed a bug, and improved a comment

edorfaus:
I noticed that my resource resolvers were sometimes called out of (expected) order, which for me was significant.

I tracked it down to some code in JaxWsServerFactoryBean that creates a new DefaultResourceManager with a copy of the resolvers from the original manager - while the code in DefaultResourceManager for adding a list of resolvers adds them in reverse order.

I think this is wrong, even if only due to the principle of least astonishment, so went ahead and fixed it, after adding some tests that show the problem (and also tests other parts of the class).

I also changed a comment on the ResourceManager interface that I thought was rather unclear (resolving the seeming self-contradiction based on the actual behavior of the code).

I'll note that this was only tested on top of the 3.1.x-fixes branch, since I'm having trouble getting master to compile (even without any changes). However, as far as I can tell only the whitespace is different in these files between those versions. (If you want the 3.1.x version, there's a branch for it in my fork on github.)

### DIFF
--- a/core/src/main/java/org/apache/cxf/resource/DefaultResourceManager.java
+++ b/core/src/main/java/org/apache/cxf/resource/DefaultResourceManager.java
@@ -73,8 +73,15 @@ public class DefaultResourceManager implements ResourceManager {
         }
     }
     public final void addResourceResolvers(Collection<? extends ResourceResolver> resolvers) {
+        int i = 0;
         for (ResourceResolver r : resolvers) {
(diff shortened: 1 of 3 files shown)


edorfaus:
I tried to look into the Jenkins test failures, and they seem unrelated to me, but I'll admit I don't really understand how they all work... So I could be wrong about that.

PR #247, opened 2017-03-21
CXF-7291: sharedPackageName for Swagger desc should ignore proxys

deki:


sberyozkin:
Hi Dennis, is it related to the client proxy thing ?

deki:
yes but in my view this should be fixed in any case

core/src/main/java/org/apache/cxf/common/util/PackageUtils.java
```diff
@@ -56,7 +57,9 @@ public static String getSharedPackageName(List<Class<?>> classes) {
         List<List<String>> lParts = new ArrayList<List<String>>(classes.size());
         List<String> currentParts = new ArrayList<>();
         for (Class<?> cls : classes) {
-            lParts.add(StringUtils.getParts(getPackageName(cls), "\\."));
+            if (!Proxy.isProxyClass(cls)) {
```
reta: In this case, should we get the real class using `getProxyClass` and extract the package from it?

deki: Good idea, I first thought to use InvocationHandler stuff for this but in the method only the class itself is available. And afaik it's not possible to get the real class as there might be multiple interfaces...

PR #253, opened 2017-03-31
[CXF-7309] JAX-RS @Context fields throw NPE in OSGI hot deployed filters

nhtzr:
- Add logic to inject using already set ThreadLocal proxy whenever possible.
- Update unit test, so getter and setter correspond to same field in Customer class.


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/InjectionUtils.java
```diff
@@ -1167,10 +1161,12 @@ public static void injectContextMethods(Object requestObject,
                 if (!cri.isSingleton()) {
                     InjectionUtils.injectThroughMethod(requestObject, method, o, message);
                 } else {
-                    ThreadLocalProxy<Object> proxy 
-                        = (ThreadLocalProxy<Object>)cri.getContextSetterProxy(method);
+                    Object proxy = extractFromSetter(requestObject, method);
+                    if (!(proxy instanceof ThreadLocalProxy)) {
```
sberyozkin: Hi, can you explain please why this check is done, what else can the proxy be, apart from null or ThreadLocalProxy ? Thanks

nhtzr: Hi! Sorry for the late response!

The intention of this condition is to prevent NPE or ClassCastException that may happen in case `extractFromSetter()` fails or in case it returns some completely unexpected value.
The unexpected value could happen from the unfortunate coincidence in which a getter exists but it doesn't follow a Bean Property pattern. 
Another possibility would be the property being overwritten by code outside cxf (That should be unlikely, but it is possible) and therefore no longer a proxy.

sberyozkin: Np, thanks. See if we see an unexpected value then it means there's a problem lying somewhere else, this is where my concern is. If some code somewhere depends on this 'unexpected' value then the proposed fix will break that code, right ? If a bean naming pattern is not followed then IMHO CXF should not worry about it. In case of CXF-7309, what is returned there ? Thanks

nhtzr: I think it will be fine, because we are not messing with the unexpected value (so no code breakage outside) and Line 1169 is now unsafe unless we make that check (so no code breakage in this method).

In case of CXF-7309, the value returned here is another `ThreadLocalProxy` different from the one in `cri.getContextSetterProxy(method)`
This happens because of a singleton filter in a OSGI bundle. Whenever I hot re deploy that bundle, a new filter provider instance is created, and all OSGI proxy references (saved as `requestObject` here) are redirected to this new instance, but CXF has no way to know it has to update its internal `AbstractResourceInfo#getSetterProxyMap` to reflect that. All of this causes CXF to inject references into stale ThreadLocalProxies that no one is gonna use.

Hence the patch, which basically translates to "check getter first; check cri otherwise if the getter is not convenient"


As a side note: Now that you mention bean naming sense.
I reread the `@Context` documentation, and it does say that it should annotate bean property method (other options are a field or a parameter). :thinking: 
Do you think it would be beneficial to drop the `AbstractResourceInfo#getSetterProxyMap` and instead rely on getters?
I think it would be good, but maybe the change is big enough for reconsideration.


I hope I addressed all you asked!

PR #256, opened 2017-04-10
Add attributes in serialize functions

renslinden:


renslinden:
Jenkins says: "Build timed out (after 90 minutes). Marking the build as aborted."

What to do now?

coheigea:
Don't worry about the build timing out. Any chance of modifying/adding some tests for these new features?

renslinden:
@coheigea As you can see I have modified the unit tests in my last commit, so they runs well again.

coheigea:
The unit tests have been modified, but I'd prefer to see a unit test added that (for example) failed with the old code but passes with the new code. Does changing from "add" to "set" for arrays break the API by the way?

dkulp:
Definitely concerned about the set->add change.  That will certainly break peoples code.   Can the original method be retained with the add method being added?

gnodet:
Closing: this PR is 9 years old. Maintainers raised breaking API concerns (set→add change). The wsdl2js attribute support feature remains unimplemented but is low priority.

PR #259, opened 2017-04-13
CXF-7328: fixed property metadata for Spring Boot CxfProperties, updated maven-bundle-plugin

deki:


deki:
I just noticed that it also works with the old version of maven-bundle-plugin. So #260 is enough.

PR #261, opened 2017-04-14
[Cxf 7239] bug fix to allow springboot to validate path in CxfProperties

javacracker:
Fix bug related to cxfProperties and added tests cases to validate.
Missing @Validated on CxfProperties


integration/spring-boot/autoconfigure/pom.xml
```diff
@@ -30,6 +30,13 @@
         <version>3.2.0-SNAPSHOT</version>
         <relativePath>../../../parent/pom.xml</relativePath>
     </parent>
+  <properties>
```
deki: please take a look at cxf-parent. most of the versions are already given there.

PR #262, opened 2017-04-18
[CXF-7339] Fix NPE due to weak reference being GC-ed

asoldano:
https://issues.apache.org/jira/browse/CXF-7339

### DIFF
--- a/core/src/main/java/org/apache/cxf/common/jaxb/JAXBContextCache.java
+++ b/core/src/main/java/org/apache/cxf/common/jaxb/JAXBContextCache.java
@@ -229,7 +229,10 @@ public final class JAXBContextCache {
                 if (cachedContextAndSchemasInternal != null) {
                     context = cachedContextAndSchemasInternal.getContext();
                     if (context == null) {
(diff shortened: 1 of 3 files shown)


asoldano:
@andymc12 , actually I expected a concern about this, that's why I opened up a PR instead of committing directly. I'm not sure we want to keep the former behavior wrt supporting null keys. I don't think entries with null keys were actually used before, null keys happened to be "supported" because of the internal implementation detail choice. This said I believe the code that I fixed is not calling remove(null) on purpose, basically the change here is exposing a bug that was there before.
@dkulp what's your opinion on this?

dkulp:
I actually think doing "both" would be the right thing to do.   The fix in this PR is likely correct as calling remove with a null key is pointless as there shouldn't ever be a null "Set" of classes as a key in the cache.  Possibly and empty set, but likely not a null set.

However, protecting against it in the CacheMap is also a good thing.

asoldano:
@andymc12 I've added another commit for the changes in the CacheMap, hopefully this should be enough.

dkulp:
Yea... that looks good to me.

asoldano:
OK, thanks, I've merged the PR.

PR #264, opened 2017-04-19
[CXF-7001] Update JAX-RS SpringBoot starter to support JAX-RS Applications

deki:
I'm not happy with the setJaxrsResources method. Can we refactor and rename it?


sberyozkin:
I agree, I'd just override createJaxrsServer, if AppPath is there then complete it otherwise delegate to the superclass, I can take care of it while working on your patch

sberyozkin:
Can you please create another PR related to JavaDocs and SpringBoot version update, thanks

sberyozkin:
Never mind, spring boot version was already set to 1.5.2.RELEASE on master, I'll merge javadocs updates

PR #265, opened 2017-04-20
CXF-7343 Also add InFault and OutFault interceptors

spark404:


sberyozkin:
Thanks, does CXF run in-fault interceptors on the server side ? May be I'm wrong, need to double check, I thought they were only run on the client side.

spark404:
I didn't check, but I have difficulties thinking up a case for in fault interceptors. I found the issue when adding an out fault interceptor.

I added the in fault more for completeness. If somebody finds a use case at least the support is there 😁

sberyozkin:
I just checked, the in fault interceptors are not run (try to throw a Fault from the in interceptor) but the in interceptors.handleFault method is called. Ideally we'd not have the option to set the inFaultIntercptors at the server factory bean level; probably too sensitive modify the API at this stage

PR #267, opened 2017-04-23
[CXF-7343] cxfOutFaultInterceptors was never queried

deki:


sberyozkin:
Ouch, thanks for spotting it. To be honest I'd rather get rid of the duplicates (in, out, out fault interceptors) at the AbstractSpringComponentScanServer

deki:
Makes sense, I introduced some add methods.

sberyozkin:
Thanks, ended just adding them directly to the returned lists

PR #270, opened 2017-04-27
CXF-7349: NPE when thread context classloader is null

andymc12:
A null context classloader is valid - it means that the JVM's boot classloader is the context.  When using the ClassLoaderUtils class, there are a few places where we get the thread's context classloader, but don't check it for null before invoking methods on it - such as getResource/getResources/etc.

This change ensures that we get a non-null classloader by using the system classloader when the current thread context classloader is null.


sberyozkin:
Hi Andy, this fix looks good, please go ahead with the merge, thanks

andymc12:
@sberyozkin @deki I am not able to merge - my view of this PR says that "Only those with write access to this repository can merge pull requests."  I should have access to commit, but maybe the ACLs have not been updated on github?

deki:
Nobody can merge Apache repos on Github. Instead you should rebase your branch on Apache repo (see http://cxf.apache.org/source-repository.html), squash commits and push there.

andymc12:
@deki thanks for the tip.  I've merged this change to master and 3.1.x-fixes.

PR #272, opened 2017-05-09
CXF-7354: Use SLF4J markers to differenciate payload-logging

davidkarlsen:


sberyozkin:
Hi David, I'd rather restore the ability to pass a custom logging category. Service qname can be configured for jaxrs endpoingts too, so this SOAP and REST markers will just cause the confusion

sberyozkin:
Actually, it looks OK, I was concerned that a 'REST' will be printed for every REST request in the default case :-). Looks good to me.

PR #273, opened 2017-05-18
lock RefreshToken entity with pissimistic locking to avoid concurrent updates

vgagara-talend:
issue: https://issues.apache.org/jira/browse/CXF-7374
orig issue: https://jira.talendforge.org/browse/TPSVC-2439

### DIFF
--- a/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/provider/JPAOAuthDataProvider.java
+++ b/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/provider/JPAOAuthDataProvider.java
@@ -19,12 +19,16 @@
 package org.apache.cxf.rs.security.oauth2.provider;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
+
 import javax.persistence.EntityManager;
 import javax.persistence.EntityManagerFactory;
 import javax.persistence.EntityTransaction;
+import javax.persistence.LockModeType;
 import javax.persistence.TypedQuery;
 
 import org.apache.cxf.helpers.CastUtils;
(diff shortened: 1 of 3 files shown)


sberyozkin:
Hi, thanks for the patch. Can you please consider locking at the AbstractOAuthDataProvider level instead, so that all its extensions (JPA2, JCache, Ehcache and future ones like Mongo-based) can benefit ?
I'd consider introducing 'Object refreshTokenLock' and use it lock at https://github.com/apache/cxf/blob/master/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/provider/AbstractOAuthDataProvider.java#L211

Let me know please if it works for you

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/provider/JPAOAuthDataProvider.java
```diff
@@ -213,6 +221,25 @@ public RefreshToken execute(EntityManager em) {
...
 
+    protected void lockRefreshTokenForUpdate(final RefreshToken refreshToken) {
+        try {
+            execute(new EntityManagerOperation<Void>() {
+
+                @Override
+                public Void execute(EntityManager em) {
+                    Map<String, Object> options = Collections.emptyMap();
+                    if (pessimisticLockTimeout > 0) {
+                        Collections.singletonMap("javax.persistence.lock.timeout", pessimisticLockTimeout);
```
andrei-ivanov: options = ...

sberyozkin: Good catch, I'm applying PR now so will fix

sberyozkin:
Viacheslav, the patch has been applied with minor updates to let people use a Java lock if they prefer (JPA2 provider is currently using the pessimistic lock as per your code).

However the test fails so I disabled it - can you please have a look and also copy it it to JPACMTOAuthDataProviderOpenJPATest class to make sure it works with OpenJPA - to make sure we don't break the JPA2 provider depending on OpenJPA ? Thanks

PR #276, opened 2017-06-03
misleading message when class not found

rosiecki:


deki:
Thanks for the pull request. 
What is shown before and after the patch?

rosiecki:
I'm assuming we have a webservice:
```
package xxx.yyy.zzz;
@WebService(endpointInterface="xxx.yyy.zzz.WsEndpointInterface")
public class WsImpl {
}
```
- Message before patch: "Unable to load the class xxx.yyy.zzz.WsImpl"
- Message after patch: "Unable to load the class xxx.yyy.zzz.WsEndpointInterface"

rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/ServiceImpl.java
```diff
@@ -589,8 +589,7 @@ private QName getPortTypeName(Class<?> serviceEndpointInterface) {
                 try {
                     seiClass = Thread.currentThread().getContextClassLoader().loadClass(epi);
                 } catch (ClassNotFoundException e) {
-                    Message msg = new Message("COULD_NOT_LOAD_CLASS", BUNDLE,
-                                              seiClass.getCanonicalName());
+                    Message msg = new Message("COULD_NOT_LOAD_CLASS", BUNDLE, api);
```
deki: Should be epi and not api

rosiecki: I fixed it.

deki:
I've applied your changes. Thank you!

PR #282, opened 2017-06-13
[CXF-7405] Removal of commons-lang dependency and upgrade to Velocity 2

asoldano:
Initial proposal for fixing CXF-7405. Two notes:
1) I had to build Velocity from sources from its 2.0 tag at https://github.com/apache/velocity-engine.git ; can we push the binaries to the maven repo?
2) I'm not sure what to do with the logger at tools/common/src/main/java/org/apache/cxf/tools/common/VelocityGenerator.java , see commented code there.


deki:
I see that you are already discussing the velocity 2.0 release on dev@velocity.apache.org so let's wait with the merge until it's done.

deki:
@asoldano Velocity 2.0 is released now, please go ahead!

PR #285, opened 2017-06-22
[WIP] Fixing CNFE build errors when building under JDK1.8 using toolchains

exabrial:
[WIP] Fixing CNFE build errors when building under JDK1.8 using toolchains

### DIFF
--- a/README
+++ b/README
@@ -85,4 +85,3 @@ Thank you for using CXF!
 
 The Apache CXF Team
 http://cxf.apache.org/
-update


coheigea:
This doesn't work with JDK 1.6:

Failed to execute goal org.jibx:maven-jibx-plugin:1.3.1:test-bind (default) on project cxf-systests-databinding: org/apache/bcel/classfile/Attribute : Unsupported major.minor version 51.0

systests/databinding/pom.xml
```diff
@@ -165,7 +165,7 @@
             <plugin>
                 <groupId>org.jibx</groupId>
                 <artifactId>maven-jibx-plugin</artifactId>
-                <version>${cxf.jibx.version}</version>
+                <version>1.3.1</version>
```
dkulp: Why are you removing a property version and putting a hard coded version number in there?

exabrial: Hey Daniel!

I think because the version specified by the variable is needed to make the code compile and produce the correct artifacts (the bindings), but the plugin needs to be a newer version to work with JDK 1.8 and modern maven. 

Keep in mind I have this branch marked WIP as I sort bugs out. The goal is to fix all the known CVEs in 2.6 and make it buildable without having to maintain a separate environment.

coheigea: Can this PR be closed? We can't apply the fix as otherwise the branch won't build with older JDK versions.

exabrial:
Yes, thanks

PR #286, opened 2017-06-26
[CXF-7429] Prevent NPE on file log and null writer

blasd:

PR #287, opened 2017-06-26
[CXF-7429] Prevent NPE on file log and null writer 

blasd:


deki:
I've closed this PR as the issue is already fixed. See also #286.

PR #292, opened 2017-07-17
[CXF-7447] Add doPrivs to avoid AccessControlExceptions

andymc12:
We found a few places in JAX-RS code where AccessControlExceptions were being thrown when using a security manager.  This suggested fix adds doPriv blocks around the offending code.


sberyozkin:
+1

PR #294, opened 2017-07-19
[CXF-7430] The method logInputStream of the LoggingInInterceptor class passes txt/plain content type to the writePayload method if the payload content was truncated.

alex-korobko:
The method logInputStream of the LoggingInInterceptor class passes txt/plain content type to the writePayload method if the payload content was truncated.


alex-korobko:
The method logInputStream of the LoggingInInterceptor class passes text/plain content type to the writePayload method if the payload content was truncated. It should avoid issues that happen in the writePayload method of the class AbstractLoggingInterceptor when the payload stream contains truncated data (and therefore invalid xml) and content type is set to text/xml. In that case the PrettyXmlStreamWriter fails to output the payload stream.

sberyozkin:
Thanks, can you please rewrite the test code with the Java 7 in mind - CXF 3.1.x should compile with Java 7

alex-korobko:
Rewrote the test code to comply Java 7 spec. Strangely enough, though, that the core project doesn't compile with default Java 7 language spec in Intellij Idea, it shows several type conversion errors in other files.

PR #296, opened 2017-07-25
CXF-7453 check subCodes only in case of SOAP 1.2

istudens:
Restricts handling SOAPFaultException subCodes  and reasonLocales only for SOAP 1.2 and skipping those in case of SOAP 1.1.

https://issues.apache.org/jira/browse/CXF-7453


dkulp:
Note:  in the future, please submit pull requests based on master instead of some other branch.

PR #297, opened 2017-07-26
[CXF-7455] Tolerate missing message part

semancik:


deki:
Thanks for your pull request. Can you provide a unit/ integration test for this?

semancik:
Looks like this has been already worked around in the winrm4j project. I was not aware of that. So I might actually cancel the pull request later on. Please, bear with me, I need some time to explore. However, I would say that it is rather sad that it is easier to work around the problem than to deal with all this checkstyle pain, unit tests and so on.

deki:
No need to cancel the PR, if we could make the code more robust, we should do that.

semancik:
I was looking around how to construct a test for this. But I'm really puzzled. There seems to be very little documentation about writing tests. And I could not find any test that would be similar to what I'm trying to do.

I have SOAP message that is causing the exception. It is a SOAP response from the server. How do I create a simple test that just attempts to process the message, runs the interceptors and checks the processed message?

deki:
Ok could you please attach the SOAP message to the JIRA issue?

Regarding test you can take a look at the existing org.apache.cxf.jaxws.holder.HolderTest.

Btw. PRs should usually point to master, we will take care of merging them to the other branches.

deki:
@semancik any chance to get the SOAP message that reproduces the issue?

semancik:
I'm sorry. I was lucky enough not to be forced to deal with that terrible Microsoft technology in the meantime. And I'm quite scared to look there just by myself unless I really have to. Please bear with me. But I'll do it ... sooner or later.

gnodet:
Closing: this PR is 8+ years old with no test case provided (as discussed). The bug is still valid — will be addressed in a fresh PR with a proper unit test. CXF-7455.

PR #300, opened 2017-07-30
CXF-7371: use objectweb-asm to extract parameter-names from the debug-info of the service-class

bbodnar:
Suggested fix for https://issues.apache.org/jira/browse/CXF-7371.


dkulp:
This creates a hard dependency on a particular ASM implementation and as such is not something we will merge.   You will likely need to update the org.apache.cxf.common.util.ASMHelper class to support the ClassReader (in addition to the current ClassWriter stuff) and make sure all the asm calls go through there.

bbodnar:
OK, I will do so.

bbodnar:
There is no good solution for reading classes with ASM without direct coding against concrete ASM-classes (the "normal" (visitor-based) way isn't usable with reflection/proxying, because the visitors are in ASM >v3 classes (no longer interfaces); the non-visitor-based ways (tree, SAX-events) require optional ASM-libraries, whose existence must here surely not be assumed).
I see under this circumstances the only reasonable solution to extend the current (self-made) byte-code-parser to handle the Java8-constants.

PR #301, opened 2017-08-01
[CXF-7462] Minor improvements to OutboundSseEventImpl

andymc12:
1) Make the fields final to reflect that the event is immutable.
2) Use defaults for the data type (String.class) and media type
(SERVER_SENT_EVENT_TYPE).
3) Restrict the constructor's visibility.

New tests.

### DIFF
--- a/rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/OutboundSseEventImpl.java
+++ b/rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/OutboundSseEventImpl.java
@@ -24,24 +24,24 @@ import javax.ws.rs.core.GenericType;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.sse.OutboundSseEvent;
 
-public class OutboundSseEventImpl implements OutboundSseEvent {
-    private String id;
-    private String name;
-    private String comment;
-    private long reconnectDelay = -1;
-    private Class<?> type;
-    private Type genericType;
-    private MediaType mediaType;
-    private Object data;
+public final class OutboundSseEventImpl implements OutboundSseEvent {
+    private final String id;
+    private final String name;
+    private final String comment;
+    private final long reconnectDelay;
+    private final Class<?> type;
+    private final Type genericType;
+    private final MediaType mediaType;
+    private final Object data;
 
     public static class BuilderImpl implements Builder {
         private String id;
         private String name;
         private String comment;
         private long reconnectDelay = -1;
-        private Class<?> type;
+        private Class<?> type = String.class;
         private Type genericType;
-        private MediaType mediaType;
+        private MediaType mediaType = MediaType.SERVER_SENT_EVENTS_TYPE;
         private Object data;
 
         @Override
@@ -112,7 +112,7 @@ public class OutboundSseEventImpl implements OutboundSseEvent {
 
     }
     //CHECKSTYLE:OFF
-    OutboundSseEventImpl(String id, String name, String comment, long reconnectDelay,
+    private OutboundSseEventImpl(String id, String name, String comment, long reconnectDelay,
             Class<?> type, Type genericType, MediaType mediaType, Object data) {
         this.id = id;
         this.name = name;
--- a/rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/OutboundSseEventImpl.java
+++ b/rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/OutboundSseEventImpl.java
@@ -39,9 +39,9 @@ public final class OutboundSseEventImpl implements OutboundSseEvent {
         private String name;
         private String comment;
         private long reconnectDelay = -1;
-        private Class<?> type = String.class;
+        private Class<?> type;
         private Type genericType;
-        private MediaType mediaType = MediaType.SERVER_SENT_EVENTS_TYPE;
+        private MediaType mediaType = MediaType.TEXT_PLAIN_TYPE;
         private Object data;
 
         @Override
@@ -77,6 +77,9 @@ public final class OutboundSseEventImpl implements OutboundSseEvent {
         @Override
         @SuppressWarnings("rawtypes")
         public Builder data(Class newType, Object newData) {
+            if (newType == null || newData == null) {
+                throw new IllegalArgumentException("Parameters 'type' and 'data' must not be null.");
+            }
             this.type = newType;
             this.data = newData;
             return this;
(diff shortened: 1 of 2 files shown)


rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/OutboundSseEventImpl.java
```diff
@@ -24,24 +24,24 @@
...
     public static class BuilderImpl implements Builder {
         private String id;
         private String name;
         private String comment;
         private long reconnectDelay = -1;
-        private Class<?> type;
+        private Class<?> type = String.class;
         private Type genericType;
-        private MediaType mediaType;
+        private MediaType mediaType = MediaType.SERVER_SENT_EVENTS_TYPE;
```
reta: The `mediaType` is not `text/event-stream`, this is actually a media type of associated event payload (`data`) so to which body writer to use.

rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/OutboundSseEventImpl.java
```diff
@@ -24,24 +24,24 @@
...
+    private final MediaType mediaType;
+    private final Object data;
 
     public static class BuilderImpl implements Builder {
         private String id;
         private String name;
         private String comment;
         private long reconnectDelay = -1;
-        private Class<?> type;
+        private Class<?> type = String.class;
```
reta: I would not make any assumptions about default type here, better to be explicit.

reta:
Thanks for the improvements, @andymc12! Just a couple of minor comments before merging. Thanks!

rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/OutboundSseEventImpl.java
andymc12: I get NPEs when writing events via the `OutboundSseEventBodyWriter` if the user does not explicitly set the MediaType, so we will need some sort of default MediaType - the spec does not require the user to specify one.  `MediaType.SERVER_SENT_EVENTS_TYPE` (i.e. `text/event-stream`) seems to work for me, but I can try another type.  Would you suggest `text/plain`?

reta: I will take a look, thanks for spotting this gap.

rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/OutboundSseEventImpl.java
andymc12: The use case here is for when a user uses the `data(String data)` API method (as opposed to the `data(Class<?> type, Object o)` method.  In the single parm case, the type is not set by the user, so we will get a NPE unless we set it - and since the API method accepts a String type, I think that makes the most sense.

That said, I think I will make a change here - to avoid the case where the user uses _both_ API methods.  For example, they could do:
`builder.data(Widget.class, new Widget(123)).data("my String");`

I believe that this will cause the type to be set to Widget in the first call, but then not reset to String in the second call.  I'll change the defaulting to be in the implementation of the `data(String s)` method.

Thanks for the review comments!

andymc12: Hmm... I must've been looking at an older version of the javadoc.  It looks like there is no `data(String s)` method - it was changed to `data(Object o)`.  I'll change the code to use the type of the passed-in object if the class is not specified by the user.

Thanks again.

rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/OutboundSseEventImpl.java
```diff
@@ -92,6 +92,7 @@ public Builder data(GenericType newType, Object newData) {
 
         @Override
         public Builder data(Object newData) {
+            this.type = newData.getClass();
```
reta: Please check for `null` here

andymc12: Good catch!  Do you have any thoughts on what we should do if a null is passed in?  The spec throws an `IllegalArgumentException` if the user passes null to `sse.newEvent(String data)`.  For `builder.data(null)`, we could either do the same, or take null as the data and use Object.class as the default type.  I'm leaning toward throwing the exception.  What do you think?

andymc12:
Andriy, I went ahead with throwing IllegalArgumentExceptions when the user passes in null as either the data or type parameters.  This is more consistent with the API in the `javax.ws.rs.sse.Sse` interface.  If you disagree, please let me know.  I'll plan to leave this PR open until tomorrow.

Thanks again for the thorough review!

rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/OutboundSseEventImpl.java
sberyozkin: Hi Andriy, should we have a default text/plain for String/Integer/etc, application/json otherwise ?

reta: Agree, if we need to pick the default, `text/plain` would be the best one I think. Not sure about trying to determine if `application/json` should be used, I think the application developer should signal that, much like with `@Produces` annotation.

rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/OutboundSseEventImpl.java
```diff
@@ -41,7 +41,7 @@
         private long reconnectDelay = -1;
         private Class<?> type;
         private Type genericType;
-        private MediaType mediaType;
+        private MediaType mediaType = MediaType.SERVER_SENT_EVENTS_TYPE;
```
reta: I would agree with @sberyozkin, `mediaType = MediaType.TEXT_PLAIN_TYPE` would be the safest option, but definitely not `MediaType.SERVER_SENT_EVENTS_TYPE`

andymc12:
@reta @sberyozkin I've updated the code again - here is the rough summary of changes:
1) Default the event's _type_ to be `data.getClass()`
2) Throw an IllegalArgumentException if `data` is null.  This is consistent with the spec's `Sse.newEvent(null)` method.
3) Default the event's _MediaType_ to be `MediaType.TEXT_PLAIN_TYPE` - this avoids NPEs that I was seeing in the `OutboundSseEventBodyWriter` class.
4) Tests for all of these cases.  

I'm going on vacation tomorrow, so hopefully I can get this merged today.  Can you two give the code one last review?  

Thanks again!
Andy

reta:
Thanks, @andymc12! It's in master, have a great vacation!

andymc12:
Thanks @reta!

PR #305, opened 2017-08-09
Fix of non thread safe map reqHeaders

vbnhfylbh:
SOAP_ACTION is shared between different requests despite thread.local.request.context=true. Reproduced under load of 100 transactions per second.

PR #307, opened 2017-08-29
Fix of non thread safe map reqHeaders - SOAP_ACTION is shared between…

vbnhfylbh:
… different requests despite thread.local.request.context=true. Reproduced under load of 100 transactions per second.


vbnhfylbh:
Same as https://github.com/apache/cxf/pull/305 but with rebased branch. @deki, please review

deki:
I've applied the changes, thank you.

PR #308, opened 2017-08-29
CXF-7489: Stax2Validation doesn't support schema imports.

facundovs:
Using the Stax2Validation for validating a wsdl, a NPE is being thown.
After some analysis I could find that the targetNamespace is being used as key to save schemas to process (please see this[ code line](https://github.com/apache/cxf/blob/master/core/src/main/java/org/apache/cxf/staxutils/validation/Stax2ValidationUtils.java#L153))
However, a WSDL can have the following format in the <wsdl:types> element:
```
<wsdl:types>
    <xsd:schema>
            <xsd:import namespace="http://www.foo.org" schemaLocation="foo.xsd"/>
            <xsd:import namespace="http://www.bar.org" schemaLocation="bar.xsd"/>
    </xsd:schema>
</wsdl:types>
```
As you can see in these cases, the targetNamespace is not present in the schema element that is only used for importing external schemas.

### DIFF
--- a/core/src/main/java/org/apache/cxf/staxutils/validation/Stax2ValidationUtils.java
+++ b/core/src/main/java/org/apache/cxf/staxutils/validation/Stax2ValidationUtils.java
@@ -25,10 +25,13 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import javax.xml.XMLConstants;
+import javax.xml.parsers.DocumentBuilder;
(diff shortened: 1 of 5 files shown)


facundovs:
Hey guys. Why do you close this PR?

dkulp:
Because it's been merged to master.

facundovs:
Great! Thank you very much.

PR #311, opened 2017-09-05
[STS] Added switch to disable the generation of a Lifetime element in a cla…

vpopescu:
Added a boolean to disable the generation of the Lifetime element in the STS response.  Default is true (don't disable) to match existing behaviour.

Discussed here by a team member: http://cxf.547215.n5.nabble.com/STS-removing-Lifetime-from-response-tc5782461.html


vpopescu:
Not sure I understand the error. Did I do something wrong?

PR #312, opened 2017-09-15
CVE-2017-3156 backport

jgallimore:
This is a backport of the patch from here: e66ce235 to 2.6.x. Although I'm opening the PR, I do understand that 2.6.x is EOL, and therefore this might not be merged or ultimately released. If anyone does feedback, it will be very gratefully received.


coheigea:
I think the last 2.6.x security release was a once-off and so I don't think there is any point in merging anything to the branch.

PR #313, opened 2017-09-18
[CXF-7493] Add a default ClassUnwrapper for CDI integration.

johnament:
This is to avoid issues with proxies and class metadata.

### DIFF
--- a/integration/cdi/src/main/java/org/apache/cxf/cdi/CdiBusBean.java
+++ b/integration/cdi/src/main/java/org/apache/cxf/cdi/CdiBusBean.java
@@ -28,6 +28,8 @@ import javax.enterprise.inject.spi.InjectionTarget;
 import org.apache.cxf.Bus;
 import org.apache.cxf.BusFactory;
 import org.apache.cxf.bus.extension.ExtensionManagerBus;
+import org.apache.cxf.common.util.ClassUnwrapper;
+import org.apache.cxf.common.util.SystemPropertyAction;
 
 final class CdiBusBean extends AbstractCXFBean< ExtensionManagerBus > {
     static final String CXF = "cxf";
@@ -63,6 +65,9 @@ final class CdiBusBean extends AbstractCXFBean< ExtensionManagerBus > {
     @Override
     public ExtensionManagerBus create(final CreationalContext< ExtensionManagerBus > ctx) {
         final ExtensionManagerBus instance = injectionTarget.produce(ctx);
+        if ("true".equals(SystemPropertyAction.getProperty("org.apache.cxf.cdi.unwrap.proxies", "true"))) {
+            instance.setProperty(ClassUnwrapper.class.getName(), new CdiClassUnwrapper());
+        }
         BusFactory.possiblySetDefaultBus(instance);
         instance.initialize();
 
--- a/integration/cdi/src/main/java/org/apache/cxf/cdi/CdiClassUnwrapper.java
+++ b/integration/cdi/src/main/java/org/apache/cxf/cdi/CdiClassUnwrapper.java
@@ -1,0 +1,40 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.cxf.cdi;
+
+import java.util.regex.Pattern;
+
+import org.apache.cxf.common.util.ClassUnwrapper;
+
+class CdiClassUnwrapper implements ClassUnwrapper {
+    private static final Pattern PROXY_PATTERN = Pattern.compile(".+\\$\\$.+Proxy");
+
+    CdiClassUnwrapper() {
+
+    }
+
+    @Override
+    public Class<?> getRealClass(Object o) {
+        Class<?> clazz = o.getClass();
+        if (PROXY_PATTERN.matcher(clazz.getSimpleName()).matches()) {
+            return clazz.getSuperclass();
+        }
+        return clazz;
+    }
+}
--- a/systests/cdi/base/src/main/java/org/apache/cxf/systests/cdi/base/AbstractCdiSingleAppTest.java
+++ b/systests/cdi/base/src/main/java/org/apache/cxf/systests/cdi/base/AbstractCdiSingleAppTest.java
@@ -56,6 +56,24 @@ public abstract class AbstractCdiSingleAppTest extends AbstractBusClientServerTe
     }
 
     @Test
+    public void testResponseHasBeenReceivedWhenQueringBooksById() {
+        final String id = UUID.randomUUID().toString();
+
+        Response r = createWebClient(getBasePath() + "/books").post(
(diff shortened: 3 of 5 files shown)


integration/cdi/src/main/java/org/apache/cxf/cdi/CdiClassUnwrapper.java
```diff
@@ -0,0 +1,42 @@
...
+ * under the License.
+ */
+package org.apache.cxf.cdi;
+
+import java.util.List;
+import static java.util.Arrays.asList;
+import org.apache.cxf.common.util.ClassUnwrapper;
+
+class CdiClassUnwrapper implements ClassUnwrapper {
+    private static final List<String> UNWRAPPED_MARKERS = asList("OwbNormalScopeProxy", "WeldClientProxy");
```
rmannibucau: OWB has other suffixes, why not using $$ as in https://github.com/apache/meecrowave/blob/7bd58d4a254c0d40fb1a2ab0d19b24dbab4014ca/meecrowave-core/src/main/java/org/apache/meecrowave/cxf/MeecrowaveBus.java#L54 ? Would also handle TomEE proxies (EJB) BTW. Also any way it just gets merge with the default unwrapper? I don't see any reason to not have it out of the box since it doesn't depend on anything.

johnament: I have only seen the one suffix from OWB.  Where can I find the list of suffixes?  I saw what you did in TomEE, and I think using `$$` could be problematic for non-CDI frameworks, so I wanted to explicitly list out the known suffixes.

I'm fine with putting something like this in core, and explicitly activating it in CDI, but would be concerned with being over eager with the swap out.

rmannibucau: For owb you need to check AbstractProxyFactory subclasses out, without more checking I had in mind $$OwbSubClass. In TomEE we also have $$LocalBeanProxy which would break probably, and a lot of frameworks use $$ for that purpose. That is why we use $$ and not something more precise which will always hit a pitfall. Do you know any framework using $$ and would break cxf unwrapping the class? If so we can surely just make the suffixes configurable. Throughts?

reta: @johnament @rmannibucau Was going through this discussion (https://issues.jboss.org/browse/CDI-10), seems to be a known problem people struggle with. Not sure if it is better or not, what if we detect the CDI implementation at runtime (class probing) and delegate the unwrapping to the framework itself? F.e. in case of OWB, we could use`NormalScopeProxyFactory.unwrapInstance(o)`, something similar should be available for Weld as well. Not ideal either, but more reliable may be? Thoughts?

johnament: Well, I don't think Weld has a behavior like `unwrapInstance` but it could be done by us if its Weld.  I also figured that checking by name would be easier since it avoid any compile time dependencies.

reta: It is definitely the easiest / simplest way to do that, not sure how reliable it is though ... May be we could have some regex, like `^.+$$.+Proxy$`, the `$$` usually depict a generated class, so we would cover `Xxx$$OwbNormalScopeProxy`, `Xxx$$WeldClientProxy` as well as `Xxx$$LocalBeanProxy` like @rmannibucau  pointed out as well. Should be pretty good heuristic, what do you think?

integration/cdi/src/main/java/org/apache/cxf/cdi/CdiClassUnwrapper.java
```diff
@@ -0,0 +1,40 @@
...
+ * under the License.
+ */
+package org.apache.cxf.cdi;
+
+import java.util.regex.Pattern;
+
+import org.apache.cxf.common.util.ClassUnwrapper;
+
+class CdiClassUnwrapper implements ClassUnwrapper {
+    private static final Pattern PROXY_PATTERN = Pattern.compile(".+\\$\\$.+Proxy");
```
johnament: @reta @rmannibucau ok, this uses a pattern based on the discussion.  Do we still want to move outside the CDI module?

reta: Thanks @johnament, I don't think it should be moved outside CDI integration, @rmannibucau any particular reasons to do that?

rmannibucau: Yep the reason is to ensure it is used by all integrations. Very few server uses cdi extension to comply to spec but all use proxies. Since the perf and behavior shouldnt hurt it can be added to the default IMO

sberyozkin: Hi guys, but the problem is only showing itself on the CDI path ?

rmannibucau: Not really. it has nothing to do with CDI actually except CDI uses proxies. It will happen each time a lib is using subclassing proxies, CDI is one big consumer but there are a lot of other ways to get it even in plain standalone.

sberyozkin: Possibly :-) but Spring users are not opening any issues...

sberyozkin: Oh, indeed, support for Spring is there, you are right; I'm just not sure the proposed ClassUnwrapper should be called there by default every time ClassHelper is called, may be do it in 2 stages, 1) test as part of the CDI 2) move it to the core ?

rmannibucau: Works for me

integration/cdi/src/main/java/org/apache/cxf/cdi/CdiBusBean.java
```diff
@@ -63,6 +65,9 @@ public String getName() {
     @Override
     public ExtensionManagerBus create(final CreationalContext< ExtensionManagerBus > ctx) {
         final ExtensionManagerBus instance = injectionTarget.produce(ctx);
+        if ("true".equals(SystemPropertyAction.getProperty("org.apache.cxf.cdi.useCdiUnwrapper", "true"))) {
```
reta: @johnament pretty minor, before merging could you please rename property `org.apache.cxf.cdi.useCdiUnwrapper` to `org.apache.cxf.cdi.unwrap.proxies`  (or similar) so it kind of explains its purpose.

integration/cdi/src/main/java/org/apache/cxf/cdi/CdiClassUnwrapper.java
reta: Sounds like a good plan, @sberyozkin are we good to merge?

integration/cdi/src/main/java/org/apache/cxf/cdi/CdiBusBean.java
johnament: Sure

integration/cdi/src/main/java/org/apache/cxf/cdi/CdiClassUnwrapper.java
johnament: It's not clear what the plan is, just leave it?

sberyozkin: Hi Andriy, if all 3 of you are happy, then yes, please merge :-), I did not really contribute myself

reta: @johnament Sorry, was not clear, I was referring to @sberyozkin suggestion for 2 stages, let us start from CDI first, and than make a decision if it worth going to core or not. Make sense?

johnament:
Sure sounds good.  Feel free to merge if good.

reta:
Thanks @johnament !

PR #315, opened 2017-09-22
[CXF-7512] Support searching swaggerUi by name and version in all bun…

cschneider:
…dles

### DIFF
--- a/rt/rs/description-swagger/src/main/java/org/apache/cxf/jaxrs/swagger/SwaggerUiResolver.java
+++ b/rt/rs/description-swagger/src/main/java/org/apache/cxf/jaxrs/swagger/SwaggerUiResolver.java
@@ -22,26 +22,13 @@ import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.Enumeration;
 
-public class SwaggerUiResolver {
+public final class SwaggerUiResolver {
     static final String UI_RESOURCES_ROOT_START = "META-INF/resources/webjars/swagger-ui/";
-    
-    static final SwaggerUiResolver HELPER;
-    static {
-        SwaggerUiResolver theHelper = null;
-        try {
-            theHelper = new OsgiSwaggerUiResolver();
-        } catch (Throwable ex) {
-            theHelper = new SwaggerUiResolver();
-        }
-        HELPER = theHelper;
+
+    private SwaggerUiResolver() {
     }
 
-
-    protected SwaggerUiResolver() {
-    }
-
-    protected String findSwaggerUiRootInternal(String swaggerUiMavenGroupAndArtifact,
-                                               String swaggerUiVersion) {
+    private String findSwaggerUiRootInternal(String swaggerUiVersion) {
         try {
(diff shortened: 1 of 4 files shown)


sberyozkin:
Hi Christian, many thanks. 
The original pattern looks cleaner to me (we apply it in various places in CXF - proxies - CGLIB or not, and Class resolution and scanning, probably in other places too), as opposed to having the code which is not relevant for the non-OSGI case which will guarantee 2 Throwable will always be thrown and caught and a non-reachable code in the end in the OSGI case if the bundle is installed and of no use - if it's not installed (it checks URLClassLoader), and all of that 'bundled' in a single function body. 

I'd like to propose that we deprecate 'swaggerUiMavenGroupAndArtifact' and have you new 'swaggerUiName' property being the only property used by the resolvers. We'd be able to use this property in the non-OSGI case too (to find the correct jar if needed). In OSGI case, as I suggested earlier, we just check if the name contains '/' then it is Maven otherwise it is a bundle name only. See many people know the Maven artifacts but finding the bundle name requires more work so if people use Maven artifacts it is just simpler for then to use the artifacts, 'swagger-ui' is not a very uniique value to be honest...Anyway, we'd still be able to use a single OSGI Resolver and be able to skip the Maven related code in the short term if the user sets a swaggerUiName property without '/'

sberyozkin:
Would you be OK I applied your name-based resolution code per the above comment for you to see how it will look like ? You may end up liking it :-)

cschneider:
Hi Sergey. Feel free to edit the code. Please remember though the swaggerUiName is not a bundle name. It is a path below META-INF/resources/webjar. I chose this approach as it can work in OSGi as well as outside.  I think the swaggerUiName should only mean this. 
If you want to still use maven coordinates then I suggest to keep using the existing swaggerUiMavenGroupAndArtifact. 

Christian

sberyozkin:
Hey Christian, thanks, I'll play with it next week. Enjoy a well-deserved break :-)

palmpatrick:
Hi @cschneider, @sberyozkin 

I'm using CXF-DOSGi to host and consume both SOAP and REST services within a plain Apache Felix OSGi distribution.
It works flawlessly, except for the swagger UI.

What is the estimate on getting this fix reintegrated to CXF?

rt/rs/description-swagger/src/main/java/org/apache/cxf/jaxrs/swagger/SwaggerUiResolver.java
```diff
@@ -87,12 +74,22 @@ protected static String checkUiRoot(String urlStr, String swaggerUiVersion) {
...
-                                                                     swaggerUiVersion);
+                                           String swaggerUiName, String swaggerUiVersion) {
+        String root = null;
+        try {
+            root = new OsgiSwaggerUiResolver()
+                .findSwaggerUiRoot(swaggerUiMavenGroupAndArtifact, swaggerUiVersion);
+        } catch (Throwable e) {
+            // Ignore
+        }
+        try {
```
reta: Should it be wrapped into `if (root == null) { ... }` block (if previous lookup succeeded)?

PR #316, opened 2017-09-27
[CXF-7519] The cxf-rt-transports-http-jetty should not rely on the ol…

gnodet:
…d blueprint compatibility bundle


gnodet:
Retest this please.

PR #318, opened 2017-10-05
CXF-7506: Upgrade guava

davidkarlsen:


deki:
Thanks for your PR. Have you checked for OSGi compatibility? we had several conflicts in the past...

davidkarlsen:
No - I did not - and the build broke - downloading the internet and running locally now...

sberyozkin:
Hi, why is this update needed, we usually update guava if some of the CXF features indirectly needs it, example, CXF needs to update a newer swagger-core/etc which requires newer guava

davidkarlsen:
AFAIK it is not needed - I just found the JIRA issue sticking around for the target release.

sberyozkin:
Thanks; yes, IMHO it would make sense to postpone until the newer Guava version is actually needed

davidkarlsen:
Agreed - close the jira and abandon this.

PR #321, opened 2017-10-18
Check @Priority annotations on providers when sorting

andymc12:
Per section 4.1.3 of the JAX-RS 2.1 spec, entity providers must be
selected/sorted based on the priority annotation.

This PR handles MBRs and MBWs.  I think we will also need to handle ContextResolvers and ExceptionMappers, but I wanted to get the ball rolling with this.


rt/frontend/jaxrs/src/test/java/org/apache/cxf/jaxrs/provider/ProviderFactoryTest.java
```diff
@@ -820,6 +822,40 @@ public void testSetSchemasFromClasspath() {
...
+        providers.add(highPriorityBookHandler);
+        pf.setUserProviders(providers);
+        assertSame(highPriorityBookHandler,
+                   pf.createMessageBodyReader(Book.class, Book.class, new Annotation[]{},
+                                              MediaType.APPLICATION_XML_TYPE, new MessageImpl()));
+        assertSame(highPriorityBookHandler,
+                   pf.createMessageBodyWriter(Book.class, Book.class, new Annotation[]{},
+                                              MediaType.APPLICATION_XML_TYPE, new MessageImpl()));
+        
+        // do the same thing again but add the providers in the reverse order to ensure add order
```
reta: Would you mind please to extract the second part of the test case to separate method, for example `public void testSortByPriorityReversed()`? Thanks.

reta:
Looks pretty good for MBRs and MBWs, just a minor comment to have a bit cleaner test cases. Otherwise could go in.

andymc12:
Thanks @reta! I updated the test case as you suggested.  I also added another commit for handling ContextResolvers and new tests.  I'll try to get ExceptionMappers done tomorrow.  Thanks for the review,

Andy

PR #323, opened 2017-10-20
Handle @Priority on ExceptionMappers

andymc12:
Like PR #321, we need to handle the `@Priority` annotation on exception mappers.

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/provider/ProviderFactory.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/provider/ProviderFactory.java
@@ -858,7 +858,7 @@ public abstract class ProviderFactory {
         return result;
     }
 
-    private static int comparePriorityStatus(Class<?> cl1, Class<?> cl2) {
+    static int comparePriorityStatus(Class<?> cl1, Class<?> cl2) {
         Integer value1 = AnnotationUtils.getBindingPriority(cl1);
         Integer value2 = AnnotationUtils.getBindingPriority(cl2);
         return value1.compareTo(value2);
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/provider/ServerProviderFactory.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/provider/ServerProviderFactory.java
@@ -598,7 +598,11 @@ public final class ServerProviderFactory extends ProviderFactory {
                     return -1;
                 }
             }
-            return super.compare(p1, p2);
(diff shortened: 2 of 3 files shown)


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/provider/ProviderFactory.java
```diff
@@ -858,7 +858,7 @@ private static int compareCustomStatus(ProviderInfo<?> p1, ProviderInfo<?> p2) {
         return result;
     }
 
-    private static int comparePriorityStatus(Class<?> cl1, Class<?> cl2) {
+    static int comparePriorityStatus(Class<?> cl1, Class<?> cl2) {
```
johnament: maybe `protected`?

andymc12: I prefer to limit visibility as much as possible.  In this case, the only caller of the `comparePriorityStatus` method is in the same package, so default/package visibility is more limiting than `protected` since it does not allow visibility to sub-classes outside of this package.  If other classes (sub-classes or otherwise) need to call this method, then I think we could change the visibility to `protected` or `public` if necessary.  Thanks for the review.

johnament: Hmmm but if I have a custom impl of this in a different package (now required in java 9) I wouldn't be able to use this utility.

andymc12: I must be misunderstanding something.  This is a new method that I added last week - originally at private visibility because only `ProviderFactory` needed to access it - and to be consistent with the `compareCustomStatus` method directly above it.  I removed the `private` modifier so that `ServerProviderFactory` (subclass in the same package) could also access it - to sort ExceptionMappers.

Does some other code need to access this method?  The unit tests do not call it directly (thus they don't need visibility), but the method is covered by tests that verify the proper sorting of different types of providers.  There shouldn't be anything else that is currently calling this method since it is so new.  If future work in a different package needs to call this method, then that work should be free to change the visibility as necessary.  Am I missing some other use case?  Thanks again.

johnament: You're correct, I'm thinking a bit ahead.

I'm thinking of use cases where the underlying `ProviderFactory` used in the code base either isn't one provided by CXF or provided by CXF but in a different module.  In those cases, it would be good to be able to reuse this logic instead of having to replicate it.

PR #324, opened 2017-10-22
RFC introduced: Supress generated date switch for wsdl2java

typekpb:
Introduced the switch: `-supress-generated-date` which enables using wsdl2java without writing the current timestamp in the generated files.

The advantage is shown once generated files are stored in source control and changes are easily diffable (as timestamps are not present => changes don't come with each and every build).

However this PR is incomplete, as `CodeGenOptionTest.testSupressGeneratedDatePresentOption()` is still failing. Any hints leading to solution would be appreciated, I just could not identify the missing pieces to be modified (to have the new switch considered).

I seem to be not the 1.st one needing it, see: http://cxf.547215.n5.nabble.com/wsdl2java-suppressing-timestamps-in-Javadoc-comments-td5731702.html


typekpb:
@deki thanks for feedback! I went all the way down to xjc and found:
* `@Generated` annotation being present for `-mark-generated` only case: https://github.com/javaee/jaxb-v2/blob/f1ef0676987b83b134b8c97fe99083423f7f3311/jaxb-ri/xjc/src/main/java/com/sun/tools/xjc/addon/at_generated/PluginImpl.java
* as well as propagation code: https://github.com/apache/cxf/blob/13bcc2362509c6d893ed21fff9461d576fda1ccc/tools/wsdlto/databinding/jaxb/src/main/java/org/apache/cxf/tools/wsdlto/databinding/jaxb/JAXBDataBinding.java#L430

 I addressed the issues:
* supress -> suppress
* moreover `@Generated` annotation is no longer affected by `-suppress-generated-date` cmd line option, as there is another switch: `mark-generated`, which also gets propagated all the way to xjc as well (otherwise xjc code would need updating as well)

typekpb:
@deki build failed: https://builds.apache.org/job/CXF-Trunk-PR/547 however I don't see how I could have caused it by my changes

deki:
Don't worry about the failing build, unfortunately this happens from time to time.

If I understand it correctly, the option now only partially suppresses the generated date?

typekpb:
Exactly. As @Generated annotation spreads across the projects and is not
generated implicitly, I just tried to keep the effort low and fulfill my
use case.

deki:
I don't mean the annotation. The date ifself is still part of the classes generated by xjc, isn't it?

typekpb:
Well, is it? Not sure

tools/wsdlto/test/src/test/java/org/apache/cxf/tools/wsdlto/jaxws/CodeGenOptionTest.java
```diff
@@ -294,6 +294,104 @@ private int countGeneratedAnnotations(String str) {
...
+                    return true;
+                }
+            }
+            expectDate = line.contains("This class was generated by");
+        }
+
+        return false;
+    }
+
+    private boolean generatedAnnotationPresent(String str) {
```
deki: This method is unused.

tools/wsdlto/databinding/jaxb/src/main/java/org/apache/cxf/tools/wsdlto/databinding/jaxb/JAXBDataBinding.java
```diff
@@ -437,24 +437,20 @@ public void initialize(ToolContext c) throws ToolException {
...
         }
-        addSchemas(opts, schemaCompiler, schemas);
-        addBindingFiles(opts, jaxbBindings, schemas);
-
 
-        for (String ns : context.getNamespacePackageMap().keySet()) {
-            File file = JAXBUtils.getPackageMappingSchemaBindingFile(ns, context.mapPackageName(ns));
+        if (context.optionSet(ToolConstants.CFG_SUPPRESS_GENERATED_DATE)) {
+            // Prevents dumping current date as part of javadocs of the Java files generated.
+            // This is done by passing '-suppress-generated-date' attribute to jaxb xjc.
```
deki: XJC has no such option, why should we pass it? After removing this codeblock the test still passes.

deki:
Ok now I get it, you leave out `-mark-generated` and set `-suppress-generated-date` to achieve it, right?

tools/wsdlto/test/src/test/java/org/apache/cxf/tools/wsdlto/jaxws/CodeGenOptionTest.java
typekpb: fixed, thanks

tools/wsdlto/databinding/jaxb/src/main/java/org/apache/cxf/tools/wsdlto/databinding/jaxb/JAXBDataBinding.java
typekpb: fixed, thanks

typekpb:
exactly!

typekpb:
@deki any chance to get this one merged and included to any near-future release?

deki:
Sorry I'm quite busy at the moment. There are no plans for 3.2.2 yet, I try to review it before the next release.

deki:
Could you please create an issue in CXF-JIRA https://issues.apache.org/jira/secure/CreateIssue!default.jspa so that we have a reference for the release notes?

I'll squash your commits, remove the whitespace changes and merge it with reference to the JIRA issue afterwards.

typekpb:
@deki done, see: https://issues.apache.org/jira/browse/CXF-7618

deki:
Thank you, merged to master.

PR #329, opened 2017-10-27
CXF-7501: Cannot inject field in ContainerRequestFilter (and generally, into any providers registered using FeatureContext)

reta:
Adding the new extension-based customization (`ServerConfigurableFactory`) in order to control the way how providers are created and registered using `FeatureContext::register` methods family. It allows to use proper mechanisms in case of managed runtimes (like CDIs f.e.).

@sberyozkin @johnament Submitting through PR to have some feedback guys you may have. The change is not really complex but touches a bit some internals. 

Thanks.

### DIFF
--- a/integration/cdi/src/main/java/org/apache/cxf/cdi/CdiServerConfigurableFactory.java
+++ b/integration/cdi/src/main/java/org/apache/cxf/cdi/CdiServerConfigurableFactory.java
@@ -1,0 +1,85 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.cxf.cdi;
+
+import javax.enterprise.context.spi.CreationalContext;
+import javax.enterprise.inject.spi.AnnotatedType;
+import javax.enterprise.inject.spi.Bean;
+import javax.enterprise.inject.spi.BeanAttributes;
+import javax.enterprise.inject.spi.BeanManager;
+import javax.enterprise.inject.spi.InjectionTargetFactory;
+import javax.ws.rs.RuntimeType;
+import javax.ws.rs.core.Configurable;
+import javax.ws.rs.core.FeatureContext;
+
+import org.apache.cxf.jaxrs.impl.ConfigurableImpl;
+import org.apache.cxf.jaxrs.impl.ConfigurableImpl.Instantiator;
+import org.apache.cxf.jaxrs.provider.ServerConfigurableFactory;
+
+/** 
+ * Creates the instance of Configurable<?> suitable for CDI-managed runtime.
+ */
+public class CdiServerConfigurableFactory implements ServerConfigurableFactory {
+    private final BeanManager beanManager;
+    
+    CdiServerConfigurableFactory(final BeanManager beanManager) {
+        this.beanManager = beanManager;
+    }
+    
+    @Override
+    public Configurable<FeatureContext> create(FeatureContext context) {
+        return new CdiServerFeatureContextConfigurable(context, beanManager);
+    }
+    
+    /** 
+     * Instantiates the instance of the provider using CDI/BeanManager 
+     */
+    private static class CdiInstantiator implements Instantiator {
+        private final BeanManager beanManager;
+        
+        CdiInstantiator(final BeanManager beanManager) {
+            this.beanManager = beanManager;
+        }
+        
+        @Override
+        public <T> Object create(Class<T> cls) {
+            final AnnotatedType<T> annotatedType = beanManager.createAnnotatedType(cls);
+            final InjectionTargetFactory<T> injectionTargetFactory = 
+                beanManager.getInjectionTargetFactory(annotatedType);
+            final BeanAttributes<T> attributes = beanManager.createBeanAttributes(annotatedType);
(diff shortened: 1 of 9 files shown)


integration/cdi/src/main/java/org/apache/cxf/cdi/CdiServerConfigurableFactory.java
```diff
@@ -0,0 +1,85 @@
...
+     */
+    private static class CdiInstantiator implements Instantiator {
+        private final BeanManager beanManager;
+        
+        CdiInstantiator(final BeanManager beanManager) {
+            this.beanManager = beanManager;
+        }
+        
+        @Override
+        public <T> Object create(Class<T> cls) {
```
johnament: I see we have a create... should we also have a destroy?  Otherwise, in CDI, you run the risk of having dependent scoped beans lying around.  I'm not sure if these are meant to be per-request objects or global objects, but if they are per request and `@Dependent` scoped then you'll want to destroy them.

reta: Was puzzled to have this as well but there is no lifecycle hook to call the destroy /dispose :(

rmannibucau: the client has close() so we can - worse case - use that. Also if the beanManager.isNormalScope(bean.getScope()) is true we don't need to call context.release()

reta: This is server-side ... :)

rmannibucau: -> bam :)

the server has it too (destroy() from memory)

reta: @rmannibucau  let me try to figure out if we could hook in there, thanks for the hint

reta: @johnament @rmannibucau Haven't found the acceptable way to dispose the created beans. It would be good to have that however `Feature`s (and `DynamicFeature`s) are instantiated at deployment (when JAX-RS resources are discovered) hereby should not be subject of request scoped but only global one. When server is shutting down (destroy) everything will be gone so may not worth introducing complex disposal mechanisms. Makes sense?

rmannibucau: Issue is more for dependent beans which must be released to avoid to leak data/memory when the app is undeployed but the server is not killed so we should really have a lifecycle for the configurable which is bound to the underlying instance and worse case to the Server.

reta: Hard to disagree with that indeed, CXF should be a good citizen there. So I've came up with quite simple solution using CDI events. Whenever we create the context in such circumstances, we register it for later disposal inside `JAXRSCdiResourceExtension` by firing the event (the same way we manage all other managed beans we create inside the CDI extension). It seems to work perfectly fine and releases the contexts on shutdown. Thoughts?

rmannibucau: Looks good for the server side :)

reta: Thanks guys for the comments, merging ...

johnament: Hmmm I just noticed something looking at another issue.  I think this may assume that all providers now registered are within a CDI bean archive.  We may want to do a fallback in case the bean cannot be located.

reta: Thanks for spotting that @johnament, I will address it shortly.

reta: @johnament hmm (my turn), so in the test case we have there are 2 providers: one from bean archive, another from regular JAR file (not a bean archive). In both cases, the bean is located successfully. Looking why, BeanManager implementation (OWB, Weld) does the fallback in case there is no bean archive for a class available.

rmannibucau: @reta if not a bean you will not get a Bean<?> instance doing the lookup (empty Set). If it is found it means the jar was scanned and the bean activated somehow. The issue can come from the test setup, did you try with jackson for instance or a plain custom-provider-nocdi.jar?

reta: I did with `AtomFeedProvider` from plain no CDI archive, let me try Jackson to be sure :)

reta: @johnament No issues, works just fine for Jackson as well (no deployment archives are present but the bean was synthesized from the class).

rmannibucau: this is because you don't lookup a cdi bean but create one, this is probably not what we want. We should lookup an existing bean or fallback on creating a plain instance (take care of context injection support ;)). If we don't we can have side effects in applications duplicating beans during lookup or just pollute the CDI context for no reason.

reta: so basically the better way would be to call beanManager.getBeans() to ensure there is one in the context, and fallback to createBean if none, right? :-)

johnament:
Could you double check with an explicitly @Vetoed provider?  I believe the
way you run the CDI teats it discovers all veans regardless of beans.xml.

reta:
@johnament yes, sure, I can add one @Vetoed guy. The tests do not hijack anything,  CDI container **does not** discover all the beans, the `bean.xml` presence is strictly honored.

integration/cdi/src/main/java/org/apache/cxf/cdi/CdiServerConfigurableFactory.java
rmannibucau: nop, try getbeans and if resolve() returns null then use standalone logic (default newInstance())

reta: sounds good, thanks @rmannibucau

sberyozkin:
Right, the fallback should be available

PR #330, opened 2017-10-29
CXF-7501: Cannot inject field in ContainerRequestFilter (and generally, into any providers registered using FeatureContext)

reta:
Using proper CDI bean resolution with fallback to default strategy. The `@Vetoed` presented an edge case when CDI runtime won't resolve the bean but we have to go over manual instantiation. 

CC @rmannibucau @johnament

### DIFF
--- a/integration/cdi/src/main/java/org/apache/cxf/cdi/CdiServerConfigurableFactory.java
+++ b/integration/cdi/src/main/java/org/apache/cxf/cdi/CdiServerConfigurableFactory.java
@@ -49,7 +49,8 @@ public class CdiServerConfigurableFactory implements ServerConfigurableFactory {
     }
     
     /** 
-     * Instantiates the instance of the provider using CDI/BeanManager 
+     * Instantiates the instance of the provider using CDI/BeanManager (or fall back
+     * to default strategy of CDI bean is not available).
      */
     private static class CdiInstantiator implements Instantiator {
         private final BeanManager beanManager;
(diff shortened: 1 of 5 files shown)


rmannibucau:
LGTM

side note: do we have a test covering @Context injection/instantiation too? Wonder if we need a third case where we do the injection normally (standalone) then create an InjectionTarget to integrate with CDI.

rmannibucau:
@ApplicationScoped the provider, will be proxied so the injection will not be done as expected ;). A workaround used can be to support @Inject @Context but that's a workaround but CDI doesn't enable us to do more yet until you add automatically an @AroundConstruct interceptor to do the injection.

reta:
@rmannibucau I see, proxies do mess things up. I think would be good to solve that as a separate issue, it has more broad impact that the one this PR concerns. I will create a test case to reproduce that and an issue in the tracker. Sounds like a plan?

rmannibucau:
+1, that's why it was a side note ;)

reta:
Done, https://issues.apache.org/jira/browse/CXF-7544. @rmannibucau @johnament Thank you guys for helping out with proper implementation.

johnament:
LGTM as well.  FWIW, I was able to cause it to fail using a CDI 2 feature of SE initialization.

```java
SeContainerInitializer.newInstance()
        .disableDiscovery()
        .addBeanClasses(Resources.class, LoggingReporter.class, BraveProducers.class, RestApplication.class, TopCDsEndpoint.class, CXFCdiServlet.class)
        .addExtensions(new JAXRSCdiResourceExtension())
        .initialize();
```

johnament:
In general, there are no issues using CXF with CDI 2, the integration continues to work fine.  This was just something I noticed and wanted to confirm as a real issue.

`@Vetoed` was the easiest way I had to explain where it would fail from a mobile

PR #334, opened 2017-11-07
[CXF-7549] test for bug of feature inside a feature

csierra:
Tests to highlight the bug described in https://issues.apache.org/jira/browse/CXF-7549


deki:
@sberyozkin looks like these tests went into ProviderFactoryTest, so this can be closed without merge right?

deki:
CXF-7549 is resolved and no other feedback was provided therefore closing...

PR #337, opened 2017-11-10
[CXF-7527] getMatchedURIs to avoid duplicate URIs

andymc12:
The UriInfo.getMatchedURIs method currently returns "duplicate" URI
entries when called from a method that is not annotated with an
@Path annotation. The "duplicate" URI is basically the same URI but
with a trailing slash. This was first discovered when using sub
resource locators, but could also occur in simpler scenarios (like
a resource method annotated with @GET but not @Path). 

When comparing against the javadoc and the RI, it looks like the
URIs we were returning started with a slash, where the RI and
javadoc examples do not. This commit removes the starting slash as
well.


andymc12:
Updated the systests to no longer check for the leading slash.

sberyozkin:
Interesting, was it not ever test in TCK given the leading slashes have always been there ?

andymc12:
It looks like "we got lucky".  There are tests in the TCK for getMatchedURIs, but it looks like it only checks for (1) is the name of the resource and the name of the path in the annotation on the method both in the path (but does not check for slashes) and (2) the size of the returned List (however every resource method in the tests also have the Path annotation, so all of the tested scenarios worked - had they tested for a case where there is no Path annotation, we would have failed).

PR #338, opened 2017-11-14
Update TypeVariable check ResourceUtils.java

evaristowb:
https://issues.apache.org/jira/browse/CXF-7545


sberyozkin:
Dennis, thanks :-) Evaristo, even though I just merged the previous update as is, without a test, Dennis is right, try copying 
https://github.com/apache/cxf/blob/master/rt/rs/description/src/test/java/org/apache/cxf/jaxrs/model/wadl/WadlGeneratorTest.java#L205

and following more or less the same path, in your case, with a single Java method only, it will be quite easy..., thanks

evaristowb:
Sorry my late, it follows UnitTests.

PR #339, opened 2017-11-14
CXF-7544: Support @Context-based injection into proxied CDI beans

reta:
The issue pop up as part of https://github.com/apache/cxf/pull/330 discussion. In case when provider / feature / resource is a proxied CDI bean, the contextual class members (annotated with @Context) are not injected.

More broadly,iIn some circumstances (like using `@ApplicationScoped` annotation for example) the CDI runtime will create a proxy class for a particular bean. As the result, the CXF side is going to bind the particular provider metadata to this proxy instance. It looks logical and unambiguous.

However, the interesting things are happening when CXF will try to inject contextual proxies (`@Context` annotations) into the provider instance. The injections are successful but the target object for them will be the proxy instance (not the real instance behind it). Consequently, at runtime, when the proxy delegates the call to a backing instance, all contextual proxies are null in there (simply put, not set).

### DIFF
--- a/integration/cdi/src/main/java/org/apache/cxf/cdi/CdiClassUnwrapper.java
+++ b/integration/cdi/src/main/java/org/apache/cxf/cdi/CdiClassUnwrapper.java
@@ -22,8 +22,18 @@ import java.util.regex.Pattern;
 
 import org.apache.cxf.common.util.ClassUnwrapper;
 
+/**
+ * Unwraps the CDI proxy classes into real classes.
+ */
 class CdiClassUnwrapper implements ClassUnwrapper {
-    private static final Pattern PROXY_PATTERN = Pattern.compile(".+\\$\\$.+Proxy");
+    /**
+     * Known proxy patterns for OWB and Weld:
+     * 
+     *  Xxx$$OwbNormalScopeProxy0
+     *  Xxx$Proxy$_$$_WeldClientProxy
+     *  
+     */
+    private static final Pattern PROXY_PATTERN = Pattern.compile(".+\\$\\$.+Proxy\\d*");
 
     CdiClassUnwrapper() {
 
--- a/integration/cdi/src/main/java/org/apache/cxf/cdi/JAXRSCdiResourceExtension.java
+++ b/integration/cdi/src/main/java/org/apache/cxf/cdi/JAXRSCdiResourceExtension.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.ServiceLoader;
 import java.util.Set;
 
+import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.context.spi.CreationalContext;
 import javax.enterprise.event.Observes;
 import javax.enterprise.inject.spi.AfterBeanDiscovery;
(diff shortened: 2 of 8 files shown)


reta:
@johnament @sberyozkin @rmannibucau could you guys take a look? The are few things I don't like (but don't have a better idea yet how to implement). The PR solves the simplest case with field injection (at the moment).

rmannibucau:
@reta did you try implementing it wrapping the bean annotated type (or attributes) to add a CxfInjectionInterceptor annotation binding and implement it with a @AroundConstruct? Should work for all "handlable" cases and looks more or less as complicated as this PR

integration/cdi/src/main/java/org/apache/cxf/cdi/CdiClassUnwrapper.java
```diff
@@ -22,8 +22,18 @@
...
 class CdiClassUnwrapper implements ClassUnwrapper {
-    private static final Pattern PROXY_PATTERN = Pattern.compile(".+\\$\\$.+Proxy");
+    /**
+     * Known proxy patterns for OWB and Weld:
+     * 
+     *  Xxx$$OwbNormalScopeProxy0
+     *  Xxx$Proxy$_$$_WeldClientProxy
+     *  
+     */
+    private static final Pattern PROXY_PATTERN = Pattern.compile(".+\\$\\$.+Proxy\\d*");
```
johnament: are you sure its always digits on the end?

reta: I see it consistently on OWB but it may not be always (hereby `\d*` to make this part optional)

reta:
@rmannibucau Yes, I tried this approach first thing (and would very likely complement this PR to handle the `@Context` arguments in constructors). The issue I hit hard is that I couldn't trace back the `ServerProviderFactory` to use for constructor injection :( Plus, the `InvocationContext::target` is set to `null` so I don't have access to the instance being constructed (very likely because is being constructed indeed).

rmannibucau:
You should have the instance after having called proceed normally if you return an Object, no? - or just use a lazy init in @AroundInvoke even if less elegant.

If you add an interceptor in the JAXRSCdiExtension you can pass to the interceptor the ServerProviderfactory and be contextual without any later lookup.

integration/cdi/src/main/java/org/apache/cxf/cdi/CdiClassUnwrapper.java
johnament: should it perhaps then be `.*`

reta:
@rmannibucau Thank a lot for the advice, I will try the implementation with `@AroundInvoke`, the `ServerProviderfactory` sadly is not a singleton but per-factory bean (so there could be more than one) ... Anyway, this is something very CXFy, figuring out the way to do that :)

integration/cdi/src/main/java/org/apache/cxf/cdi/CdiClassUnwrapper.java
reta: Hm ... not sure,`\d*` could be explained (and I have example for it), .* not sure ....

reta:
Better solution is required, closing

sberyozkin:
Not my fault this PR has been closed :-)

PR #340, opened 2017-11-15
[CXF-7562] correction for truncated flag

forsthofer:


rt/features/logging/src/main/java/org/apache/cxf/ext/logging/LoggingOutInterceptor.java
```diff
@@ -82,7 +82,10 @@ private OutputStream createCachingOut(Message message, final OutputStream os, Ca
             newOut.setThreshold(threshold);
         }
         if (limit > 0) {
-            newOut.setCacheLimit(limit);
+            // make the limit for the cache greater than the limit for the truncated payload in the log event, 
+            // this is necessary for finding out that the payload was truncated 
+            //(see boolean isTruncated = cos.size() > limit && limit != -1;)  in method copyPayload
+            newOut.setCacheLimit(limit + 1);
```
forsthofer: instead of just setting the limit to 'limit +1' one should check whether the limit == Integer.MAX_VALUE and if this is the case just use Integer.MAX_VALUE

sberyozkin:
Hi, thanks for this PR, it looks good, can you please sync it with the current master, also note a minor typo in the new function name, thanks.

deki:
Compile error due to missing import in LoggingInInterceptor, Checkstyle error in TruncatedTest

deki:
I've fixed the errors and merged it to master.

PR #343, opened 2017-11-20
Modify error message on failure to close synchronous conduit

joel-shemtov:
Failure to close synchronous conduit might occur even after a message has been successfully sent. For example it may happen when the response times out. In that case, the message "Could not send message" should be replaced by a more appropriate message.


joel-shemtov:
Hi Sergei, thanks for highlighting the typo. I have rectified that.

If an error occurs while sending the message then the wrapping interceptor class deals with it. I haven't changed that code. It creates a Fault exactly as the inner interceptor class did prior to this modification.

Sure I will apply this change to 3.1.x and 3.0.x without the java 8 structures (and without the typo).

sberyozkin:
Thanks, only update the current PR by moving Java8 code out of it, it will be simpler then for us to merge to 3.2.x and then to 3.1.x

joel-shemtov:
Any idea why the test failure?

sberyozkin:
"[ERROR] /home/jenkins/jenkins-slave/workspace/CXF-Trunk-PR/core/src/main/java/org/apache/cxf/interceptor/MessageSenderInterceptor.java:64: Line is longer than 120 characters (found 164). [LineLength]"

joel-shemtov:
Thanks. Is the console output the right place to find test errors?

sberyozkin:
Console def helps if it is related to the checkstyle/PMD failures

joel-shemtov:
> org.junit.ComparisonFailure: expected:<Could not [send Messa]ge.> but was:<Could not [complete Exchan]ge.>

I will look at modifying this test and possibly other tests too.

joel-shemtov:
Hi @sberyozkin 
> [INFO] Running org.apache.cxf.systest.ws.rm.MessageCallbackOnewayTest
Build timed out (after 90 minutes). Marking the build as aborted.

While this test found no faults, it was aborted by the clock.

This test started at 15:11:19 GMT. Is there a better time to test?

Is there a way to trigger a test without committing? (and by the way it is possible to commit without triggering a test?)

sberyozkin:
I'm not sure why Jenkins fails, but it's important to have a local build passing.
By the way, I'm looking at the changes and I'm not sure I understand them. For ex, the test there where a schema first validation fails, does it fail before the message has been sent ? Similarly, in the failover cases, that def fails because the message has not been sent properly. Thus, in those cases, seeing "Can not complete Exchange" would be now confusing ?

joel-shemtov:
I don't know for sure whether or not the message has been sent. I have seen, as a user, cases where the message has been sent then. Indeed it might not be the case in these build tests.

Considering your comment, I will revert the changes I've made to the testing code.

Instead I'll set a much finer criterion for placing this new error message. If it fails the test, I will know for sure that the  and try to pass it again.

It will be something like:
`

    public static class MessageSenderEndingInterceptor extends AbstractPhaseInterceptor<Message> {
        public MessageSenderEndingInterceptor() {
            super(Phase.PREPARE_SEND_ENDING);
        }

        public void handleMessage(Message message) throws Fault {
            try {
                getConduit(message).close(message);
            } catch (SocketTimeoutException e) {
                throw new Fault(new org.apache.cxf.common.i18n.Message("COULD_NOT_COMLETE", BUNDLE), e);
            } catch (IOException e) {
                throw new Fault(new org.apache.cxf.common.i18n.Message("COULD_NOT_SEND", BUNDLE), e);
            }
        }
    }

`

where the first catch clause is the only addition to the current code.

Would it not be better to close this pull request and open a fresh new one?

sberyozkin:
Indeed, it may actually be the best solution, react to the socket exception, as in this case we know the message has gone as far as the client conduit is concerned.

joel-shemtov:
Closing this pull request. Opening a new one with the solution discussed above.

PR #345, opened 2017-11-24
When failing to close conduit, distinct read faults from write faults

joel-shemtov:
Consider SocketTimeoutExceptions to be incoming message faults while all other
IOExceptions to be outgoing message faults. Each type will have its own error message.

### DIFF
--- a/core/src/main/java/org/apache/cxf/interceptor/MessageSenderInterceptor.java
+++ b/core/src/main/java/org/apache/cxf/interceptor/MessageSenderInterceptor.java
@@ -20,6 +20,7 @@
 package org.apache.cxf.interceptor;
 
 import java.io.IOException;
+import java.net.SocketTimeoutException;
 import java.util.ResourceBundle;
 
 import org.apache.cxf.common.i18n.BundleUtils;
@@ -60,6 +61,8 @@ public class MessageSenderInterceptor extends AbstractPhaseInterceptor<Message> 
         public void handleMessage(Message message) throws Fault {
             try {
                 getConduit(message).close(message);
+            } catch (SocketTimeoutException e) {
+                throw new Fault(new org.apache.cxf.common.i18n.Message("COULD_NOT_RECEIVE", BUNDLE), e);
             } catch (IOException e) {
                 throw new Fault(new org.apache.cxf.common.i18n.Message("COULD_NOT_SEND", BUNDLE), e);
             }
--- a/core/src/main/java/org/apache/cxf/interceptor/Messages.properties
+++ b/core/src/main/java/org/apache/cxf/interceptor/Messages.properties
@@ -29,6 +29,7 @@ NO_DATAREADER=No DataReader is available for Service: {0}
 NO_DATAWRITER=No DataWriter is available for Service: {0}
 COULD_NOT_UNRWAP=Could not unwrap message parts.
 COULD_NOT_SEND=Could not send Message.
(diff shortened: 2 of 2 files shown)


sberyozkin:
I merged it, thanks, though I got confused first that I may've accidentally pushed in some unrelated changes, it looks good, thanks

joel-shemtov:
Thanks for merging. Proud to have my tiny share in CXF. To be honest I was surprised to see that the build worked without any failure. Is it possible that there's no timeout scenario test? Such that would trigger SocketTimeoutException?

sberyozkin:
Thanks, please keep the pull requests coming. Are you working with JAX-WS ? Perhaps you can experiment with copying one of the system tests in systests/jaxws and updating the endpoint to wait for longer than 60 secs, or may be the configured HttpClient receieveTime, say 5 secs ? Thanks

joel-shemtov:
Hi Sergei, I checked and found that there actually is a method called testTimeout() in _org.apache.cxf.systest.dispatch.DispatchClientServerTest_. However I'm not sure if it really tests timeout.  

This is the current code on master


`

    @Test
    public void testTimeout() throws Exception {
        //CXF-2384
        URL wsdl = getClass().getResource("/wsdl/hello_world.wsdl");
        assertNotNull(wsdl);

        //pick one of the other service/ports that would have an address
        //without a service running
        QName otherServiceName = new QName("http://apache.org/hello_world_soap_http",
                "SOAPProviderService");
        QName otherPortName = new QName("http://apache.org/hello_world_soap_http", "SoapProviderPort");


        SOAPService service = new SOAPService(wsdl, otherServiceName);
        assertNotNull(service);

        Dispatch<SOAPMessage> disp = service
            .createDispatch(otherPortName, SOAPMessage.class, Service.Mode.MESSAGE);
        disp.getRequestContext().put(BindingProvider.ENDPOINT_ADDRESS_PROPERTY,
                                     "http://localhost:"
                                     + TestUtil.getPortNumber("fake-port")
                                     + "/SOAPDispatchService/SoapDispatchPort");

        DispatchImpl<?> dispImpl = (DispatchImpl<?>)disp;
        HTTPConduit cond = (HTTPConduit)dispImpl.getClient().getConduit();
        cond.getClient().setConnectionTimeout(500);

        InputStream is = getClass().getResourceAsStream("resources/GreetMeDocLiteralReq.xml");
        SOAPMessage soapReqMsg = MessageFactory.newInstance().createMessage(null, is);
        assertNotNull(soapReqMsg);

        try {
            disp.invoke(soapReqMsg);
            fail("Should have faulted");
        } catch (SOAPFaultException ex) {
            fail("should not be a SOAPFaultException");
        } catch (WebServiceException ex) {
            //expected
            assertTrue(ex.getCause().getClass().getName(),
                       ex.getCause() instanceof java.net.ConnectException
                       || ex.getCause() instanceof java.net.SocketTimeoutException);
        }
        dispImpl.close();

    }
`

To me it seems that the usage of "fake-port", guarantees that only ConnectionException gets caught. SocketTimeoutException is never actually thrown.

As far as I understand, these are two different scenarios. Shouldn't they be in two separate tests? As you suggested I'll try to run it locally. When I have it modified and tried, I'll open a new pull request.

sberyozkin:
Hi, that looks like it's testing a timeout that might occur when a client is posting a large payload and no transport-level confirmation has arrived in time...Consider copying any end-to-end JAX-WS system test and modify the service impl...

PR #346, opened 2017-11-25
CXF-7544: Support @Context-based injection into proxied CDI beans

reta:
The issue pop up as part of #330 discussion. In case when provider / feature / resource is a proxied CDI bean, the contextual class members (annotated with @context) are not injected.

More broadly, in some circumstances (like using @ApplicationScoped annotation for example) the CDI runtime will create a proxy class for a particular bean. As the result, the CXF side is going to bind the particular provider metadata to this proxy instance. It looks logical and unambiguous.

However, the interesting things are happening when CXF will try to inject contextual proxies (@Context annotations) into the provider instance. The injections are successful but the target object for them will be the proxy instance (not the real instance behind it). Consequently, at runtime, when the proxy delegates the call to a backing instance, all contextual proxies are null in there (simply put, not set).

@johnament @rmannibucau Addressing the issue requires a significant revamp of the CXF injection mechanism, it is turned out to be hard to solve even with `@AroundConstruct` or, previously, providing custom `@Context` injectors. The good news is that the field-based injection could be easily workarounded using setter-based injection. After conversations with @sberyozkin, we decided that encapsulating CXF injection implementation and than delegating the hard work to appropriate strategy (CDI, Spring, ...) would be better time investment than trying to patch the existing one. Since we have simple workaround, it sounds like an acceptable tradeoff (https://issues.apache.org/jira/browse/CXF-7571).

This PR fixes the patterns in `CdiClassUnwrapper`, adds tests for it and fixes the `FilterProviderInfo` resource / service class resolution (to match it to the `ProviderInfo`)

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/provider/ServerProviderFactory.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/provider/ServerProviderFactory.java
@@ -211,7 +211,11 @@ public final class ServerProviderFactory extends ProviderFactory {
                 for (Object featureProvider : cfg.getInstances()) {
                     Map<Class<?>, Integer> contracts = cfg.getContracts(featureProvider.getClass());
                     if (contracts != null && !contracts.isEmpty()) {
-                        allProviders.add(new FilterProviderInfo<Object>(featureProvider,
+                        Class<?> providerCls = ClassHelper.getRealClass(getBus(), featureProvider);
+                        
+                        allProviders.add(new FilterProviderInfo<Object>(featureProvider.getClass(),
+                                                                        providerCls,
+                                                                        featureProvider,
                                                                         getBus(),
                                                                         contracts));
                     } else {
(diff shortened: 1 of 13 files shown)


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/provider/ServerProviderFactory.java
```diff
@@ -211,7 +211,11 @@ protected void setProviders(boolean custom, boolean busGlobal, Object... provide
                 for (Object featureProvider : cfg.getInstances()) {
                     Map<Class<?>, Integer> contracts = cfg.getContracts(featureProvider.getClass());
                     if (contracts != null && !contracts.isEmpty()) {
-                        allProviders.add(new FilterProviderInfo<Object>(featureProvider,
+                        Class<?> providerCls = ClassHelper.getRealClass(getBus(), featureProvider);
```
reta: Turned out we handle `ProviderInfo` differently than `FilterProviderInfo`, this minors fixes will match the resource and service class resolution.

rmannibucau:
+1 until CXF-7571 is put in place with good defaults.

sberyozkin:
Hi Andriy

Why this has been merged ?

sberyozkin:
Sorry, just got confused given that the JIRA issue itself was closed as Won't fix

sberyozkin:
Andriy, will it make sense to merge to 3.1.x as well ?

reta:
Yeah, sure, I will cherry-pick it to 3.1.x. Did you want to add / comment before merge? Or confusion is resolved? :-)

sberyozkin:
Hi Andriy, indeed, the fix looks good, it's really about the filter providers also providing the real and service classes :-), thanks

reta:
Thanks! Yeah, that's right, also fixing CdiClassUwrapper proxy pattern for OWB, but aside from that, it is mostly about test cases.

PR #351, opened 2017-12-02
[CXF-7571] Adding support for CDI injection of @Context objects.

johnament:
This is a WIP right now (needs more testing).  But it does the gist of what I was aiming to do, which is to make it so that CDI takes over injection.

I figure there's areas of the code that have to be removed now.  If you guys have some pointers about where they might be, would be great.


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResourceContextImpl.java
```diff
@@ -30,13 +32,15 @@
...
     private static final String CONTEXT_PROVIDER_PROP = "org.apache.cxf.jaxrs.resource.context.provider";
-    private ClassResourceInfo cri;
-    private Class<?> subClass;
-    private Message m;
-    public ResourceContextImpl(Message m, OperationResourceInfo ori) {
+    private final ClassUnwrapper classUnwrapper;
+    private final ClassResourceInfo cri;
+    private final Class<?> subClass;
+    private final Message m;
+    public ResourceContextImpl(Message m, OperationResourceInfo ori, Bus bus) {
```
johnament: I need to revert this change.

systests/cdi/base/src/main/java/org/apache/cxf/systests/cdi/base/BookStore.java
```diff
@@ -38,13 +39,11 @@
...
     @Inject private BookStoreService service;
-    @Inject private String version;
+    @Inject private BookStoreVersion bookStoreVersion;
 
-    @GET
     @Path("/version")
-    @Produces(MediaType.TEXT_PLAIN)
-    public String getVersion() {
-        return version;
+    public BookStoreVersion getVersion(@Context ResourceContext resourceContext) {
```
johnament: I need to remove this.

integration/cdi/src/main/java/org/apache/cxf/cdi/DelegateContextAnnotatedType.java
```diff
@@ -0,0 +1,182 @@
...
+        this.replacedFields = replaceFields(original);
+    }
+
+    private Set<AnnotatedField<? super X>> replaceFields(AnnotatedType<? super X> delegate) {
+        return delegate.getFields().stream().map(this::wrap).collect(toSet());
+    }
+
+    Set<Type> getContextFieldTypes() {
+        return replacedFields.stream()
+                .filter(f -> f.isAnnotationPresent(Context.class) || f.isAnnotationPresent(ContextResolved.class))
```
reta: Wondering if we need to check for `ContextResolved` here? Technically, it should not be present, we add it along with inject, right?

johnament: It would be for anything that isn't in the `STANDARD_CONTEXT` map, but is explicitly annotated and provided as a `Context` object.  Could be anything with a custom `ContextProvider` implemented that isn't injected using `@Context` but has the qualifier.

johnament: Just to make sure the issue is understood, not sure if I'm not clear.  I'm registering a CDI bean per `@Context` object that can be injected.  I'm discovering them two ways:

- Any `@Context` injection point
- Any of the standard context objects, as defined by `InjectionUtils.STANDARD_CONTEXT_CLASSES`.

The latter is what enables `@Inject UriInfo` to work properly.  Without it, the injection point would not be satisfied.  The problem is that this works fine for the standard context classes, but not custom objects (e.g. `TracerContext`).  The closest I can do is if they have an injection point `@Inject @ContextResolved TracerContext` then I can still resolve that injection point, but I can't resolve `@Inject TracerContext` since I have no idea its supposed to be from the `@Context`.  

If we changed the behavior of these custom context objects so that they were added to a map in `InjectionUtils`, then I could resolve those classes as well and remove that extra check.

reta: Aha, thanks for explanation @johnament, the role of `@ContextResolved` becomes clear now. That's actually an interesting challenge to solve ... We definitely know all of them at the moment `JAXRSServerFactoryBean` is being constructed but it is too late with respect to CDI lifecycle (and what can we do at this point).

johnament: So what do you think about adding support for registering the context type, manually?

reta: I think we could only manually register the context providers which come from CXF, but not the user ones. May be could deduct the context providers by introspecting the providers / resources / features during the `collect`? We have interface to check for, `ContextProvider`, this should for all cases. What do you think?

johnament: We could, however it seems that the registration of the `ContextProvider` is done within CXF, which would be too late for CDI.  In addition, I don't want to require/expect that all `ContextProvider` impls are CDI beans.

reta: Sorry for keeping silence, haven't had time to get back to the issue. Spent some time playing with the context providers, my personal option (purely from user perspective), using `@Inject @ContextResolved SomeContext context` is not very intuitive.  

Getting back to the initial problem: `@Inject TracerContext` and `@Context TracerContext`. Besides the solution you have with `@ContextResolved`, we could also introduce 1) a marker interface for context objects 2) or an annotation (similar to `@ContextResolved`) for context objects 3) require context providers to be placed into CDI archives (than at least we could discover the possible context objects while processing annotated types) 

There are certainly more options available, but I think the 3rd one is as best as we could do without introducing too much of confusion. That would certainly impose some constraints (as you mentioned before) but may be it is not as bad as it sounds? What do you think?

johnament: @reta  just want to clarify my point a bit more.

The simplest solution I can think of is to add this snippet to `BraveContextProvider`

```java
static {
        InjectionUtils.STANDARD_CONTEXT_CLASSES.add(BraveTracerContext.class.getName());
    }
```

Or do something like require a service loader to give the class name.  Then you wouldn't need to do things like make the providers CDI component classes (which doesn't sound right).

FWIW, I'm almost inclined to add `<beans bean-discovery-mode="none"/>` to modules that shouldn't have CDI beans within them, to avoid excessive classloadering.  But another day.

reta: Right, forgot about that, this is what you was referring to as manual registration. Ok, here is my suggestion, let us start from this simple solution (either using `InjectionUtils.STANDARD_CONTEXT_CLASSES`, or in safer with another collection, f.e. `InjectionUtils.NON_STANDARD_CONTEXT_CLASSES`), and revisit it later when the PR would be more or less feature complete. I suspect we are going to have quite a few challenges ahead, so we may use absolutely different approach at the end :-) At least we could get rid of `@ContextResolved` for now and move on. Sounds good?

johnament: I would still recommend keeping `@ContextResolved` just to allow multiple beans to be registered (avoiding close tie in).  I'll work on getting the additional field processed and update this PR.

johnament: Alright, so I've done some digging around and have bad news :-(

For the custom created context objects, pretty much everything fails CDI use cases.  In order to get the scoping correct, the context objects must be request scoped.  However, all of the custom context objects CXF has fail CDI requirements, they don't have default constructors.  As a result they can't be request scoped beans.

So each of those would need to remain as `@Context` objects for injection purposes.

sberyozkin:
"My thinking is to replace the current API objects that are concrete with interfaces" - what do you mean ?

integration/cdi/src/main/java/org/apache/cxf/cdi/DelegateContextAnnotatedType.java
reta: I expected the roadblocks here, `@Context` it be ...

reta:
@johnament  Hm ... most context objects do have interfaces, if we look on STANDARD_CONTEXT_CLASSES. Custom contexts also do have them, f.e. `TracerContext` is the interface. Or you mean different API objects?

johnament:
@reta so should a user be doing `@Context OpenTracingContext` or `@Context TracerContext` to get the injection?  CDI can support the latter, but the former isn't a valid bean.

I also think the OIDC context objects fail in this area as well.

reta:
`@Context TracerContext` only, the provider is specific to tracing framework, but interface is same for everyone

sberyozkin:
IdTokenContext is an interface: https://github.com/apache/cxf/blob/master/rt/rs/security/sso/oidc/src/main/java/org/apache/cxf/rs/security/oidc/rp/IdTokenContext.java

johnament:
Ok, that makes me feel better.  there is one that doesn't have an interface, so it will be an issue: `OAuthContext`.  But we could just skip it for now.

Here's something that is a possible issue.  `UserInfoContext` extends `IdTokenContext`, so you could end up with ambiguous dependencies (but may not if i only register one type).

So I probably wasn't too far off, just off by 1.

sberyozkin:
Yeah, that OAuthContext has never been injectable :-), I did it awhile back and the idea was that may be the RS code might want to get the current access token properties like messageContext,getContent(OAuthContext.class), but indeed, it would need to be converted to the interface for it to be accessed as Context...

integration/tracing/tracing-opentracing/src/main/java/org/apache/cxf/tracing/opentracing/jaxrs/OpenTracingTracerContextClassProvider.java
```diff
@@ -0,0 +1,29 @@
...
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.cxf.tracing.opentracing.jaxrs;
+
+import org.apache.cxf.jaxrs.ext.ContextClassProvider;
+import org.apache.cxf.tracing.TracerContext;
+
+public class OpenTracingTracerContextClassProvider implements ContextClassProvider {
```
johnament: This is one approach.  Now that I know what to do to fix the registration, I'm going to try with the explicit class addition via static method, see if that works.  But this would be the alternative via service loader.

systests/cdi/base/src/main/java/org/apache/cxf/systests/cdi/base/OpenTracingProducer.java
```diff
@@ -0,0 +1,32 @@
...
+ * under the License.
+ */
+package org.apache.cxf.systests.cdi.base;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.inject.Produces;
+import org.apache.cxf.tracing.opentracing.jaxrs.OpenTracingFeature;
+
+@ApplicationScoped
+public class OpenTracingProducer {
```
johnament: I couldn't get automatic registration to work.  Not sure why, but I went with the hint from Sergey and added an explicit producer to make sure it was a CDI bean.

integration/tracing/tracing-opentracing/src/main/java/org/apache/cxf/tracing/opentracing/jaxrs/OpenTracingTracerContextClassProvider.java
johnament: I just tested it, using a static method won't work.  I'm assuming that the class is loaded too late to register via the CDI extension.  Are there any concerns with a service loader based approach?

systests/cdi/base/src/main/java/org/apache/cxf/systests/cdi/base/OpenTracingProducer.java
reta: I will take a look, should be working (in general but we haven't tested with CDI)

reta: Adding `bean.xml` to `opentracing` module fixed the issue right away, the `OpenTracingFeature` has been discovered. As a side note, could you please move these testcases to `systest-tracing` project, it would be good to keep the CDI tests clean from external integration. Thank you.

systests/cdi/base/src/main/java/org/apache/cxf/systests/cdi/base/BookStoreVersion.java
```diff
@@ -22,20 +22,25 @@
...
+import org.apache.cxf.systests.cdi.base.context.CustomContext;
+
 @RequestScoped
 public class BookStoreVersion {
     @Inject
     private String version;
-    @Context
+    @Inject
     private HttpHeaders httpHeaders;
+    @Inject
```
johnament: @sberyozkin @reta  I'm a bit stumped at this point.  Even when I annotate the field with `@Context`, I don't see it being processed using my `ContextProvider` yet somehow there is a value set..

reta: @johnament Not sure I understood the exact nature of the problem, but with `@Context` annotation the `CustomContext` is being processed by `DelegateContextAnnotatedType` which injects into it the proxy value (CDI managed one), and later on the instance is created by `ContextProducerBean` on first invocation.  Is it something along this lines or absolutely different flow?

johnament: @reta yes, that's what I'm expecting to happen, however I'm never seeing the bean invoked for this context object.

reta: @johnament I conducted a short debug session, it looks like it is working as expected / described, I see `DelegateContextAnnotatedType`, `ContextProducerBean` and finally `CustomContextImpl` being invoked in case when `CustomContext` is annotated with `@Context` annotation. Interesting, what is happening in your case ...

johnament:
Ok, so I had a logic issue (and sorry for the delay in getting this done, I've been a bit more offline the past week and a half) in the new context object that I created (`message.getId()` is always null for rest messages).

So this is all working well, I'm going to plan to add the service loader to the identified modules but I don't see a lot of value in testing them.  What do you think @reta @sberyozkin ?

integration/cdi/src/main/java/org/apache/cxf/cdi/DelegateContextAnnotatedType.java
johnament: This is now working fine.  With the service loader approach, we can `@Inject` any context object we want now.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/InjectionUtils.java
```diff
@@ -1526,4 +1529,13 @@ public static Type processGenericTypeIfNeeded(Class<?> serviceCls, Class<?> para
     public static Object getEntity(Object o) {
         return o instanceof GenericEntity ? ((GenericEntity<?>)o).getEntity() : o;
     }
+
+    public static Set<Class<?>> getCustomContextClasses() {
```
johnament: Is this utility OK to include here?  Or should I move it into the CDI module?

johnament:
BTW, I'm assuming that there was some code added already that handled `@Context` for CDI objects.  I should probably pull that out in this change, but I'm not sure where that's happening.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/InjectionUtils.java
reta: Moving to CDI module would be better I believe, so it would be clear where this flow belongs to

johnament: Sure, but just to confirm the actual interface `ContextClassProvider` stays in frontend/jax-rs since multiple modules will use it (and shouldn't rely on CDI)

reta:
@johnament  I think it is acceptable to not test each module (the service loader). You have a test case for that in the CDI test suite, it should work just fine I think.

reta:
One specific use case which I was not able to find the clean way to implement is using constructor-based injection for `@Context` instances (in this case the `Message` instance is not available). Are you planning to include the support for that?

johnament:
@reta originally I wasn't going to do that, since its just a CDI feature.  But I updated the existing resources to demonstrate both constructor and field injection.

johnament:
I'll address any remaining issues in a separate PR.  I'm still curious about removing the current logic for providers, but can address that later.

PR #352, opened 2017-12-06
Config handling for MP rest client

andymc12:
@johnament Can you take a look at these changes?

This should help us handle the `Configurable` methods.  Notice that the overrides of the priority do not currently work (note the test that has the `@Ignore` annotation).  I ran a similar test using the JAX-RS client and that worked, so I'm probably just missing something simple...  if you see it, please let me know.  Thanks!


johnament:
@andymc12 I think this is fine.  weird that the priority isn't coming over.  I see in `ConfigurableImpl` we read the priority.

johnament:
I see why it doesn't work.  It's related to https://issues.apache.org/jira/browse/CXF-7543 , basically the JAX-RS client impl is done as a completely different path than the WebClient codepath.  Proxies are based on the WebClient codebase.

johnament:
This one is `ProviderFactory.comparePriorityStatus` only considers the annotation, not the registered priority.

PR #353, opened 2017-12-07
[CXF-7579] More config impl.

johnament:
@andymc12  I'm not sure if this makes it worse.  I'm now seeing the call hit the server (I have something on port 80).

### DIFF
--- a/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/DefaultResponseExceptionMapper.java
+++ b/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/DefaultResponseExceptionMapper.java
@@ -1,0 +1,34 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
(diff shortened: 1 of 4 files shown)


johnament:
I have to dig in further, but it seems that the priority registered is ignored for client request filters, they don't go through the sorting algorithm.

johnament:
@andymc12  I'm going to bring my branch into the repo shortly.  What were your thoughts on how to handle the interface validation?  There is nothing in CXF today.

andymc12:
@johnament 
> What were your thoughts on how to handle the interface validation?

My thought was that we would add some validation code in the `CxfTypeSafeClientBuilder.build(...)` method.  That way, we avoid doing expensive things like looking up providers, etc. if the interface is bad.

johnament:
@andymc12  FYI, if you want to take a look

```
java.lang.IllegalStateException: The entity is not backed by an input stream, entity class is : java.lang.String
	at org.eclipse.microprofile.rest.client.tck.CustomHttpMethodTest.invokesUserDefinedHttpMethod(CustomHttpMethodTest.java:52)
```

PR #356, opened 2017-12-12
[CXF-7587] use Jetty context path when setting HTTP handler name

johnpoth:
https://issues.apache.org/jira/browse/CXF-7587

Thanks!


ffang:
I just verified the JAXRSClientServerUserResourceAsteriskTest can run successfully with http-undertow transport

johnpoth:
Thanks @ffang !

PR #359, opened 2017-12-19
Adding an implementation of the MicroProfile Rest Client v1.0

johnament:
This is an implementation of the MP Client 1.0.  I plan to cut the spec release tomorrow, so don't quite approve this yet but need input on any code clean up required.

There's still one TCK test to add in, but not critical for a first pass.  I'd also like to add more TCK running options (Tomcat, OWB) to prove out the different runtime scenarios but also not critical for getting the code in (don't really want to stay diverged for too much longer).

I also think it'll be useful to get this started to integrate into consumers so having it out there will help.

I did end up going with a dedicated module for the code, since I had to do some amount of subclassing and code alignment to support some of the features missing from the native client.  In the end, functionally the native client works fine, but there's some tweaking needed for provider ordering and instantiation that I had to work with.  May be worth merging it into that module at some point, but it could be a breaking change for users.


rt/rs/microprofile-client/pom.xml
```diff
@@ -0,0 +1,230 @@
...
+    <url>http://cxf.apache.org</url>
+    <parent>
+        <groupId>org.apache.cxf</groupId>
+        <artifactId>cxf-parent</artifactId>
+        <version>3.2.2-SNAPSHOT</version>
+        <relativePath>../../../parent/pom.xml</relativePath>
+    </parent>
+    <properties>
+        <cxf.osgi.import>
+            org.apache.aries*;version="${cxf.aries.version.range}";resolution:=optional,
```
johnament: I am not an OSGi expert, but I suspect this section is wrong.

rt/rs/microprofile-client/pom.xml
```diff
@@ -0,0 +1,230 @@
...
+            org.springframework*;resolution:="optional";version="${cxf.osgi.spring.version}"
+        </cxf.osgi.import>
+        <!-- keep in sync with services exported in activator -->
+        <cxf.export.service>
+            org.apache.aries.blueprint.NamespaceHandler;osgi.service.blueprint.namespace="http://cxf.apache.org/blueprint/jaxrs-client"
+        </cxf.export.service>
+        <cxf.bundle.activator>org.apache.cxf.jaxrs.client.blueprint.Activator</cxf.bundle.activator>
+    </properties>
+    <dependencies>
+        <dependency>
```
johnament: I don't know how many of these dependencies are actually needed, I want to clean it up a bit.

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/JsonPProvider.java
```diff
@@ -0,0 +1,26 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.cxf.microprofile.client;
+
+import javax.annotation.Priority;
+import org.apache.cxf.jaxrs.provider.jsrjsonp.JsrJsonpProvider;
+
+@Priority(Integer.MAX_VALUE)
+public class JsonPProvider extends JsrJsonpProvider {
```
johnament: I believe I can now get rid of this, and instead manually add a ProviderInfo with custom = false.

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientFactoryBean.java
```diff
@@ -0,0 +1,118 @@
...
+            return new MicroProfileClientProxyImpl(actualState, proxyLoader, cri, isRoot,
+                    inheritHeaders, varValues);
+        }
+    }
+
+    Configuration getConfiguration() {
+        return configuration;
+    }
+
+    private Set<Object> processProviders() {
```
johnament: I suspect this will be useful to move into the main client, since it handles converting filters into `FilterProviderInfo` objects, rather than regular `ProviderInfo`s

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/RestClientBean.java
```diff
@@ -0,0 +1,162 @@
...
+    public boolean isNullable() {
+        return false;
+    }
+
+    @Override
+    public Object create(CreationalContext<Object> creationalContext) {
+        CxfTypeSafeClientBuilder builder = new CxfTypeSafeClientBuilder();
+        String baseUrl = getBaseUrl();
+        try {
+            return builder.baseUrl(new URL(baseUrl)).build(clientInterface);
```
johnament: @andymc12 I just realized when reading this, we are missing TCK tests for the configurable CDI properties outside of URL.

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/proxy/MicroProfileClientProxyImpl.java
```diff
@@ -0,0 +1,65 @@
...
+                if (t instanceof RuntimeException) {
+                    throw t;
+                } else if (t != null && m.getExceptionTypes() != null) {
+                    // its a checked exception, make sure its declared
+                    for (Class c : m.getExceptionTypes()) {
+                        if (t.getClass().isAssignableFrom(c)) {
+                            throw t;
+                        }
+                    }
+                    // TODO Log the unhandled declarable
```
johnament: Still have this pending TODO.

sberyozkin:
Hi John, it looks good, as far as the initial implementation is concerned. Indeed the ordering of the providers is a very sensitive issue, so lets discuss it further at the dev, if needed something can be easily customized at the lower level to suit the needs of the microprofile client

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientConfigurableImpl.java
```diff
@@ -0,0 +1,69 @@
...
+
+    boolean isDefaultExceptionMapperDisabled() {
+        Object prop = getConfiguration().getProperty(CONFIG_KEY_DISABLE_MAPPER);
+        if (prop instanceof Boolean) {
+            return (Boolean)prop;
+        } else {
+            Config config = ConfigProvider.getConfig();
+            Optional<Boolean> optionalValue = config.getOptionalValue(CONFIG_KEY_DISABLE_MAPPER,
+                    Boolean.class);
+            if (optionalValue.isPresent()) {
```
reta: Minor, just to make it compact `return config.getOptionalValue(CONFIG_KEY_DISABLE_MAPPER, Boolean.class).getOrElse(false);"

johnament:
My plan is to merge this branch into master later today, since I'm way too far off master at this point.  I'll address any remaining issues afterwards.  Those include:

- TCK against OWB
- Remaining configuration changes

PR #360, opened 2017-12-20
MP Rest Client interface validation and tests

andymc12:
Checks:
- that the passed-in type is an interface
- that each method has, at most, one HTTP method annotation
- that URI template variables in @Path annotations are resolved by
@PathParam annotations


rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/CxfTypeSafeClientBuilder.java
```diff
@@ -42,6 +42,7 @@ public RestClientBuilder baseUrl(URL url) {
         if (baseUri == null) {
             throw new IllegalStateException("baseUrl not set");
         }
+        Validator.checkValid(aClass);
```
johnament: Can you also remove the line in the TCK module?

PR #362, opened 2017-12-28
[CXF-7596] Adding javax.net OSGI package import into HTTP Transports …

precoder:
…bundle to fix Class Not Found Exceptions.

### DIFF
--- a/rt/transports/http/pom.xml
+++ b/rt/transports/http/pom.xml
@@ -33,6 +33,7 @@
     <properties>
         <cxf.bundle.activator>org.apache.cxf.transport.http.osgi.HTTPTransportActivator</cxf.bundle.activator>
         <cxf.osgi.import>
+            javax.net,
             javax.servlet*;version="${cxf.osgi.javax.servlet.version}",
             org.apache.aries*;version="${cxf.aries.version.range}";resolution:=optional,
             org.springframework*;resolution:="optional";version="${cxf.osgi.spring.version}",


precoder:
Hey, 
I have made this pull request for the issue I made in the JIRA by following the instructions on the CXF home page.

Here is the brief description of the problem:
**javax.net** is a package in the Java rt.jar and it has **javax.net.SocketFactory** class, which is the base class for the **javax.net.ssl.SSLSocketFactory**

org.apache.cxf.cxf-rt-transports-http bundle does import "javax.net.ssl" but it does not import "javax.net".
As a result of this behaviour, we get a Class Not Found Exception in case of **SSLSocketFactory** is loaded by  transports-http bundle.

Currently I can have a workaround by setting the boot delegation property for the Felix Framework
(or any other OSGI framework) as following:

org.osgi.framework.bootdelegation=javax.net

However, in my opinion this causes an inconsistent behaviour since javax.net.ssl is imported by the bundle but the base package must be boot delegated.
In my opinion either we can say everything in the rt.jar should be boot delegated or rt.jar packages must be imported same as any other third party libraries. I personally would say boot deletion should not be the default way.

If you do not accept this as a Bug do not accept solution proposal here,
could you please show me the correct configuration proposal.

Note: In Karaf I do not get this error since I guess package is somehow boot delegated. We found it by debugging the org.apache.felix.framework.BundleWiringImpl but I could not find the exact configuration for the boot delegation in Karaf.

PR #365, opened 2018-01-11
[CXF-7520]:Ensure CXF can build with JDK9 GA(build 9+181)

jimma:


ffang:
@jimma , looks good to me.
Just one thing, as this is JDK9 specific, could you please add a java9 profile in pom.xml and put jaxb-impl dependency there?
Also how about add a code to check the jdk version , if it's jdk9, set the NoEscapeHandler?

Thanks!

asoldano:
@ffang, do you think that's really needed? As long as there's no side effect when running on older jdks, I don't think we should be increasing the complexity of the build.

dkulp:
The deps are scope=provided so they shouldn't affect anyone downstream as they won't be pulled in.  Thus, the profile is likely not needed.

HOWEVER, there is an issue at runtime if the com.sun.xml.bind.marshaller.CharacterEscapeHandler class is not found (example: using in JDK version of jaxb instead of the jar), then the writer won't be createable at all and everything will fail.

ffang:
java9 profile make the pom more organized, we know it's for java9 only, like a marker.  I think it's important for us to clearly know which is java9 specific, can help us to resolve similar issue in the future, as we have some clear marker to draw upon. 

My 2 cents.

dkulp:
One more question:  does this actually WORK?    What happens if you try to send a < or a > character?  Those need to be escaped.

jimma:
> Just  one thing, as this is JDK9 specific, could you please add a java9 profile in pom.xml and put jaxb-impl dependency there?
Also how about add a code to check the jdk version , if it's jdk9, set the NoEscapeHandler?

@ffang  This fix is not only for jdk9. If we want to upgrade to jaxb2.3, we still need to set an escape handler to make jdk8 makes.

jimma:
> Similar reasoning for the NoEscapeHandler: @jimma , does the addition of that affect the execution with jdk 8?
@asoldano  It doesn't. If a old jaxb Marshaller doesn't support this property setting, the PropertyException will be ignored.

ffang:
Hi @jimma 

As Dan pointed out, the big issue is that your code impose compulsory restrictions that the CXF runtime must have jaxb-impl jar dependency. Otherwise  you run into ClassNotFoundException, as the class name is com.sun.xml.internal.bind.marshaller.CharacterEscapeHandler in JDK version but not com.sun.xml.bind.marshaller.CharacterEscapeHandler, the package name is different. In this case the DataWriter can't be created at all. I don't think we should introduce this compulsory dependency.

So I suggest you have some static code to detect it first, something like

` static {
                   
            try {
               DataWriterImpl.class.getClassLoader().
                    loadClass("com.sun.xml.bind.marshaller.CharacterEscapeHandler");
                    //if no CNFE means we have saaj-impl dependeceny so in this case 
                   //register the customized handler
            } catch (ClassNotFoundException cnfe) {
                LOG.debug("can't load class com.sun.xml.bind.marshaller.CharacterEscapeHandler", cnfe);
            }
        }
    }
 `

jimma:
> The deps are scope=provided so they shouldn't affect anyone downstream as they won't be pulled in. Thus, the profile is likely not needed.
> 
> HOWEVER, there is an issue at runtime if the com.sun.xml.bind.marshaller.CharacterEscapeHandler class is not found (example: using in JDK version of jaxb instead of the jar), then the writer won't be createable at all and everything will fail.

@dkulp  I saw the provided scope is set in cxf-core, so I set the same thing. I think this can allow downstream to use the jaxb version from the place like tomcat container lib or user provided a higher jaxb version than 2.2.11 to quickly pick up some jaxb issue fix.  But it looks currently cxf can't compile or run with JDK's jaxb version because cxf core already depends on jaxb impl class NamespacePrefixMapper :
 https://github.com/apache/cxf/blob/master/core/src/main/java/org/apache/cxf/common/jaxb/NamespaceMapper.java

jimma:
> As Dan pointed out, the big issue is that your code impose compulsory restrictions on that the CXF runtime must have jaxb-impl jar dependency. Otherwise you run into ClassNotFoundException, as the class name is com.sun.xml.internal.bind.marshaller.CharacterEscapeHandler in JDK version but not com.sun.xml.bind.marshaller.CharacterEscapeHandler, the package name is different. In this case the DataWriter can't be created at all. I don't think we should introduce this compulsory dependency.

@ffang  cxf-core now depends on jaxb impl class NamespacePrefixMapper (
https://github.com/apache/cxf/blob/master/core/src/main/java/org/apache/cxf/common/jaxb/NamespaceMapper.java), so we must have this class in the classpath to compile . cxf-core doesn't work for jdk8 without jaxb impl under classpath.

jimma:
> One more question: does this actually WORK? What happens if you try to send a < or a > character? Those need to be escaped.
@dkulp   Looked at jaxb code again, we may need to look at if setting MinimumEscapeHandler https://github.com/javaee/jaxb-v2/blob/master/jaxb-ri/runtime/impl/src/main/java/com/sun/xml/bind/marshaller/MinimumEscapeHandler.java work for these failed tests with JDK9.  NoEscapeHandler is too much.

ffang:
> @ffang Is this from NewLineEscapeHandler ?
Yes, I refer the NewLineEscapeHandler in the 
./jaxws/src/java.xml.bind/share/classes/com/sun/xml/internal/bind/v2/runtime/output/XMLStreamWriterOutput.java
of jdk9,  as it is the handler to be used if no customized one specified. And I believe we should extend it and only need ensure when text length is 0 we still can call into out.write

dkulp:
@jimma - cxf core requires the jaxb-impl for compile, but not runtime.   There are places where, if we detect we are using the in-jdk jaxb, we change code paths.   In particular, we have spots where we use ASM (if available) to create a NamespaceMapper that subclasses the in-jdk version.

Can I also ask what test is failing?   All the test in CXF seem to pass with the latest Java9 on my Mac.

ffang:
> Can I also ask what test is failing? All the test in CXF seem to pass with the latest Java9 on my Mac.

Hi @dkulp ,

They are
JSONProviderTest.testWriteNullValueAsString
JSONProviderTest.testWriteNullValueAsNull

Both are disabled with java9 currently, and that's the reason why I leave CXF-7520 still open.
I think it's a jdk issue and reported here[1] as it's caused by jdk jaxb impl change,  but JDK team don't think this is something they wanna fix, add direct the customized escapehandling solution.

[1]http://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8188080

ffang:
@jimma yeah, verified specify the MimiumEscapeHandler can fix the 
JSONProviderTest.testWriteNullValueAsString
JSONProviderTest.testWriteNullValueAsNull
with jdk9, however, we need set it in org.apache.cxf.jaxrs.provider.AbstractJAXBProvider when create the marshaller. 
I will go ahead to fix the broken jdk9 test failures based on your suggested solution.

ffang:
@jimma 
Continued test with jaxb-core 2.3.0.
JSONProviderTest.testWriteNullValueAsString
JSONProviderTest.testWriteNullValueAsNull
still failed with jdk9, with MimiumEscapeHandler or your NoEscapeHandler I got same error

jimma:
@ffang  Looks we have to add our customized MininumEscpapeHandler to write a empty string like

 ```
public class CXFMinimumEscapeHandler implements CharacterEscapeHandler {
    private CXFMinimumEscapeHandler() {
    }
    public static final CharacterEscapeHandler HANDLER = new CXFMinimumEscapeHandler();
    public void escape(char[] ch, int start, int length, boolean isAttVal, Writer out) throws IOException {
        if (length == 0 && start == 0) {
            out.write(ch[start]);
            return;
        } else {
            MinimumEscapeHandler.theInstance.escape(ch, start, length, isAttVal, out);
        }
    }
}
```
Besides this new class, we need to add another method in JAXBUtil to set escape handler property to jMarshaller. Then DataWriterImpl and AbstractJAXBProvider can both invoke this util method to set customized escape handler after marshller created.  To support jdk's internal jaxb class, we may need to create a  minimum escape handler class with asm like what we did for NamespaceMapper.  Do you think this will fix all the things ?

ffang:
@jimma about we need extend EscapeHandler to support the empty string, I just found same and got 
JSONProviderTest.testWriteNullValueAsString
JSONProviderTest.testWriteNullValueAsNull
passed with jdk9 + jaxb-core 2.3,  ensure empty string can call into out.write is the key part to reverse the behavior changed in both jdk9 and jaxb-core 2.3.
And I agree that we should add Extended EscapeHandler in cxf-core module, as it has jaxb-core dependency and even a jaxb package there already.

Append a patch what I got so far, this is to fix JSONProviderTest failure with jdk9 oriented so it didn't include your fix in jaxb databinding module
[CXF-7520-20180115.txt](https://github.com/apache/cxf/files/1631328/CXF-7520-20180115.txt)

jimma:
@ffang @dkulp @asoldano  I updated the change to set our customized MinimumEscapeHandler and support jdk internal jaxb, please review.

PR #366, opened 2018-01-11
Handle ResponseExceptionMappers

andymc12:
We should not need to call handleMapper for the ResponseExceptionMappers
 - they already require the parameterized type to be a subclass of 
Throwable. And the handleMapper method checks whether the Throwable
class can be assigned to the type in the user's REM - which it never
will be (unless they define ResponseExceptionMapper<Throwable>).

This also adds unit tests ensuring that the REM is executed.

@johnament - can you review this to make sure I'm on the right track?

### DIFF
--- a/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientProviderFactory.java
+++ b/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientProviderFactory.java
@@ -21,8 +21,8 @@ package org.apache.cxf.microprofile.client;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
-import java.util.LinkedList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import javax.ws.rs.core.Configuration;
 
@@ -91,18 +91,14 @@ public final class MicroProfileClientProviderFactory extends ProviderFactory {
 
     public List<ResponseExceptionMapper<?>> createResponseExceptionMapper(Message m, Class<?> paramType) {
 
-        List<ResponseExceptionMapper<?>> candidates = new LinkedList<>();
-
-        for (ProviderInfo<ResponseExceptionMapper<?>> em : responseExceptionMappers) {
-            if (handleMapper(em, paramType, m, ResponseExceptionMapper.class, true)) {
-                candidates.add(em.getProvider());
-            }
-        }
-        if (candidates.size() == 0) {
+        if (responseExceptionMappers.isEmpty()) {
             return Collections.emptyList();
         }
-        candidates.sort(new ResponseExceptionMapperComparator());
-        return Collections.unmodifiableList(candidates);
+        return Collections.unmodifiableList(responseExceptionMappers
+                                            .stream()
+                                            .map(ProviderInfo::getProvider)
+                                            .sorted(new ResponseExceptionMapperComparator())
+                                            .collect(Collectors.toList()));
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientProviderFactory.java
```diff
@@ -91,16 +91,11 @@ protected void setProviders(boolean custom, boolean busGlobal, Object... provide
...
-        for (ProviderInfo<ResponseExceptionMapper<?>> em : responseExceptionMappers) {
-            if (handleMapper(em, paramType, m, ResponseExceptionMapper.class, true)) {
-                candidates.add(em.getProvider());
-            }
-        }
-        if (candidates.size() == 0) {
+        if (responseExceptionMappers.isEmpty()) {
             return Collections.emptyList();
         }
+        List<ResponseExceptionMapper<?>> candidates =
```
reta: Even shorter: 
```
return Collections.unmodifiableList(responseExceptionMappers
    .stream()
    .map(ProviderInfo::getProvider)
    .sorted(new ResponseExceptionMapperComparator())
    .collect(Collectors.toList()));
```

andymc12: @reta - thanks for the suggestion - I'm making that change and squashing now.

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientProviderFactory.java
```diff
@@ -91,18 +91,14 @@ protected void setProviders(boolean custom, boolean busGlobal, Object... provide
...
 
-        List<ResponseExceptionMapper<?>> candidates = new LinkedList<>();
-
-        for (ProviderInfo<ResponseExceptionMapper<?>> em : responseExceptionMappers) {
-            if (handleMapper(em, paramType, m, ResponseExceptionMapper.class, true)) {
-                candidates.add(em.getProvider());
-            }
-        }
-        if (candidates.size() == 0) {
+        if (responseExceptionMappers.isEmpty()) {
```
johnament: They should never actually be empty, since we require the default one to be registered.

rt/rs/microprofile-client/src/test/java/org/apache/cxf/microprofile/client/CxfTypeSafeClientBuilderTest.java
```diff
@@ -103,41 +108,28 @@ public void testInvokesPostOperationWithRegisteredProviders() throws Exception {
...
+    @Test(expected = NoSuchEntityException.class)
+    public void testResponseExceptionMapper() throws Exception {
+        ExceptionMappingClient client = new CxfTypeSafeClientBuilder()
+            .register(NotFoundExceptionMapper.class)
+            .register(NotFoundClientReqFilter.class)
+            .baseUrl(new URL("http://localhost/null"))
+            .build(ExceptionMappingClient.class);
+
+        Response r = client.getEntity();
+        System.out.println(r.getStatus());
```
johnament: We can probably move the SOP into the `fail` method.

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientProviderFactory.java
andymc12: The only time where it would be empty would be if the user set the `microprofile.rest.client.disable.default.mapper` property to true.  In that case we can avoid the sort code.

rt/rs/microprofile-client/src/test/java/org/apache/cxf/microprofile/client/CxfTypeSafeClientBuilderTest.java
andymc12: Agreed.  Will move.  Thanks for the review!

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientProviderFactory.java
johnament: Ah yes, forgot about the property.

PR #367, opened 2018-01-12
[CXF-7608] Propagate trace ids as it lets log correlation to be consi…

aldex32:
…stent even if not sampling


reta:
Thank you for the fix, @aldex32 . Would you mind please adding a test case for this particular issue into `org.apache.cxf.systest.jaxrs.tracing.brave.BraveTracingTest`? Thanks!

aldex32:
Hi @reta , I tried to do an integration test in the test class you mentioned above but didn't succeeded.
What I did is:
* Created a server with brave feature with sampling `NEVER_SAMPLER`
* Invoke one of `bookstore` API
* Then created a `POST_INVOKE` interceptor which would get the request attribute `org.apache.cxf.tracing.brave.span` and check if the `TraceScope#SpanInScope` is in `AbstractTracingProvider.TraceScopeHolder`.
The issue I faced is that I cannot cast `((HttpServletRequest)message.get("HTTP.REQUEST")).getAttribute("org.apache.cxf.tracing.brave.span")` to `TraceScopeHolder` since it is a protected and then I cannot assert on it.

Do you have any idea how should I test it?

reta:
@aldex32 absolutely, here is the test case I have in mind:
```
    @Test
    public void testThatNoSpansAreRecordedWnenNotSampled() {
        final Tracing never = Tracing
            .newBuilder()
            .reporter(new TestSpanReporter())
            .sampler(Sampler.NEVER_SAMPLE)
            .build();

        final Response r = createWebClient("/bookstore/books", new BraveClientProvider(never)).get();
        assertEquals(Status.OK.getStatusCode(), r.getStatus());

        assertThat(TestSpanReporter.getAllSpans().size(), equalTo(0));
        assertThatTraceHeadersArePresent(r, false);
    }
```
If we run it against the current master, it will fail and still record some spans. However, with your fix the trace / spans are correctly apply the sampler rules (in this case - never sampler) and nothing is being recorded. Please let me know if it makes sense to you. Thanks!

aldex32:
@reta Test added. Thanks!

PR #368, opened 2018-01-19
Ensure that the JAX-RS CompletionStage uses the correct ExecutorService

andymc12:
Clients can specify an ExecutorService via the property() method, but
currently, that ES is only used for the initial rx() method. When the
returned CompletionStage is invoked with a *Async() method, the
CompletionStage uses the ForkJoinPool.commonPool since that is the
default executor for CompletableFuture.  Instead, it ought to use the
same executor specified in the client properties as the default - keep
in mind that users can override the default by specifying a different
executor.

### DIFF
--- a/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/CompletionStageRxInvokerImpl.java
+++ b/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/CompletionStageRxInvokerImpl.java
@@ -146,7 +146,7 @@ public class CompletionStageRxInvokerImpl implements CompletionStageRxInvoker {
         if (ex == null) {
             return CompletableFuture.supplyAsync(() -> wc.sync().method(name, entity, responseType));
         }
-        return CompletableFuture.supplyAsync(() -> wc.sync().method(name, entity, responseType), ex);
+        return CxfCompletableFuture.supplyAsync(() -> wc.sync().method(name, entity, responseType), ex);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


andymc12:
Looking at this a little more, this is not a complete solution... This will work for the _next_ stage, but not subsequent stages because the `CompletableFuture` class will start returning instances of itself that continues to use the `ForkJoinPool.commonPool` rather than the specified Executor.  I'll try to get a little more work done on this after the weekend.

reta:
@andymc12 I don't think this is a problem we need to solve, this is how `CompletableFuture` works, and I believe the expectations are set. So `rx` stage does respect the executor, the consequent calls to `thanXxxAsync` should provide the executor explicitly as it becomes the responsibility of the caller to decide where the execution should continue.

johnament:
@andymc12 I kind of agree with @reta (but for a slightly different reason I think..), since we're using the static method that passes in the `ExecutorService`, the same `ExecutorService` will be used for the remaining invocations of the methods here, while CXF is in control.  By definition, a user shouldn't be using `supplyAsync` that doesn't take an `ExecutorService` afterwards; if this is what they're after.

What we do probably have to make happen is a way to get the `ExecutorService` from the underlying client, if the user doesn't have access to it already (maybe they can get via the `Configuration` class?)

andymc12:
@johnament @reta That makes, given that the `*Async` methods all allow have a variation that allows the user to pass in an executor.  And they should be able to get the executor service used in the client via the `executorService` property (see `AbstractClient.EXECUTOR_SERVICE_PROPERTY`) in the config (in CXF anyway - not sure about other JAX-RS implementations).  My initial reading of the spec led me to believe that the default executor for subsequent CompletionStages should be the same as the initial one, but as long as the user can continue to re-use that executor, I think we can close this PR.  Thanks for the review!

PR #369, opened 2018-01-20
[CXF-7610] - Adding SPI to handle customizations to the server bean.

johnament:
I'm only marking this as a WIP for now, to make sure this makes sense first.  Still need to add some tests and port the existing SSE code to use this.


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/ResourceUtils.java
```diff
@@ -902,9 +904,18 @@ public static JAXRSServerFactoryBean createApplication(Application app,
             bean.getProperties(true).putAll(appProps);
         }
         bean.setApplication(app);
-
+        customize(bean);
         return bean;
     }
+
+    private static void customize(JAXRSServerFactoryBean bean) {
+        ServiceLoader<JAXRSServerFactoryCustomizationExtension> extensions
```
rmannibucau: shouldn't it be loaded from the bus as other extensions?

johnament: Not sure what you mean.

rmannibucau: bus.getExtension

main issue is ServiceLoader doesnt work accross environments cxf works so cxf has another "portable" abstraction

johnament: If you're talking about `ExtensionRegistry`, the javadoc on that indicates OSGi preference.

rmannibucau: META-INF/cxf/bus-extensions.txt / Extension (check ExtensionManagerImpl). Thinking about it we already have events on the bus so it can be used to customize the server without requiring any new API or specific kind of SPI

johnament: I dunno, don't forget that this is simply another way to automatically discover resources.  From the input I've received, it seems like we do almost no automatic discovery (outside of CXF specific feature objects).  Note that I'm also not removing the CDI automatic extension loading, just adding to it that `ServiceLoader` can be used, and that should be the default supported by SSE (which is really where this all came from, the SSE JAR incorrect has a `bean-discovery-mode=all` declared)

johnament: My understanding of those files is that they are all singular, you only have one instance.  But perhaps you are on to something, but what I don't see is how to get the bean just by implementing `BusLifeCycleListener`

johnament: BTW, [this is the relevant thread on dev@](https://lists.apache.org/thread.html/4346a742a4179129cd6d9e4c7edd60910404871f745a09842a793455@%3Cdev.cxf.apache.org%3E)

rmannibucau: was more thinking about adding a sendEvent passing the bean as parameter, this way you implement a standard listener and can customize the bean and register the listener or not as you need. The big advantage is to not have to add a SPI which will not work in some environments and will require a deactivation mecanism since jar can always end up in the classpath in an undesired fashion even in a modular project (you dont always have the choice of the deployment).

reta: As I remember, the concern @johnament has about current implementation is `JAXRSServerFactoryCustomizationExtension` being a CDI bean. If we re-implement customization using events, we will still need a CDI bean to handle the event, so from my perspective, not buying much.

Regarding the scope, I think we can limit it to CDI-only (essentially just doing the service loader calls where they are needed, in CDI extension). I believe it would address some of your concerns, @rmannibucau? I would also prefer not to generalize it now because we don't have a real use case beyond CDI actually. Or do you have, @johnament ?

rt/rs/sse/src/main/resources/META-INF/beans.xml
```diff
@@ -2,5 +2,5 @@
 <beans xmlns="http://java.sun.com/xml/ns/javaee"
 	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
 	xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/beans_1_1.xsd"
-	bean-discovery-mode="all">
```
reta: Agree than `all` was not correct here, but I would like SSE to be discoverable with no efforts when run in CDI container. I think we need only one single class to be discoverable , `SseFeature`, so @johnament could we add some exclusion / inclusion rules instead of just cutting everything off?

johnament: Since `SseFeature` is a core CXF feature, not a JAX-RS feature, I believe its autoloaded that way.  Isn't it?

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/ResourceUtils.java
johnament: So move the service loader into CDI extension?  You're correct, all of my current use cases are CDI as the DI container, so it would solve my issue.

rt/rs/sse/src/main/resources/META-INF/beans.xml
reta: It is core CXF feature, right, and is it automatically discovered by CDI extension (as an annotated type). But I believe if we set `bean-discovery-mode="none"`, it won't be automatically discovered  anymore, no?

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/ResourceUtils.java
reta: Yes, that would be better option I think, thanks.

rmannibucau: @reta not sure i get it. I was speaking of cxf extension and events, not cdi ones

rt/rs/sse/src/main/resources/META-INF/beans.xml
johnament: I'll test it out and get back to you.  If it doesn't work, I'd actually prefer to move it into the extension.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/ResourceUtils.java
johnament: @reta thought about this a bit.  I want to leave the method that discovers and executes the extensions here, but invoke it from the CDI extension.  The reason being, if others want to leverage the `ServiceLoader` pattern, they can without having to rewrite it.

@rmannibucau I'm pretty sure the whole reason this feature exists in the first place is because that CDI event pattern doesn't work properly (in this use case).  BTW, does meecrowave use the CDI extension for CXF?

reta: @rmannibucau Sorry, got confused by events in CDI. I think it is possible to implement the customization using CXF events along with bus extension. However, we don't have event types which are fired at the moment when we have to apply the customization (in case of SSE at least) so we would have to add one. But in general it is doable with a bit more code changes.

reta: @johnament That's the thing, right, if you leave it here it should become a public instance method of the `JAXRSServerFactoryBean`. Than the question would be, it is there but never being called inside `JAXRSServerFactoryBean`, what the purpose of this method? I think we could always move it back if the use case pops up. Make sense?

rmannibucau: Yes meecrowave uses the extension.

The fact to not use a SPI but a native cxf loading makes the feature generic IMHO. If we dont care about it as a feature then just testing if sse jar is here (loadClass is fine) and set the sse transportid is enough and doesnt require any SPI.

johnament: @rmannibucau I'm planning to use this same approach for adding MP OpenTracing support as well; TBD though if that's going to be part of Apache CXF.

@reta the method right now is on `ResourceUtils`.  I was going to pull this out into something like a `JAXRSServerFactoryCustomizationUtils`, so the question is whether this should sit in the CDI module or the `jaxrs-frontend` module.  Obviously my vote is `jaxrs-frontend`

reta: `JAXRSServerFactoryCustomizationUtils` in `jaxrs-frontend` sounds good, thanks

rt/rs/sse/src/main/resources/META-INF/beans.xml
johnament: I ended up going the route of adding the feature's behavior into the extension.  Thoughts?

PR #372, opened 2018-01-31
[CXF-6134] Apache CXF generating constructor with duplicate argument names causing compilation error

ivy-rew:
reference qualified 'java.lang.Throwable' in order to avoid compile conflicts with 'Throwable' type in the same package


ivy-rew:
please run checks again. seems like random failures to mee...

deki:
yes don't worry about it

PR #373, opened 2018-02-01
JAX-RS 2.1 TCK issues

andymc12:
In testing the JAX-RS 2.1 TCK, we've found a few issues:
1) The `Headers.toHttpLanguage(...)` method will return things like "en-" in some cases - I believe this is because there is an empty (but non-null) country passed in as part of the Locale.  This seems like an error condition to me, but it seems to be tested by the TCK, so the fix is pretty simple - just checking whether the country is non-empty rather than just non-null.
2) The javadoc for `Invocation.Builder.headers(...)` method says that when called, all pre-existing headers must be removed.  Currently, we're just adding the new headers, but not removing the old ones.  

I provided test cases for these conditions, but I won't have confirmation that these fixes actually pass the TCK until tomorrow, but I wanted to get the PR out now to ensure that the fixes get into the 3.2.2 release.  If the fixes are unsuccessful, I'll abort this PR and try again after 3.2.2 is released.  If they are successful (and assuming no glaring issues uncovered during review), then I'll plan to merge ASAP.

Thanks, Andy

### DIFF
--- a/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/spec/InvocationBuilderImpl.java
+++ b/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/spec/InvocationBuilderImpl.java
@@ -253,10 +253,13 @@ public class InvocationBuilderImpl implements Invocation.Builder {
 
     @Override
     public Builder headers(MultivaluedMap<String, Object> headers) {
-        RuntimeDelegate rd = HttpUtils.getOtherRuntimeDelegate();
-        for (Map.Entry<String, List<Object>> entry : headers.entrySet()) {
-            for (Object value : entry.getValue()) {
-                doSetHeader(rd, entry.getKey(), value);
+        webClient.removeAllHeaders();
+        if (headers != null) {
+            RuntimeDelegate rd = HttpUtils.getOtherRuntimeDelegate();
+            for (Map.Entry<String, List<Object>> entry : headers.entrySet()) {
+                for (Object value : entry.getValue()) {
+                    doSetHeader(rd, entry.getKey(), value);
+                }
             }
         }
         return this;
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/Headers.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/Headers.java
@@ -43,7 +43,6 @@ import javax.servlet.http.HttpServletResponse;
 
 import org.apache.cxf.common.logging.LogUtils;
 import org.apache.cxf.common.util.PropertyUtils;
-import org.apache.cxf.common.util.StringUtils;
 import org.apache.cxf.helpers.CastUtils;
 import org.apache.cxf.helpers.HttpHeaderHelper;
 import org.apache.cxf.message.Message;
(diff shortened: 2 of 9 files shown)


andymc12:
@sberyozkin I agree with your update.  I'm not sure whether the clone is required for somewhere else though.  I figured the safest thing was a new method to clear the headers, so as to not disrupt other code flows.  

I did make a couple more changes though - the spec also says that the user can pass in a null headers map, which means that we need to just clear the headers - so I added a null check.  I also changed the `removeAllHeaders` method to clear the headers map rather than creating a new one.

Thanks for the review!

rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/spec/InvocationBuilderImpl.java
```diff
@@ -253,10 +253,13 @@ public Builder header(String name, Object value) {
 
     @Override
     public Builder headers(MultivaluedMap<String, Object> headers) {
-        RuntimeDelegate rd = HttpUtils.getOtherRuntimeDelegate();
-        for (Map.Entry<String, List<Object>> entry : headers.entrySet()) {
-            for (Object value : entry.getValue()) {
-                doSetHeader(rd, entry.getKey(), value);
+        webClient.removeAllHeaders();
+        if (headers != null) {
```
johnament: does the TCK check that `.headers(null)` should clear all headers?  Seems odd to me.

rt/transports/http/src/main/java/org/apache/cxf/transport/http/Headers.java
```diff
@@ -588,9 +589,10 @@ public static String toHttpDate(Date date) {
     public static String toHttpLanguage(Locale locale) {
         StringBuilder sb = new StringBuilder();
         sb.append(locale.getLanguage());
-        if (locale.getCountry() != null) {
+        String country = locale.getCountry();
+        if (!StringUtils.isEmpty(country)) {
```
johnament: what about `Locale.toString().replace('_','-');`?

rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/spec/InvocationBuilderImpl.java
andymc12: I agree it is odd... I'm not sure if the TCK tests it or not, but the [javadoc](https://jax-rs.github.io/apidocs/2.1/javax/ws/rs/client/Invocation.Builder.html#headers-javax.ws.rs.core.MultivaluedMap- ) says that if the passed-in map is null, then all existing headers should be removed.

rt/transports/http/src/main/java/org/apache/cxf/transport/http/Headers.java
andymc12: Yes!  That is much cleaner.  Thanks!

andymc12:
I added a new commit.  Some SSE systests were failing because the SSE code was calling `builder.headers(...)` which was clobbering some required headers.  This commit changes it to only add the one header rather than replacing them all.  With this last commit, the SSE systests pass for me locally.  Assuming they pass Jenkins here, I think this PR should be good to merge.

Thanks All!

andymc12:
And one more TCK fix.  CXF was failing with a ClassCastException when the TCK added a new non-String object header in a ClientRequestFilter.  CXF converts these header objects to Strings earlier in the flow, but does not convert them after filters have been invoked.  When CXF goes to put those non-String headers on the wire, it fails because it cannot cast them a String.  This fix converts the headers after invoking the filter method.

PR #374, opened 2018-02-01
[CXF-7628]

1984shekhar:
Enable JMX connector and threading-parameters mbeans for endpoints with custom threading parameters configured in blueprint xml.


1984shekhar:
There are multiple commit which are unnecessary.

PR #380, opened 2018-02-09
[CXF-7639] Update dependency to servlet 4.0

andymc12:
Updating the Import-Package of javax.servlet.* packages to include 4.0.  This is a small change, but is used in several modules.  I'm mainly looking for an automated build to confirm that this change doesn't break anything, but review/suggestions welcome.  Thanks!


andymc12:
There are two failures, but they definitely seem unrelated to the change in the OSGi export/import versions.

andymc12:
https://issues.apache.org/jira/browse/CXF-7639

PR #381, opened 2018-02-11
[CXF-7642] Introduce separate modules for RxJava & RxJava2

johnament:
- Introduce built in customizer for RxJava2


rt/rs/extensions/reactor/src/main/java/org/apache/cxf/jaxrs/reactor/server/ReactorCustomizer.java
```diff
@@ -0,0 +1,39 @@
...
+public class ReactorCustomizer implements JAXRSServerFactoryCustomizationExtension {
+    @Override
+    public void customize(JAXRSServerFactoryBean bean) {
+        Boolean useStreamingSubscriber = (Boolean)bean.getProperties(true)
+                .getOrDefault("useStreamingSubscriber", false);
+        ReactorInvoker invoker = new ReactorInvoker();
+        invoker.setUseStreamingSubscriberIfPossible(useStreamingSubscriber);
+        bean.setInvoker(invoker);
+        StreamingResponseProvider<Object> streamProvider = new StreamingResponseProvider<>();
+        streamProvider.setProduceMediaTypes(Collections.singletonList("application/json"));
```
reta: Minor, may be `MediaType.APPLICATION_JSON` instead of `application/json`?

reta:
Looks good to me, @johnament. Extracting RxJava2 into standalone extension is certainly a good idea.

rt/rs/extensions/reactor/src/main/java/org/apache/cxf/jaxrs/reactor/server/ReactorCustomizer.java
johnament: Agreed.  Actually, after I pulled this out I decided I really need a base class for these extensions.    So I'm going to take care of that before merging.

reta: Sounds good, thanks!

PR #384, opened 2018-02-23
[CXF-5052] Classpath references should be understood using wsdlRoot batch processing options in cxf-codegen

oli-ver:
First proposal for a classpath replacement of local file url in generated class file.

A wsdlPath set in defaultOptions beginning with classpath;<path to replace> replaces the file url "file:/path/to/package1/package2/myfile.wsdl" with "classpath:/package1/package2/myfile.wsdl".

Just the first idea on how to achieve this, I am very open for different ideas.


deki:
Thanks for your PR. Could you please add a testcase to the src/it folder? There are already two testcases so just copy and adapt one of them...

oli-ver:
@deki I added a testcase, but I am not sure, how it should be executed. Nevertheless: When I run mvn generate-sources in the folder src/it/wsdl-classpath-resolution the example generates classes with classpath URLs.

deki:
@oli-ver this could be part of the 3.2.5 release. are you able to finish it?

oli-ver:
@deki Sorry for my contiously late answers. What is the due date for 3.2.5? Are you ok with the solution I suggested here, when the tests do work? I personally think it is better to have a special XML node for this, but I would have to figure out how to create new ones first and find time to implement it. Sorry, at the moment there are too many other things going on :)

deki:
The release is planned for today. 

Regarding your solution: Why not using classpath: instead of classpath;?

oli-ver:
I did not want the expression to look like a classpath, because it‘s only a replacement string. Do you think classpath: would be a better syntax?

deki:
Ah ok got it. I this case a separate tag would be cleaner to avoid confusion.

gnodet:
Closing: this PR is 8 years old with no activity since 2018. The `classpath;` prefix syntax was questioned during review — a cleaner approach with a dedicated XML parameter would be preferred. CXF-5052 remains open for a future clean implementation.

PR #385, opened 2018-02-26
[CXF-7657] Reader and Writer interceptors are not sorted by the provi…

csierra:
…der comparator


csierra:
hi @andymc12, 

I can try but the problem is that the failures are random because the fallback case when the providers come with the same `@Priority` is ```Integer.compare(o1.hashCode(), o2.hashCode())```

csierra:
hey @andymc12, 

sorry this took so long. I updated the test showing the problem with the ReaderInterceptor and WriterInterceptor and the `Comparator`. If you revert the fix you will see the test failing. There is a small chance, though, that the test passes because the current behaviour is random, so it might happen that the writers fall in the expected order even though they are not being sorted.

PR #386, opened 2018-02-27
CXF-7659: Improve SSE automatic configuration. 

reta:
Enhanced the default transport selection mechanism to consult bus property and already registered destination factories. It gives customization extensions (like SSE for example) to hint about preferred transport to use in case none of them is specified explicitly.


reta:
@johnament Would you mind taking a quick look? Thanks.

PR #387, opened 2018-03-07
[CXF-7640] Create a form to set the use of Spring in the classHelper …

osvaldopina:
…on a per client way


osvaldopina:
The build was successful in my machine. My config is:

- Apache Maven 3.5.2 (138edd61fd100ec658bfa2d307c43b76940a5d7d; 2017-10-18T05:58:13-02:00)

- Java version: 1.8.0_72, vendor: Oracle Corporation

- OS name: "mac os x", version: "10.12.6", arch: "x86_64", family: "mac"

What is the jenkins build environment so I can try to reproduce it locally and fix the build problem?

deki:
Don't worry about the Jenkins build, the failure is caused by a flapping test not related to your changes.

core/src/main/java/org/apache/cxf/common/util/ClassHelper.java
```diff
@@ -28,9 +28,15 @@
  *
  */
 public class ClassHelper {
+
+    public static final String USE_DEFAULT_CLASS_HELPER = "org.apache.cxf.useDefaultClassHelper";
+
     static final ClassHelper HELPER;
+    static final ClassHelper CXF_HELPER;
```
reta: Probably better name would be `DEFAULT_HELPER`?

core/src/main/java/org/apache/cxf/common/util/ClassHelper.java
```diff
@@ -28,9 +28,15 @@
  *
  */
 public class ClassHelper {
+
+    public static final String USE_DEFAULT_CLASS_HELPER = "org.apache.cxf.useDefaultClassHelper";
```
reta: Could you please rename it to `org.apache.cxf.useDefaultClassHelpers`  to follow similar `Spring` property name, thanks.

core/src/main/java/org/apache/cxf/common/util/ClassHelper.java
```diff
@@ -28,9 +28,15 @@
...
+
     static final ClassHelper HELPER;
+    static final ClassHelper CXF_HELPER;
+
     static {
         HELPER = getClassHelper();
```
reta: I think more often than not we will create 2 instances of `ClassHelper` (HELPER  and DEFAULT_HELPER would be the same), I would suggest to do that:

```
DEFAULT_HELPER = new ClassHelper();
HELPER = getClassHelper(DEFAULT_HELPER);
```
And

```
private static ClassHelper getClassHelper(ClassHelper defaultHelper) {
 ...
    return defaultHelper;
}
```

So we would be using the same instance in case Spring helpers are not being asked for.

core/src/main/java/org/apache/cxf/common/util/ClassHelper.java
```diff
@@ -69,19 +76,44 @@ protected Object getRealObjectInternal(Object o) {
     }
 
     public static Class<?> getRealClassFromClass(Class<?> cls) {
```
reta: I would suggest to add the overloaded method with accepts `Bus bus`, there are many places where it could be provided from context. What do you think?

core/src/main/java/org/apache/cxf/common/util/ClassHelper.java
```diff
@@ -69,19 +76,44 @@ protected Object getRealObjectInternal(Object o) {
...
-        return HELPER.getRealClassFromClassInternal(cls);
+        Bus bus = getBus(null);
+        if (checkUseDefaultClassHelper(bus)) {
```
reta: We could get rid of some duplication here:

````
private static getContextClassHelper(Bus bus) {
    // We could check if DEFAULT_HELPER and HELPER  are referencing the same instance and than the bus property
    return (DEFAULT_HELPER == HELPER || checkUseDefaultClassHelper(bus)) ? DEFAULT_HELPER : HELPER;
}
```

reta:
@osvaldopina Could you please add at least one (better, a few) test case to make sure things are working as expected? Thanks.

osvaldopina:
all suggested changes were done

reta:
retest this please

PR #388, opened 2018-03-07
[CXF-7670] create a single ClassResourceInfo per class + annotated method

K0zka:
In each resource class, an annotated method should be registered only once.


K0zka:
.

PR #390, opened 2018-03-12
[CXF-7676] Create a proxy OutputStream to create an EntityStream that…

WhiteCat22:
… a user's ClientRequestFilter can interact with and holds a reference that we can set our URLConnectionWrappedOutputStream to later.


WhiteCat22:
Currently creating a systest, will add before merge

WhiteCat22:
Test has been created and is passing.

PR #393, opened 2018-03-18
CXF-7643 ensure CDI extension defaults are CDI compliant

rmannibucau:
This PR tries to make defaults of the CDI extension better, here are the changes:

1. if it detects OWB-web or weld-web it deactivate default context beans for HttpServletRequest and ServletContext
2. it observes beans to detect the types conflicting with "auto" context beans
3. it fires the extension and exposes the list of exclusion to let integration layers configure it properly if 1/2 defaults are not enough 
4. when it adds context beans it removes @Default from the ones which are found with 1/2/3 - but it keeps it with the cxf specific qualifier if needed (best practise being to use this classifier for internal/extensions needs)


johnament:
Can you update one of the existing systests to confirm that the various injection points are available, with the needed dependencies on the classpath?

rmannibucau:
@johnament done

PR #396, opened 2018-03-23
[CXF-7673] MP Rest Client 1.1 Impl

andymc12:
Includes async method support and tests, updated TCK, etc.

This will still require some changes - particularly in the pom.xml files since the MP Rest Client 1.1 APIs and TCK has not yet been released.  

I'm looking for confirmation that my changes in the rt/rs/client module are acceptable - and an overall review of the async functionality.

Thanks in advance!


rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/ClientProxyImpl.java
```diff
@@ -285,7 +285,8 @@ private static boolean isIgnorableParameter(Method m, Parameter p) {
             return true;
         }
         return p.getType() == ParameterType.REQUEST_BODY
-            && m.getParameterTypes()[p.getIndex()] == AsyncResponse.class;
+            && (m.getParameterTypes()[p.getIndex()] == AsyncResponse.class
```
johnament: this would make it so that any client proxy can use `InvocationCallback`.  Is it worth adding sys tests that verify this?

systests/microprofile-tck/pom.xml
```diff
@@ -0,0 +1,137 @@
+<?xml version="1.0"?>
```
johnament: This seems like its already covered by `/systests/microprofile/client/weld/` ?  I had intended to add client tests for weld and owb at some point, and if CXF implements other specs (talk around OpenAPI support in a JIRA ticket) then it could all live in one tree.

rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/ClientProxyImpl.java
andymc12: Good point.  This probably should be isolated to just MP Rest Clients.  I'll try to refactor this to avoid impacting non-MP proxy clients.

systests/microprofile-tck/pom.xml
andymc12: Right - I tried to move those tests to another systest tree, but something weird happened with git - instead of showing that they moved, it shows that the files were deleted and then re-created.  I'll try to fix this so that the history is more accurate.  And I will try to refactor it so that they can all live under the same directory tree.  Thanks!

andymc12:
@johnament I've updated the PR with your changes.  The build is failing because I am using a non-published build of MP Rest Client 1.1.  I really need to get a release candidate out to build against properly... there is still a question on how the async methods should be implemented in MP Rest Client PR [85](https://github.com/eclipse/microprofile-rest-client/pull/85).

andymc12:
@johnament - I know it has been a while, but could take another look at the changes here?  I've updated them to use public 1.1 API/TCK and I can get the TCK to pass locally - note that I had to use a testng.xml file to work around an issue with URI invocation checking (fixed by MP Rest Client PR [108](https://github.com/eclipse/microprofile-rest-client/pull/108) ).  Thanks!

systests/microprofile/client/async/pom.xml
```diff
@@ -0,0 +1,137 @@
...
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.apache.cxf.systests</groupId>
+    <artifactId>cxf-microprofile-async</artifactId>
+    <name>Apache CXF MicroProfile Async Sys Tests</name>
+    <description>Apache CXF System Tests - MicroProfile Rest Client Async Tests</description>
+    <url>http://cxf.apache.org</url>
+    <properties>
+        <cxf.geronimo.config.version>1.0</cxf.geronimo.config.version>
```
reta: Could you please add the `cxf.module.name` to the `<properties>` section:
```
        <cxf.module.name>org.apache.cxf.systests.microprofile.async</cxf.module.name>
```
Thanks!

andymc12: @reta done.  thanks for the review.

reta: :+1: thanks!

PR #400, opened 2018-04-05
[CXF-7696] Validate audience restriction using wildcard

dreamweaver360:
Implemented using wildcard with explicit match to take precendence


coheigea:
The patch doesn't apply with the latest master code. I think it would be better to introduce a new configuration variable to support wildcard matching, rather than re-using JwtConstants.EXPECTED_CLAIM_AUDIENCE.

gnodet:
Closing: this PR is 8 years old with merge conflicts. The regex-based wildcard approach has security concerns (unescaped dots, no anchors). If wildcard audience validation is still needed, it should be reimplemented with proper URL pattern matching and a dedicated configuration flag as suggested by @coheigea. CXF-7696 remains open for a future clean implementation.

PR #405, opened 2018-04-13
CXF-7708 Options mark-generated and suppress-generated-date don't effect on generating Java classes

dmvolod:


deki:
@typekpb fyi

typekpb:
Well, I've noticed that with the latest version of cxf it didn't work. So
would be very thankful for the fix.

deki:
Hmm I wonder why does the unittest not fail? If this is a different case, could you please add one for it so that we are covered in future?

deki:
Ok, would it be possible for one of you to add a unit test, please?

dmvolod:
Yes, working on it for codegen plugin, but looks like no enforce plugin configured for build and integration test fails don't broke build pipeline. Also it contains jdk6 test which is always fails and i don't know is it actual or not.

deki:
I think the jdk6 testcase is no longer valid for 3.2.x but we should probably have one for toolchains. I try to modify it and switch the invoker-plugin to break the build afterwards...

deki:
Ok the build now fails but I had to disable the toolchain test due to CXF-7714.

dmvolod:
@deki , i've added it and fix current build process. Please review.

PR #408, opened 2018-04-20
[CXF-7317] Non-URL encoded Content-Id href does not seem be serialized by CXF's AttachmentSerializer

rnetuka:
JIRA: https://issues.apache.org/jira/browse/CXF-7317
downstream: https://issues.jboss.org/browse/JBWS-4064

According to https://tools.ietf.org/html/rfc2392, Content-ID header shouldn't be decoded

> For example, "cid:foo4%25foo1@bar.net" corresponds to Content-ID: \<foo4%25foo1@bar.net\>

See comments in JBWS-4064 for more info


rnetuka:
@asoldano done

PR #409, opened 2018-04-20
CXF-7710: ClientImpl is memory-leak prone

facundovs:


facundovs:
Hi  @deki . What do you think about using [thread id](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#getId--) instead?

dkulp:
Nothing about this change will actually work.   Changing to a WeakHashMap<String, ....> and using a "new String(...)" for the key will result in nothing actually holding onto the key and the context could be garbage collected immediately.    The test case you added (requestContextIsGarbageCollected) shows the exact problem.  Two calls to getRequestContext() on the same thread should return the same context.    It's very common to do:
```
client.getRequestContext().put("A", aValue);
client.getRequestContext().put("B", bValue);
client.getRequestContext().put("C", cValue);
```
and that NEEDS to work.    If something triggers a GC between there, values will get lost.

As part of CXF-7591, we made sure the  getResponseContext().clear() will work so that users can explicitly clear the context when done with it to avoid this problem.

dkulp:
I think adding a method to Client like:
```
    default AutoCloseable autoCloseContexts() {
        return new AutoCloseable() {
            @Override
            public void close() throws Exception {
                getRequestContext().clear();
                getResponseContext().clear();
            }
        };
    }
```
might be a useful idea.   In the code, you could use the try with resources to make sure the contexts are completely cleared:
```
try (((Client)proxy).autoCloseContexts()) {
   proxy.callWhateverMethods(....);
}
```

Thoughts?

core/src/main/java/org/apache/cxf/endpoint/ClientImpl.java
```diff
@@ -1083,12 +1087,29 @@ public void clear() {
...
         }
     }
 
+    /**
+     * Returns the current thread name. As this value will be used as a key in a {@link WeakHashMap},
+     * we should use the String constructor for avoiding scenarios where strong references are kept
+     * to the string names causing that the map entries can't be garbage collected.
+     * @return
+     */
+    private String getThreadName() {
```
rmannibucau: you probably want to use the thread id and not the name which doesn't have to be unique?

rmannibucau:
Why not closing the client for such cases if issue is the threadlocals, most of them can be cleaned up once used and re-initialied from the client at next request no? It avoids new API.

dkulp:
Closing the entire client is very different (which we already support).  When close on the client is called, EVERYTHING gets shutdown and cleaned up including conduit connections, pointers to the WSDL's, endpoints, etc...  Basically, you are saying the client is no longer going to be used.

rmannibucau:
Yes, basically all the volatile (threadlocal) states could be closed here and the longer time cache (wsld etc) can be put in the bus or equivalent (by default) and therefore keep the client.close fast and the overall client efficient. Issue with these new API is that you can't write a portable app anymore. This is my main concern.
Ensuring the client set and reset the state properly is surely saner for all consumers (native cxf, jaxws, jaxrs), no?

core/src/main/java/org/apache/cxf/endpoint/ClientImpl.java
johnament: Also, you're not going to want to `new` the string each time, for performance reasons.  Can the methods that call this reduce the number of times they call it?

fsgonz:
I am taking the issue from @facundovs and I was reviewing the fix introduced in CXF-7591 as @dkulp mentioned. At first I thought that that may provide the users with a way to clear the context using: getResponseContext().clear() (as you said) , but I find that in every case where responseContext.put is used the clear method is not overridden so the when I invoke client.getResponseContext().clear(), the map associated to the thread key is removed but not the the entry with the thread. 
In the case where a ThreadPool is used, a WeakHashMap will make no difference.
Shouldn't the clear method of the hashmap always be overridden? For example here:

https://github.com/apache/cxf/blob/1396f046333108ec0e7819bf4de154505ff4564c/core/src/main/java/org/apache/cxf/endpoint/ClientImpl.java#L338

gnodet:
Closing this PR — CXF-7710 has been properly resolved on main through a superior approach: `releaseThreadContexts()` + an `AutoCloseable` `Client.Contexts` API that allows try-with-resources cleanup. The approach in this PR (using `new String(threadName)` as `WeakHashMap` key) was identified by reviewers as fundamentally flawed — the keys have no strong references and can be GC'd immediately. Thank you for the contribution\!

PR #412, opened 2018-04-25
[CXF-7726] allows targetSelector to be set in CircuitBreakerFailoverFeature

Juwit:
Adds a setTargetSelector method to CircuitBreakerFailoverFeature


coheigea:
Why not instead just remove the targetSelector from CircuitBreakerFailoverFeature and override the get method instead, e.g. something like:

@Override
    public FailoverTargetSelector getTargetSelector() {
        if (super.getTargetSelector() == null) {
            super.setTargetSelector(new CircuitBreakerTargetSelector(threshold, timeout,
                                                                   super.getClientBootstrapAddress()));
        }
        return super.getTargetSelector();
    }

PR #418, opened 2018-05-22
CXF-7742 handling scopes in CdiResourceProvider

rmannibucau:


integration/cdi/src/main/java/org/apache/cxf/cdi/JAXRSCdiResourceExtension.java
```diff
@@ -303,6 +306,10 @@ public void release(@Observes final BeforeShutdown event) {
             for (final CreationalContext<?> disposableCreationalContext: disposableCreationalContexts) {
                 disposableCreationalContext.release();
             }
+            disposableCreationalContexts.clear();
+        }
+        synchronized (disposableResourceProviders) {
+            disposableResourceProviders.forEach(p -> p.getLifecycle().destroy());
```
reta: `disposableResourceProviders.clear()`?

integration/cdi/src/main/java/org/apache/cxf/cdi/JAXRSCdiResourceExtension.java
```diff
@@ -372,7 +379,14 @@ private ClassifiedClasses classes2singletons(final Application application, fina
 
             for (final Bean< ? > bean: serviceBeans) {
                 if (classes.contains(bean.getBeanClass())) {
-                    classified.addResourceProvider(new CdiResourceProvider(beanManager, bean));
+                    final CdiResourceProvider provider = new CdiResourceProvider(beanManager, bean);
+                    // if not a singleton we manage it per request
+                    // if @Singleton the container handles it
+                    // so we only need this case here
+                    if (provider.isSingleton() && Dependent.class == bean.getScope()) {
+                        disposableResourceProviders.add(provider);
```
reta:
```
synchronized (disposableResourceProviders) {
    disposableResourceProviders.add(provider);
}
```

reta:
@rmannibucau Looks good, a few minor comments, I am wondering if it worth creating a test case for different resource providers?

integration/cdi/src/main/java/org/apache/cxf/cdi/JAXRSCdiResourceExtension.java
rmannibucau: Oh good catch, copied the pattern of disposableCreationalContexts but actually we don't need that for that list which is just used in the context of the extension so thread safe, will clean up

rmannibucau:
@reta I updated the PR, I split the provider for singletons and per request handling, it looks easier to me to follow what is done and avoid to be tempted to be back to something not thread safe, I added a test on the scope handling too

integration/cdi/src/main/java/org/apache/cxf/cdi/SingletonResourceProvider.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.apache.cxf.jaxrs.lifecycle.ResourceProvider;
+import org.apache.cxf.message.Message;
+
+public class SingletonResourceProvider implements ResourceProvider {
+    private final Class<?> clazz;
+    private Object instance;
+
+    SingletonResourceProvider(final Lifecycle lifecycle, final Class<?> clazz) {
+        this.instance = lifecycle.create();
```
reta: Should we create instance proactively here? It is simpler like that but it may not be needed (in theory, at least)

reta:
@rmannibucau Looks better like that, with providers split between different classes

integration/cdi/src/main/java/org/apache/cxf/cdi/SingletonResourceProvider.java
rmannibucau: Well for all normal scoped beans it is still lazy thanks to CDI proxies and for @Singleton/@Dependent it becomes eager but it is ok from a business perspective + the extension already does it in several cases so not a big deal IMHO.

rmannibucau:
Anything preventing a merge or are we good to go like that? (No pression, just checking if something is expected from my side)

reta:
@rmannibucau Looks good to me, do you want me to merge?

rmannibucau:
@reta if possible, sure. Thanks a lot.

reta:
Done, thanks a lot for fixing that, @rmannibucau !

PR #419, opened 2018-05-23
[CXF-7743] wadl2java: generate throws declaration for fault response

amarkevich:


tools/wadlto/jaxrs/src/main/java/org/apache/cxf/tools/wadlto/jaxrs/SourceGenerator.java
```diff
@@ -1203,20 +1211,26 @@ private void writeJaxrResponse(StringBuilder sbCode, Set<String> imports) {
...
+    }
+
+    private static List<Element> getErrorResponses(List<Element> responseEls) {
+        return getResponses(responseEls, HTTP_ERROR_STATUSES);
+    }
+
+    private static List<Element> getResponses(List<Element> responseEls, Predicate<Integer> checker) {
+        final List<Element> result = new ArrayList<>();
+        for (Element responseEl : responseEls) {
+            for (String statusValue : responseEl.getAttribute("status").split("\\s")) {
```
coheigea: "status" is optional right? In which case it might be better to wrap this in an if statement : if (responseEl.hasAttribute("status")) {}

tools/wadlto/jaxrs/src/main/java/org/apache/cxf/tools/wadlto/jaxrs/SourceGenerator.java
```diff
@@ -1203,20 +1211,26 @@ private void writeJaxrResponse(StringBuilder sbCode, Set<String> imports) {
...
-            String statusValue = responseEls.get(i).getAttribute("status");
-            if (statusValue.length() == 0) {
-                return responseEls.get(i);
-            }
-            String[] statuses = statusValue.split("\\s");
-            for (String status : statuses) {
-                if (HTTP_OK_STATUSES.contains(status)) {
-                    return responseEls.get(i);
+    private static Element getOKResponse(List<Element> responseEls) {
+        final List<Element> result = getResponses(responseEls, HTTP_OK_STATUSES);
```
coheigea: One problem with sharing the getResponses() method is that we are creating a list for the OK case which is not really required, as we are only ever returning the first element.

tools/wadlto/jaxrs/src/main/java/org/apache/cxf/tools/wadlto/jaxrs/SourceGenerator.java
amarkevich: getAttribute returns empty string in case attribute missing; statusValue.isEmpty() -> 200

tools/wadlto/jaxrs/src/main/java/org/apache/cxf/tools/wadlto/jaxrs/SourceGenerator.java
amarkevich: previously a first OK element returned; now all OK elements collected and the same first element returned - so getOKResponse behaviour is the same

coheigea: The behaviour is the same, but we are creating a list just to store one element for the OK case.

coheigea:
LGTM thanks.

PR #423, opened 2018-06-12
CXF-7758 MTOM + SchemaValidation results in empty input stream from d…

TomasHofman:
…ata handler

https://issues.apache.org/jira/browse/CXF-7758
https://issues.jboss.org/browse/JBEAP-14830

This is to allow checking `endpoint.getEndpointInfo()` even when `exchange.getBindingOperationInfo()` is not yet set.

The `JAXBAttachmentSchemaValidationHack` is checking if the the validation is enabled in a phase before `exchange.getBindingOperationInfo()` is set. This leads to not correctly determining that validation is enabled and attachment datasources not being cached. Because the datasource is not cached, it returns the same InputStream to the validation unmarshaller and then to the application, resulting in the app not being able to read any data from the attachment DS.


TomasHofman:
Test failure seems unrelated, can we retest?

PR #434, opened 2018-08-09
NPE when annotations bundle found but not active

rgra:
bundle.getBundleContext() will give an NPE (in Equinox) if the bundle can be found but is not active.
Would also be good to at least log the exception in debug mode to point to a possible problem.

### DIFF
--- a/rt/rs/description-swagger-ui/src/main/java/org/apache/cxf/jaxrs/swagger/OsgiSwaggerUiResolver.java
+++ b/rt/rs/description-swagger-ui/src/main/java/org/apache/cxf/jaxrs/swagger/OsgiSwaggerUiResolver.java
@@ -48,8 +48,8 @@ public class OsgiSwaggerUiResolver extends SwaggerUiResolver {
                 return null;
             }
             if (bundle.getState() != Bundle.ACTIVE) {
-				bundle.start();
-			}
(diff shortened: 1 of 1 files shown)


rt/rs/description-swagger-ui/src/main/java/org/apache/cxf/jaxrs/swagger/OsgiSwaggerUiResolver.java
```diff
@@ -47,6 +47,9 @@ public String findSwaggerUiRootInternal(String swaggerUiMavenGroupAndArtifact,
             if (bundle == null) {
                 return null;
             }
+            if (bundle.getState() != Bundle.ACTIVE) {
+				bundle.start();
```
andrei-ivanov: is it me or the indentation is mixed, space and tabs?

rgra: Uh.... sorry, fixed. Was lazy and didn't edit it in the IDE but directly in GitHub and thought GitHub would be smart and tell me if indent uses the wrong characters. It doesn't.

coheigea:
Looks good thanks

PR #436, opened 2018-08-22
introduce getStreamResult

slavb18:
to allow override, e.g. for pdf generation using apache fop


slavb18:
Usage:
```java
@XSLTTransform(value = "path/to/file.xsl", mediaTypes = "application/pdf", type = XSLTTransform.TransformType.SERVER)
```

PR #437, opened 2018-08-24
[CXF-7823]:JAXBEncoderDecoder doesn't respect @XmlType's propOrder in…

jimma:
… Exception class

### DIFF
--- a/systests/databinding/src/test/java/org/apache/cxf/systest/jaxb/TestServiceTest.java
+++ b/systests/databinding/src/test/java/org/apache/cxf/systest/jaxb/TestServiceTest.java
@@ -27,10 +27,15 @@ import javax.xml.ws.BindingProvider;
 import org.w3c.dom.Document;
 
 import org.apache.cxf.Bus;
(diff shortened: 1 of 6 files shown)


systests/databinding/src/test/java/org/apache/cxf/systest/jaxb/TestServiceTest.java
```diff
@@ -73,7 +78,37 @@ public void testExtraSubClassWithJaxbFromEndpoint() throws Throwable {
...
+    }
 
+    
+    static class TestLoggingEventSender implements LogEventSender {
+        private String logMessage;
+
+        public String getMessage() {
+            return logMessage;
+        }
+        public void cleaerMessage() {
```
andrei-ivanov: *clear?

jimma: Fixed. Thanks for review @andrei-ivanov

PR #451, opened 2018-10-01
CXF-7854: Refactor RxJava2 Flowable and Observable Rx Invokers to not…

jkoehler22:
… use internal APIs

### DIFF
--- a/rt/rs/extensions/rx2/src/main/java/org/apache/cxf/jaxrs/rx2/client/FlowableRxInvokerImpl.java
+++ b/rt/rs/extensions/rx2/src/main/java/org/apache/cxf/jaxrs/rx2/client/FlowableRxInvokerImpl.java
@@ -19,6 +19,7 @@
 package org.apache.cxf.jaxrs.rx2.client;
 
 import java.util.concurrent.ExecutorService;
+import java.util.function.Supplier;
 
 import javax.ws.rs.HttpMethod;
 import javax.ws.rs.client.Entity;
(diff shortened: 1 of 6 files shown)


rt/rs/extensions/rx2/src/main/java/org/apache/cxf/jaxrs/rx2/client/FlowableRxInvokerImpl.java
```diff
@@ -148,33 +151,122 @@ public FlowableRxInvokerImpl(WebClient wc, ExecutorService ex) {
     @Override
     public <T> Flowable<T> method(String name, Entity<?> entity, Class<T> responseType) {
         if (sc == null) {
-            return Flowable.fromFuture(wc.async().method(name, entity, responseType));
+            return Flowable.create(new FlowableOnSubscribe<T>() {
```
andrei-ivanov: If these 2 objects are the same and the only difference is the `subscribeOn`+`observerOn` when `sc != null`, in my opinion it would look nicer to create the `Flowable` just once and then apply the logic on it.

andymc12: The same comment should also apply to the changes in `ObservableRxInvokerImpl`.

jkoehler22: Agree.  Thanks for the comments.  Done.

rt/rs/extensions/rx2/src/main/java/org/apache/cxf/jaxrs/rx2/client/FlowableRxInvokerImpl.java
```diff
@@ -147,34 +150,90 @@ public FlowableRxInvokerImpl(WebClient wc, ExecutorService ex) {
 
     @Override
     public <T> Flowable<T> method(String name, Entity<?> entity, Class<T> responseType) {
+        
+        Flowable<T> flowable = Flowable.create(new FlowableOnSubscribe<T>() {
+            @Override
+            public void subscribe(FlowableEmitter<T> emitter) throws Exception {
+                    try {
+                        T response = syncInvoker.method(name, entity, responseType);
+                        emitter.onNext(response);
```
reta: @jkoehler22  Do you think we should check if emitter is cancelled before calling oNext / etc?
```
if (!emitter.isCancelled()) {
       emitter.onNext(response);
}
```

rt/rs/extensions/rx2/src/main/java/org/apache/cxf/jaxrs/rx2/client/FlowableRxInvokerImpl.java
```diff
@@ -147,34 +150,90 @@ public FlowableRxInvokerImpl(WebClient wc, ExecutorService ex) {
...
-        return Flowable.fromFuture(wc.async().method(name, entity, responseType), sc);
+        return flowable.subscribeOn(sc).observeOn(sc);        
     }
 
     @Override
     public <T> Flowable<T> method(String name, Entity<?> entity, GenericType<T> responseType) {
+        
+        Flowable<T> flowable = Flowable.create(new FlowableOnSubscribe<T>() {
+            @Override
+            public void subscribe(FlowableEmitter<T> emitter) throws Exception {
```
reta: We may generalize code a bit, too many repetition I think:
```
    private <T> Flowable<T> create(Supplier<T> supplier) {
        Flowable<T> flowable = Flowable.create(new FlowableOnSubscribe<T>() {
            @Override
            public void subscribe(FlowableEmitter<T> emitter) throws Exception {
                try {
                    T response = supplier.get();
                    if (!emitter.isCancelled()) {
                        emitter.onNext(response);
                    }
                    
                    if (!emitter.isCancelled()) {
                        emitter.onComplete();
                    }
                } catch (Throwable e) {
                    if (!emitter.isCancelled()) {
                        emitter.onError(e);
                    }
                }
            }
        }, BackpressureStrategy.DROP);
        
        if (sc == null) {
            return flowable;
        }
        
        return flowable.subscribeOn(sc).observeOn(sc);
    }
```

Then the implementation is just needs a supplier:
```
   @Override
    public <T> Flowable<T> method(String name, Entity<?> entity, GenericType<T> responseType) {
        return create(() -> syncInvoker.method(name, entity, responseType));
    }

    @Override
    public <T> Flowable<T> method(String name, Class<T> responseType) {
        return create(() -> syncInvoker.method(name, responseType));
    }

    @Override
    public <T> Flowable<T> method(String name, GenericType<T> responseType) {
        return create(() -> syncInvoker.method(name, responseType));
    }
```

What do you think?

rt/rs/extensions/rx2/src/main/java/org/apache/cxf/jaxrs/rx2/client/ObservableRxInvokerImpl.java
```diff
@@ -147,34 +148,90 @@ public ObservableRxInvokerImpl(WebClient wc, ExecutorService ex) {
 
     @Override
     public <T> Observable<T> method(String name, Entity<?> entity, Class<T> responseType) {
+        
+        Observable<T> observable = Observable.create(new ObservableOnSubscribe<T>() {
```
reta: If you think previous comments for `Flowable` worth the change, we could replicate it for `Observable` as well.

rt/rs/extensions/rx2/src/main/java/org/apache/cxf/jaxrs/rx2/client/FlowableRxInvokerImpl.java
```diff
@@ -147,34 +150,90 @@ public FlowableRxInvokerImpl(WebClient wc, ExecutorService ex) {
...
+                        emitter.onComplete();
+                    } catch (Throwable e) {
+                        emitter.onError(e);
+                    }
+            }
+        }, BackpressureStrategy.DROP);
+        
         if (sc == null) {
-            return Flowable.fromFuture(wc.async().method(name, entity, responseType));
+            return flowable;
```
reta: This part is concerning. I do understand we couldn't do much if `executorService` is not provided. But it means the flow is going to be synchronous (whereas `RxInvoker` states explicitly in javadocs `... asynchronous computation ...`). May be we could default the executor to `Schedulers.io()` in this case?

reta: Just to give a reasoning behind it: the sensible default (like `Schedulers.io()` in this case) is used in many part of Java standard library (notably, `ForkJoin.commonPool()`): when the executor is not provided, the default one is used. Not the perfect choice but the asynchronous nature of the call is preserved.

rt/rs/extensions/rx2/src/main/java/org/apache/cxf/jaxrs/rx2/client/FlowableRxInvokerImpl.java
jkoehler22: Agree.  Thanks for the comments.  Done.

rt/rs/extensions/rx2/src/main/java/org/apache/cxf/jaxrs/rx2/client/FlowableRxInvokerImpl.java
jkoehler22: Agree.  Thanks for the comments.  Done.

rt/rs/extensions/rx2/src/main/java/org/apache/cxf/jaxrs/rx2/client/ObservableRxInvokerImpl.java
jkoehler22: Agree.  Thanks for the comments.  Done.

rt/rs/extensions/rx2/src/main/java/org/apache/cxf/jaxrs/rx2/client/FlowableRxInvokerImpl.java
jkoehler22: Agree.  Thanks for the comments.  Done.

reta:
LGTM, thanks a lot, @jkoehler22

reta:
@andrei-ivanov Looks good for you?

andrei-ivanov:
Yes, look fine 🙂

PR #453, opened 2018-10-02
CXF-7860: Reprocess @FormParam parms so they contain the latest data

andymc12:
This addresses an issue surfaced by the Jakarta JAX-RS API community in Issue [659](https://github.com/eclipse-ee4j/jaxrs-api/issues/659) where a parameter marked as `@FormParam` may be injected with a value that is inconsistent with an entity `Form` parameter.  The latter will have the latest contents of the HTTP form body stream whereas the former will not - modifications made to the HTTP form body during filters or interceptors will be lost, thus it is possible that the following code would fail:

```
@POST
public Response processForm(@FormParam("value") String value, Form form) {
    assertEquals(value, form.asMap().getFirst("value"));
}
```

This change ensures that the `@FormParam` parameters will match the values in the `Form` parameter.  There is probably an optimization that I am missing that would only perform this logic if the input stream were modified - suggestions welcome on how I could best do this.  Thanks!


reta:
@andymc12 Thanks a lot for addressing this very edgy case. I was looking how the current implementation works and basically I was trying to understand if we could defer the evaluation of the HTTP parameters (like `@FormParam`) to a later phase. In general, I think it would be feasible but would significantly impact the existing design.

However, I think I found a simple trick to make it work: reordering parameters initialization. Here is an interesting exercise:

- `Response processForm(@FormParam("value") String value, Form form)` -> does not work
- `Response processForm(Form form, @FormParam("value") String value)` -> works just fine

So the idea in the nutshell is to a modify `JAXRSUtils::processParameters` to initialize parameters in a different order. In this case, the `Form` should be initialized before any `@FormParam`, it triggers the reader interceptors and the right values are propagated in `Form` and `@FromParam` arguments. It should also work in case of constructors or class members.

What do you think? 
Thanks!

andymc12:
@reta Thanks so much!  That is a much better solution that what I had originally proposed.  I had also considered that the "right approach" would be to refactor the phase when parameters were processed, but I think this solution is cleaner - and should not disrupt normal cases.

I've left the test cases unaltered, but I removed my changes from `JAXRSInvoker` and instead changed `JAXRSUtils` as you suggested.  This passes all of the unit tests and the systests (including my new one).  Can you take a look at the latest changes?

Thanks again!

reta:
Looks good, thanks @andymc12 !

PR #457, opened 2018-10-10
Issue 7868: Ensure providers registered via MP Config are honored

andymc12:
Registers the providers specified via MP Config.

### DIFF
--- a/rt/rs/microprofile-client/pom.xml
+++ b/rt/rs/microprofile-client/pom.xml
@@ -87,6 +87,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.easymock</groupId>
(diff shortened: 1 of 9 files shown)


reta:
LGTM, @andymc12 would you mind please to add a test case for it? Thanks!

andymc12:
@reta thanks for the review, and sorry for the delayed response.  I added a new commit that adds a unit test for this case.  When the MP Rest Client 1.2 spec is released, it should also include a TCK test (similar to our systests) - I'll pull that in once it is available.  

This other commit also ensures that properties (passed in via `RestClientBuilder.property(...)`) are honored.

reta:
Thanks @andymc12 , looks good to me!

PR #458, opened 2018-10-11
[CXF-7871] Adding RelatesTo WS-Addressing headers in the list returned by MAPCodec#getUnderstoodHeaders()

asoldano:


asoldano:
I've restarted the Jenkins run as the two tests that failed are actually passing here locally.

PR #459, opened 2018-10-15
[CXF-6576] Handle exceptions in MessageListener container without usi…

amrMohamedFoad:
…ng setExceptionListener


deki:
The PR doesn't make sense as the changes are already in master.

PR #465, opened 2018-10-23
StringUtils: lazy Pattern evaluation

amarkevich:
https://stackoverflow.com/questions/36742258/map-getordefault-vs-getorusesupplier


amarkevich:
I will recheck this method usage; some cases can be switched to String.split fastpath way

PR #466, opened 2018-10-24
[CXF-7883] fix: handle connectionRequestTimeout in AsyncHTTPConduitFactory

gyoetam:
Motivation: if connections are contended towards a slow target, it may make sense
to set connectionTimeout and connectionRequestTimeout to values much lower than
the receiveTimeout. Expected client behavior is to receive an error if a connection
does not become available within connectionRequestTimeout. Current behavior however
is that the error is only received after up to receiveTimeout has passed, when a
current request to the target has finished and the connection is released or returned
to the pool.

Fix: instead of only checking expired leases when a connection becomes available,
periodically check them in the same thread that checks idle connections as
well. This makes sure that an error is returned to the client after
a maximum of connectionRequestTimeout + selectTimeout.

As a side effect, PoolingNHttpClientConnectionManager logs a debug level message
at every selectTimeout interval - this should be suppressed by the logging framework.

### DIFF
--- a/rt/transports/http-hc/src/main/java/org/apache/cxf/transport/http/asyncclient/AsyncHTTPConduitFactory.java
+++ b/rt/transports/http-hc/src/main/java/org/apache/cxf/transport/http/asyncclient/AsyncHTTPConduitFactory.java
@@ -375,11 +375,9 @@ public class AsyncHTTPConduitFactory implements HTTPConduitFactory {
         client = httpAsyncClientBuilder.build();
         // Start the client thread
         client.start();
-        if (this.connectionTTL == 0) {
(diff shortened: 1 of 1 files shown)


gyoetam:
Looks like the Jenkins build runs out of memory during the build, it is not related to the change in the PR.

gyoetam:
@ffang Thanks for the merge! Do you have a planned release date for 3.2.8 that will include the fix?

coheigea:
@gyoetam We just released 3.2.7, so 3.2.8 will about 2 months away probably. We hope to release 3.3.0 before then though, maybe in around 5/6 weeks.

PR #473, opened 2018-11-10
Fix newRemoteAddr is never assigned

mo7ty:
* Using the wrong variable for not null validation before HttpServletRequestXForwardedFilter.newRemoteAddr assignment.

That way, the header **_X-Forwarded-For_**, passed from AbstractHTTPServlet.checkXForwardedHeaders() was never considered.


ffang:
should be random failure

deki:
Good catch. Could you please squash your commits? I'll merge it also to 3.2.x and 3.1.x branch.

PR #475, opened 2018-11-15
Make org..client.cache.Entry serializable (cache by value), handle 30…

nbuso:
Hi I did a simple modification to permit to use cache by value and handle 304 Http status responses. Please have a look and let me know if you need further modifications/explanations

### DIFF
--- a/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/cache/CacheControlClientRequestFilter.java
+++ b/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/cache/CacheControlClientRequestFilter.java
@@ -32,6 +32,8 @@ import javax.ws.rs.client.ClientRequestFilter;
 import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.Response;
 
+import org.apache.cxf.common.util.StringUtils;
+
 @Priority(Priorities.USER - 1)
 public class CacheControlClientRequestFilter implements ClientRequestFilter {
     static final String NO_CACHE_PROPERTY = "no_client_cache";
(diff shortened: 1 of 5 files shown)


rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/cache/Entry.java
```diff
@@ -64,12 +66,19 @@ public void setData(final Serializable data) {
         this.data = data;
     }
 
-    public MultivaluedMap<String, String> getHeaders() {
+    public Map<String, List<String>> getHeaders() {
```
coheigea: This an API change, perhaps best to convert back into a MultivaluedMap for backwards compatiblity

rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/cache/Entry.java
```diff
@@ -64,12 +66,19 @@ public void setData(final Serializable data) {
...
         return headers;
     }
 
+    private void initHeaders(final  MultivaluedMap<String, String> mHeaders) {
+        this.headers = new HashMap<>();
+        mHeaders.keySet().forEach(key -> {
```
coheigea: This can be replaced with: headers.putAll(mHeaders);

rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/cache/Entry.java
nbuso: I will do, in any case that API is only used in org.apache.cxf.jaxrs.client.cache.CacheControlClientRequestFilter where the mere Map API is really used

nbuso:
Hi I added few modifications as by request. Regards

rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/cache/CacheControlClientRequestFilter.java
```diff
@@ -64,7 +66,15 @@ public void filter(final ClientRequestContext request) throws IOException {
...
             //      which may be contained in the current request
             if (entry.isOutDated()) {
-                cache.remove(key, entry);
+                String ifNoneMatchHeader = entry.getCacheHeaders().get(HttpHeaders.IF_NONE_MATCH);
+                String ifModifiedSinceHeader = entry.getCacheHeaders().get(HttpHeaders.IF_MODIFIED_SINCE);
+                
+                if (StringUtils.isEmpty(ifNoneMatchHeader) && StringUtils.isEmpty(ifModifiedSinceHeader)) {
+                    cache.remove(key, entry);
+                } else {
+                    request.getHeaders().add(HttpHeaders.IF_NONE_MATCH, ifNoneMatchHeader);
```
coheigea: I think you're also missing this:
request.getHeaders().add(HttpHeaders.IF_MODIFIED_SINCE, ifModifiedSinceHeader);

PR #477, opened 2018-11-26
CXF-7908 - Allow to customise JAXRS server in spring boot when using …

mswiderski:
…cxf-spring-boot-autoconfigure

### DIFF
--- a/integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/CxfAutoConfiguration.java
+++ b/integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/CxfAutoConfiguration.java
@@ -21,6 +21,7 @@ package org.apache.cxf.spring.boot.autoconfigure;
 import java.util.Map;
 
 import org.apache.cxf.bus.spring.SpringBus;
+import org.apache.cxf.endpoint.Server;
 import org.apache.cxf.jaxrs.JAXRSServerFactoryBean;
 import org.apache.cxf.jaxrs.spring.SpringComponentScanServer;
 import org.apache.cxf.jaxrs.spring.SpringJaxrsClassesScanServer;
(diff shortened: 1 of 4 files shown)


mswiderski:
I have checked the Jenkins reports and they seem to be caused by some download of Karaf. Is this a known issue or could it be in anyway connected to the changes of the PR?

deki:
Don't worry that's a known issue.

reta:
Thanks for the PR, @mswiderski, may be we could apply a bit simple / straightforward solution by adding `@ConditionalOnMissingBean` to `jaxRsServer` itself? F.e.

```
public class SpringComponentScanServer extends AbstractSpringComponentScanServer {
    @Bean
    @ConditionalOnMissingBean
    public Server jaxRsServer() {
        return super.createJaxRsServer();
    }
}
```

And same for `SpringJaxrsClassesScanServer`? What do you think? 
Thanks.

mswiderski:
@reta I was thinking about that at first but the issue I see is that the ConditionalOnMissingBean annotation is a spring boot specific annotation and cxf-rt-frontend-jaxrs module does not (and probably should not) depend on spring boot stuff...

reta:
@mswiderski you are absolutely right, I missed that (the `@Bean` annotation mislead me). Than your approach is certainly in the right direction. What concerns me is the presence of should-be-well-known bean name, like `jaxRsServerConfiguration`. I believe that `@ConditionalOnMissingBean(Server.class)` -> `@ConditionalOnMissingBean(name = "jaxRsServerConfiguration")` should have the same effect, what do you think?

mswiderski:
@reta  updated PR to use Server.class instead of arbitrary name, thanks

reta:
Thanks @mswiderski !

mswiderski:
@reta excellent, thanks a lot. Any information when (and in what version) this will be available as released version? Especially looking for info about cxd compatible with spring boot 2 as we (jbpm) plan to upgrade.

reta:
I am going to backport this change to `3.2.x` maintenance branch (but it would stay on Spring Boot 1.5.x), for 3.3.0 / Spring Boot 2, we have only one hard upstream dependency (Apache Karaf 4.2.2 release) to unblock us. Unfortunately we cannot commit to any date, but we are getting there. Hope it helps, thank you.

mswiderski:
@reta thanks for info, we look at the support for it around beginning of 2019 so hopefully we can get it aligned with 3.3.0 as soon as it comes out

PR #479, opened 2018-12-04
String.split via fastpath instead of precompiled Pattern

amarkevich:
https://stackoverflow.com/questions/29393086/pattern-split-slower-than-string-split

benchmark:

`
public class SplitTest {

    private static final int CNT = 131072;
    private static String[] STR = new String[CNT];
    private static final String SEPARATOR = " ";
    static {
        for (int i = 0; i < CNT; ++i) {
            STR[i] = i + SEPARATOR + i;
        }
    }

    @org.junit.Test
    public void stringSplit() {
        metrics("String.split", s -> s.split(SEPARATOR));
    }

    @org.junit.Test
    public void pattern() {
        metrics("Pattern", java.util.regex.Pattern.compile(SEPARATOR)::split);
    }

    private static void metrics(String label, Function<String, String[]> f) {
        Runtime.getRuntime().gc();
        long m = Runtime.getRuntime().freeMemory();
        long t = System.currentTimeMillis();
        for (int i = 0; i < CNT; ++i) {
            org.junit.Assert.assertSame(2, f.apply(STR[i]).length);
        }
        System.out.println(label + ": "
                + ((m - Runtime.getRuntime().freeMemory()) / 1000000) + " MB "
                + (System.currentTimeMillis() - t) + " ms");
    }

    public static void main(String[] args) {
        SplitTest splitTest = new SplitTest();
        for (int i = 0; i < 10; ++i) {
            splitTest.stringSplit();
            splitTest.pattern();
        }
    }
}
`

results using java version "1.8.0_192":
`
String.split: 33 MB 16 ms
Pattern: 57 MB 47 ms
`

### DIFF
--- a/core/src/main/java/org/apache/cxf/common/util/StringUtils.java
+++ b/core/src/main/java/org/apache/cxf/common/util/StringUtils.java
@@ -26,22 +26,16 @@ import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Predicate;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
 public final class StringUtils {
-    public static final Map<String, Pattern> PATTERN_MAP = new HashMap<>();
-    static {
-        String[] patterns = {"/", " ", ":", ",", ";", "=", "\\.", "\\+"};
-        for (String p : patterns) {
-            PATTERN_MAP.put(p, Pattern.compile(p));
-        }
-    }
+    private static final Map<String, Pattern> PATTERN_MAP = new ConcurrentHashMap<>();
     private static final Predicate<String> NOT_EMPTY = (String s) -> !s.isEmpty();
 
     private StringUtils() {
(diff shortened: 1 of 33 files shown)


andymc12:
Changes look good - thanks again!

core/src/main/java/org/apache/cxf/common/util/StringUtils.java
```diff
@@ -26,37 +26,33 @@
...
     }
 
+    /* String.split via fastpath preferred */
     public static String[] split(String s, String regex) {
         return split(s, regex, 0);
     }
+    @Deprecated
     public static String[] split(String s, String regex, int limit) {
-        Pattern p = PATTERN_MAP.getOrDefault(regex, Pattern.compile(regex));
+        Pattern p = PATTERN_MAP.computeIfAbsent(regex, key -> Pattern.compile(key));
```
reta: There is an well-known issue with JDK8 related to `computeIfAbsent`, https://bugs.openjdk.java.net/browse/JDK-8161372, not sure if we should take into account or not ...

coheigea:
Hi @amarkevich,

The following methods are not used in source (main) code any more and could be removed:

StringUtils.getFirstNotEmpty
StringUtils.splitAsStream

StringUtils.getParts could also be removed, it is called here:

AbstractHttpDestination: StringUtils.getParts(credentials, " ");
PackageUtils: StringUtils.getParts(getPackageName(cls), "\\.")

For both these cases you could replace with String.split + convert the non-empty values to a List.

Finally, that leaves HttpHeadersImpl as the only place that is calling StringUtils.split. For several of the cases here it is a single character, for others not. You could move the StringUtils split functionality into this class + see if you can re-use split for the case that it's a single character, otherwise use a pre-compiled pattern separator?

amarkevich:
I'll prepare a new PR according to proposals

PR #481, opened 2018-12-07
CXF-7923 String.split via fastpath instead of precompiled Pattern

amarkevich:
continue #479

### DIFF
--- a/core/src/main/java/org/apache/cxf/common/util/PackageUtils.java
+++ b/core/src/main/java/org/apache/cxf/common/util/PackageUtils.java
@@ -21,6 +21,7 @@ package org.apache.cxf.common.util;
 
 import java.lang.reflect.Proxy;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.StringTokenizer;
 
@@ -57,7 +58,7 @@ public final class PackageUtils {
         List<String> currentParts = new ArrayList<>();
         for (Class<?> cls : classes) {
             if (!Proxy.isProxyClass(cls)) {
-                lParts.add(StringUtils.getParts(getPackageName(cls), "\\."));
+                lParts.add(Arrays.asList(getPackageName(cls).split("\\.")));
             }
         }
         for (int i = 0; i < lParts.get(0).size(); i++) {
@@ -76,7 +77,7 @@ public final class PackageUtils {
         StringBuilder sb = new StringBuilder();
         for (String part : currentParts) {
             if (sb.length() > 0) {
-                sb.append(".");
+                sb.append('.');
             }
             sb.append(part);
         }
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/cxf/common/util/PackageUtils.java
```diff
@@ -57,7 +58,7 @@ public static String getSharedPackageName(List<Class<?>> classes) {
         List<String> currentParts = new ArrayList<>();
         for (Class<?> cls : classes) {
             if (!Proxy.isProxyClass(cls)) {
-                lParts.add(StringUtils.getParts(getPackageName(cls), "\\."));
+                lParts.add(Arrays.asList(getPackageName(cls).split("\\.")));
```
coheigea: StringUtils.getParts() only added the Strings if they were non-empty, whereas String.split will include empty Strings. I think it should be changed to preserve the old behaviour. Same goes for AbstractHTTPDestination.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/HttpHeadersImpl.java
```diff
@@ -254,10 +257,15 @@ public MediaType getMediaType() {
...
+    private static List<String> getHeaderValues(String headerName, String originalValue, String sep) {
         if (!originalValue.contains(QUOTE)
             || HEADERS_WITH_POSSIBLE_QUOTES.contains(headerName)) {
-            String[] ls = StringUtils.split(originalValue, sep);
+            final String[] ls; 
+            if (COOKIE_SEPARATOR_CRLF_EXPRESSION != sep) {
+                ls = originalValue.split(sep);
+            } else {
+                ls = COOKIE_SEPARATOR_CRLF_PATTERN.split(originalValue);
+            }
```
coheigea: I just have a minor reservation about this change. If the user specifies a custom separator that is larger than one character, then String.split will create a Pattern each time this is evaluated. Maybe instead we should move the cache of Patterns internally into this class to avoid this?

amarkevich: only CRLF accepted (HttpHeadersImplTest#testInvalidCookieSeparator)

coheigea: Oh yeah sorry, I didn't spot the length of the custom value

amarkevich:
hope https://builds.apache.org/job/CXF-Trunk-PR/857/org.apache.cxf.systests$cxf-systests-ws-rm/ is not related to current changes
@coheigea Is 'Changes requested' still actual?

coheigea:
No the Jenkins build is just the flaky ws-rm tests, please go ahead and merge

PR #482, opened 2018-12-11
[CXF-7926] Initial MP Rest Client 1.2 impl

andymc12:
This implements MP Rest Client 1.2 Milestone 2 which includes:
* specifying the baseURI in the `@RegisterRestClient` annotation.
* `removeContext` method in `AsyncInvocationInterceptor` interface.
* new `RestClientListener` interface for intercepting client instances.
* portable connect and read timeout properties.
* invoking CDI interceptors.

This commit includes the runtime changes, some unit test changes, updates to the systests to include the latest TCK test cases, etc.

### DIFF
--- a/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/spec/ClientImpl.java
+++ b/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/spec/ClientImpl.java
@@ -40,7 +40,6 @@ import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.UriBuilder;
 
 import org.apache.cxf.common.util.ClassHelper;
-import org.apache.cxf.common.util.SystemPropertyAction;
 import org.apache.cxf.configuration.jsse.TLSClientParameters;
 import org.apache.cxf.jaxrs.client.AbstractClient;
 import org.apache.cxf.jaxrs.client.ClientConfiguration;
@@ -51,24 +50,19 @@ import org.apache.cxf.jaxrs.model.FilterProviderInfo;
 import org.apache.cxf.message.Message;
 import org.apache.cxf.transport.https.SSLUtils;
 
+import static org.apache.cxf.jaxrs.client.ClientProperties.DEFAULT_THREAD_SAFETY_CLIENT_STATUS;
+import static org.apache.cxf.jaxrs.client.ClientProperties.HTTP_AUTOREDIRECT_PROP;
+import static org.apache.cxf.jaxrs.client.ClientProperties.HTTP_CONNECTION_TIMEOUT_PROP;
+import static org.apache.cxf.jaxrs.client.ClientProperties.HTTP_MAINTAIN_SESSION_PROP;
+import static org.apache.cxf.jaxrs.client.ClientProperties.HTTP_PROXY_SERVER_PORT_PROP;
+import static org.apache.cxf.jaxrs.client.ClientProperties.HTTP_PROXY_SERVER_PROP;
+import static org.apache.cxf.jaxrs.client.ClientProperties.HTTP_RECEIVE_TIMEOUT_PROP;
+import static org.apache.cxf.jaxrs.client.ClientProperties.HTTP_RESPONSE_AUTOCLOSE_PROP;
+import static org.apache.cxf.jaxrs.client.ClientProperties.THREAD_SAFE_CLIENT_PROP;
+import static org.apache.cxf.jaxrs.client.ClientProperties.THREAD_SAFE_CLIENT_STATE_CLEANUP_PERIOD;
+import static org.apache.cxf.jaxrs.client.ClientProperties.THREAD_SAFE_CLIENT_STATE_CLEANUP_PROP;
+
 public class ClientImpl implements Client {
-    public static final String HTTP_CONNECTION_TIMEOUT_PROP = "http.connection.timeout";
-    public static final String HTTP_RECEIVE_TIMEOUT_PROP = "http.receive.timeout";
-    private static final String HTTP_PROXY_SERVER_PROP = "http.proxy.server.uri";
-    private static final String HTTP_PROXY_SERVER_PORT_PROP = "http.proxy.server.port";
-    private static final String HTTP_AUTOREDIRECT_PROP = "http.autoredirect";
-    private static final String HTTP_MAINTAIN_SESSION_PROP = "http.maintain.session";
-    private static final String HTTP_RESPONSE_AUTOCLOSE_PROP = "http.response.stream.auto.close";
-    private static final String THREAD_SAFE_CLIENT_PROP = "thread.safe.client";
-    private static final String THREAD_SAFE_CLIENT_STATE_CLEANUP_PROP = "thread.safe.client.state.cleanup.period";
-    private static final Boolean DEFAULT_THREAD_SAFETY_CLIENT_STATUS;
-    private static final Integer THREAD_SAFE_CLIENT_STATE_CLEANUP_PERIOD;
-    static  {
-        DEFAULT_THREAD_SAFETY_CLIENT_STATUS =
-            Boolean.parseBoolean(SystemPropertyAction.getPropertyOrNull(THREAD_SAFE_CLIENT_PROP));
-        THREAD_SAFE_CLIENT_STATE_CLEANUP_PERIOD =
-            getIntValue(SystemPropertyAction.getPropertyOrNull(THREAD_SAFE_CLIENT_STATE_CLEANUP_PROP));
-    }
 
     private Configurable<Client> configImpl;
     private TLSConfiguration secConfig;
--- a/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/spec/ClientImpl.java
+++ b/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/spec/ClientImpl.java
@@ -52,8 +52,8 @@ import org.apache.cxf.message.Message;
 import org.apache.cxf.transport.https.SSLUtils;
 
 public class ClientImpl implements Client {
-    static final String HTTP_CONNECTION_TIMEOUT_PROP = "http.connection.timeout";
-    static final String HTTP_RECEIVE_TIMEOUT_PROP = "http.receive.timeout";
+    public static final String HTTP_CONNECTION_TIMEOUT_PROP = "http.connection.timeout";
+    public static final String HTTP_RECEIVE_TIMEOUT_PROP = "http.receive.timeout";
     private static final String HTTP_PROXY_SERVER_PROP = "http.proxy.server.uri";
     private static final String HTTP_PROXY_SERVER_PORT_PROP = "http.proxy.server.port";
     private static final String HTTP_AUTOREDIRECT_PROP = "http.autoredirect";
--- a/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/CxfTypeSafeClientBuilder.java
+++ b/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/CxfTypeSafeClientBuilder.java
@@ -20,21 +20,55 @@ package org.apache.cxf.microprofile.client;
 
 import java.net.URI;
 import java.net.URL;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
+import java.util.ServiceLoader;
+import java.util.WeakHashMap;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import javax.ws.rs.core.Configurable;
 import javax.ws.rs.core.Configuration;
 
+import org.apache.cxf.common.logging.LogUtils;
+import org.apache.cxf.jaxrs.client.spec.ClientImpl;
+import org.apache.cxf.microprofile.client.config.ConfigFacade;
 import org.eclipse.microprofile.rest.client.RestClientBuilder;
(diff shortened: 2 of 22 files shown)


rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/CxfTypeSafeClientBuilder.java
```diff
@@ -20,22 +20,56 @@
...
+    private static final Map<ClassLoader, Collection<RestClientListener>> REST_CLIENT_LISTENERS = 
+        new WeakHashMap<>();
+
     private String baseUri;
     private ExecutorService executorService;
     private final MicroProfileClientConfigurableImpl<RestClientBuilder> configImpl =
             new MicroProfileClientConfigurableImpl<>(this);
 
+    private static Collection<RestClientListener> listeners() {
+        ClassLoader threadContextClassLoader = AccessController.doPrivileged(
```
reta: Should we check if `SecurityManager` is present? 
```
if (System.getSecurityManager() == null) {
  // non priviledged 
} else {
  // priviledged 
}
```
Since most of the apps are run without `SecurityManager`, we could gain some efficiency here. What do you think?

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/CxfTypeSafeClientBuilder.java
```diff
@@ -20,22 +20,56 @@
...
+            (PrivilegedAction<ClassLoader>)() -> Thread.currentThread().getContextClassLoader());
+        synchronized (REST_CLIENT_LISTENERS) {
+            return REST_CLIENT_LISTENERS.computeIfAbsent(threadContextClassLoader, key -> {
```
reta: May be just with steams:
```
return REST_CLIENT_LISTENERS.computeIfAbsent(threadContextClassLoader, key -> StreamSupport
    .stream(ServiceLoader.load(RestClientListener.class).spliterator(), false)
    .collect(Collectors.toList())
```

rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/spec/ClientImpl.java
```diff
@@ -52,8 +52,8 @@
 import org.apache.cxf.transport.https.SSLUtils;
 
 public class ClientImpl implements Client {
-    static final String HTTP_CONNECTION_TIMEOUT_PROP = "http.connection.timeout";
-    static final String HTTP_RECEIVE_TIMEOUT_PROP = "http.receive.timeout";
+    public static final String HTTP_CONNECTION_TIMEOUT_PROP = "http.connection.timeout";
```
reta: May be we could extract these props to `ClientProperties` since they are used in quite a few places?

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/CxfTypeSafeClientBuilder.java
```diff
@@ -76,8 +132,32 @@ public RestClientBuilder executorService(ExecutorService executor) {
...
+
+        final String interfaceName = aClass.getName();
+        Optional<Long> timeout = ConfigFacade.getOptionalValue(
+            String.format(REST_CONN_TIMEOUT_FORMAT, interfaceName), Long.class).map(
+                timeoutValue -> {
+                    connectTimeout(timeoutValue, TimeUnit.MILLISECONDS);
+                    return timeoutValue;
+                });
+        if (!timeout.isPresent() && LOG.isLoggable(Level.FINEST)) {
+            LOG.finest("connectionTimeout set by MP Config: " + timeout.get());
```
reta: Uh ... `timeout.get()` would blow up since `if` statement has `!timeout.isPresent()` ... Did you mean
```
if (LOG.isLoggable(Level.FINEST)) {
            timeout.ifPresent(t -> 
                LOG.finest("connectionTimeout set by MP Config: " + t)
            );
        };
```?

File: rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/CxfTypeSafeClientBuilder.java [hunk 5]
```diff
@@ -76,8 +132,32 @@ public RestClientBuilder executorService(ExecutorService executor) {
...
+        }
+        timeout = ConfigFacade.getOptionalValue(
+            String.format(REST_READ_TIMEOUT_FORMAT, interfaceName), Long.class).map(
+                timeoutValue -> {
+                    readTimeout(timeoutValue, TimeUnit.MILLISECONDS);
+                    return timeoutValue;
+                });
+        if (!timeout.isPresent() && LOG.isLoggable(Level.FINEST)) {
```
reta: Same as above :)

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/CxfTypeSafeClientBuilder.java
```diff
@@ -76,8 +132,32 @@ public RestClientBuilder executorService(ExecutorService executor) {
...
+            LOG.finest("readTimeout set by MP Config: " + timeout.get());
+        }
+
+        for (RestClientListener listener : listeners()) {
```
reta: Minor, `listeners().forEach(l -> l.onNewClient(aClass, this));`

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/CxfTypeSafeClientBuilder.java
```diff
@@ -76,8 +132,32 @@ public RestClientBuilder executorService(ExecutorService executor) {
                 }
             }
         }
+
+        final String interfaceName = aClass.getName();
+        Optional<Long> timeout = ConfigFacade.getOptionalValue(
```
reta: `OptionalLong` please, just for `Long` :)

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/CxfTypeSafeClientBuilder.java
```diff
@@ -76,8 +132,32 @@ public RestClientBuilder executorService(ExecutorService executor) {
...
+            String.format(REST_CONN_TIMEOUT_FORMAT, interfaceName), Long.class).map(
+                timeoutValue -> {
+                    connectTimeout(timeoutValue, TimeUnit.MILLISECONDS);
+                    return timeoutValue;
+                });
+        if (!timeout.isPresent() && LOG.isLoggable(Level.FINEST)) {
+            LOG.finest("connectionTimeout set by MP Config: " + timeout.get());
+        }
+        timeout = ConfigFacade.getOptionalValue(
```
reta: May be use `connectTimeout` and `readTimeout` instead of reusing the `timeout`?

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientFactoryBean.java
```diff
@@ -97,13 +98,16 @@ protected void initClient(AbstractClient client, Endpoint ep, boolean addHeaders
     @Override
     protected ClientProxyImpl createClientProxy(ClassResourceInfo cri, boolean isRoot,
                                                 ClientState actualState, Object[] varValues) {
+        CDIInterceptorWrapper interceptorWrapper = CDIInterceptorWrapper.createWrapper(getServiceClass());
+        MicroProfileClientProxyImpl impl;
         if (actualState == null) {
-            return new MicroProfileClientProxyImpl(URI.create(getAddress()), proxyLoader, cri, isRoot,
-                    inheritHeaders, executorService, configuration, varValues);
+            impl = new MicroProfileClientProxyImpl(URI.create(getAddress()), proxyLoader, cri, isRoot,
```
reta: I think the variant with `return` statements was more readable and straightforward, just an opinion

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/CDIInterceptorWrapper.java
```diff
@@ -0,0 +1,70 @@
...
+            return AccessController.doPrivileged((PrivilegedExceptionAction<CDIInterceptorWrapper>) () -> {
+                Class<?> cdiClass = Class.forName("javax.enterprise.inject.spi.CDI", false,
+                                                  restClient.getClassLoader());
+                Method currentMethod = cdiClass.getMethod("current");
+                Object cdiCurrent = currentMethod.invoke(null);
+
+                Method getBeanMgrMethod = cdiClass.getMethod("getBeanManager");
+                getBeanMgrMethod.invoke(cdiCurrent);
+
+                return new CDIInterceptorWrapperImpl(restClient);
```
reta: May be we could pass the `BeanManager` from `getBeanMgrMethod.invoke(cdiCurrent)` to `CDIInterceptorWrapperImpl` and not invoke `CDI.current()` in there?

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/RestClientBean.java
```diff
@@ -141,18 +142,30 @@ private String getBaseUri() {
         String property = String.format(REST_URI_FORMAT, interfaceName);
         String baseURI = null;
```
reta: A bit simpler
```
String baseURI = ConfigFacade
        .getOptionalValue(String.format(REST_URI_FORMAT, interfaceName), String.class)
        .orElseGet(() -> ConfigFacade.getOptionalValue(String.format(REST_URL_FORMAT, interfaceName), String.class).orElse(null));
```

rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/spec/ClientImpl.java
andymc12: Does `ClientProperties` already exist, or are you suggesting I create a new class for it?

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/CxfTypeSafeClientBuilder.java
andymc12: Yes, exactly - good catch!

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientFactoryBean.java
andymc12: Sure, I can switch it back - I think I changed it for some debugging and just left it, but it's fine with me either way.

andymc12:
@reta - Thanks so much for the thorough review!  I've updated the PR with a new commit (Code review comments) that addresses everything you mentioned except for putting the timeout properties in `ClientProperties` - do we need to create a new class/interface for that?

reta:
hey @andymc12 , sorry was crazy day at work, yeah I think we could create dedicated class/interface, like `ClientProperties` and move these common constants here. What do you think? Thanks a lot!

andymc12:
@reta I introduced a new `ClientProperties` interface with all of the properties from `ClientImpl` - afaict, only the timeout-based properties are widely used.  Can you take another look?

Thanks again!

reta:
Thanks @andymc12, LGTM! Would be great if @rmannibucau could take a look on CDI part, thanks!

andymc12:
@rmannibucau sorry for the slow response.  Do you have (or can you provide) any examples of both 1 and 2?  My CDI knowledge is pretty limited, and I'd like to make sure this works for CDI extensions - and avoid unnecessary code if possible.  Thanks!

rmannibucau:
Hi @andymc12 , happy new year :)

about 1 maybe check out https://github.com/apache/geronimo-metrics/blob/master/geronimo-metrics/src/main/java/org/apache/geronimo/microprofile/metrics/cdi/CountedInterceptor.java#L91, it is not the best impl - which would use an extension to capture it and inject it in a specific interceptor - but a good compromise regarding the complexity. Then you have the annotated model from cdi which replaces all java reflect usages.

About 2: [1] is the one, it enables to manage the interceptors (configured) for you

Let me know - or ping me on slack/irc/gmail - if you need more pointers/help.

[1] https://docs.jboss.org/cdi/api/2.0/javax/enterprise/inject/spi/InterceptionFactory.html

andymc12:
Happy New Year @rmannibucau! And thanks - that helps a lot!  

I've updated the code to use CDI AnnotatedTypes which should address your first concern.

As for the second concern, I agree that the InterceptionFactory should help, but I think that it will not work (at least for wrapping the MP Rest Client instance) because it would be creating a dynamic proxy to wrap another dynamic proxy.  I'm not sure if that is possible - and I think that it falls under the [unproxyable category](http://docs.jboss.org/cdi/spec/2.0-PRD/cdi-spec.html#unproxyable).

rmannibucau:
Well there is a ticket for that at cdi level (isnt it stupid technically? ;)). But the fix can be: provide the "internal most nested impl" as a specific bean and inject this "impl bean" to the factory :). You can even do the impl as an interceptor (no more invocation handler at all) for that case with no impl and just use the factory with a bean returning null as base.

andymc12:
Per discussion in Google Chat, this (suggestion 2) fails in OWB but can work in Weld.  For now, I'll plan to merge this as-is - then (after OWB is fixed) we can make these suggested changes.  Hopefully by that time, we'll also have an official release of MP Rest Client 1.2 for the APIs/TCKs.

rmannibucau:
Just for tracking purpose: OWB fix is on the snapshot so 2.0.10 will get it and the interceptor stack handling impl will be droppable

PR #483, opened 2018-12-14
HttpHeaderHelper: String.replace instead of Pattern.matcher().replaceAll

amarkevich:
benchmark [1] results:
- old: 40 MB 47 ms
- new: 9 MB 0 ms
1.
```
public class HttpHeaderHelperMapCharsetTest {
    private static final int CNT = 131072;
    static final String charset = java.nio.charset.StandardCharsets.UTF_8.name();
    private static final String[] STR = new String[CNT];
    static {
        for (int i = 0; i < CNT; ++i) {
            if (i % 3 == 0) {
                STR[i] = "\"" + charset + "\"";
            } else if (i % 3 == 1) {
                STR[i] = "'" + charset + "'";
            } else {
                STR[i] = charset;
            }
        }
    }

    @org.junit.Test
    public void mapCharset() {
        metrics("HttpHeaderHelper::mapCharset", HttpHeaderHelper::mapCharset);
    }

    @org.junit.Test
    public void mapCharset2() {
        metrics("HttpHeaderHelper::mapCharset2", HttpHeaderHelper::mapCharset2);
    }

    private static void metrics(String label, java.util.function.Function<String, String> f) {
        Runtime.getRuntime().gc();
        long m = Runtime.getRuntime().freeMemory();
        long t = System.currentTimeMillis();
        for (int i = 0; i < CNT; ++i) {
            org.junit.Assert.assertEquals(charset, f.apply(STR[i]));
        }
        System.out.println(label + ": " + ((m - Runtime.getRuntime().freeMemory()) / (1024 * 1024)) + " MB "
                + (System.currentTimeMillis() - t) + " ms");
    }

    public static void main(String[] args) {
        final HttpHeaderHelperMapCharsetTest splitTest = new HttpHeaderHelperMapCharsetTest();
        for (int i = 0; i < 10; ++i) {
            System.out.println("Round #" + i);
            splitTest.mapCharset();
            splitTest.mapCharset2();
        }
    }

}
```

### DIFF
--- a/core/src/main/java/org/apache/cxf/helpers/HttpHeaderHelper.java
+++ b/core/src/main/java/org/apache/cxf/helpers/HttpHeaderHelper.java
@@ -21,12 +21,12 @@ package org.apache.cxf.helpers;
 
 import java.nio.charset.Charset;
 import java.nio.charset.IllegalCharsetNameException;
+import java.nio.charset.StandardCharsets;
 import java.nio.charset.UnsupportedCharsetException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.regex.Pattern;
 
 public final class HttpHeaderHelper {
     public static final String ACCEPT_ENCODING = "Accept-Encoding";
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/cxf/helpers/HttpHeaderHelper.java
```diff
@@ -112,8 +111,8 @@ public static String mapCharset(String enc, String deflt) {
         }
         // Charsets can be quoted. But it's quite certain that they can't have escaped quoted or
         // anything like that.
-        enc = charsetPattern.matcher(enc).replaceAll("").trim();
-        if ("".equals(enc)) {
+        enc = enc.replace('"', ' ').replace('\'', ' ').trim();
```
coheigea: Is there a reason to replace the previous behaviour (empty String) with a whitespace? Also, it's only replacing the first instance of each? I'm just wondering if there might be an attack where an attacker could force using a charset by escaping twice in a row with the new code.

amarkevich: char replacement performed without Pattern usage. its required only in case of quotes and removed by existing trim. result value used as argument for Character.forName which is safe and throws exception on any wrong input

coheigea: LGTM thanks

PR #484, opened 2018-12-18
Update maven-bundle-plugin

amarkevich:
Solution for [1]

1. https://stackoverflow.com/questions/31231453/when-using-bundle-packaging-with-maven-bundle-plugin-goals-are-executed-twice


deki:
Great someone tried it earlier but this update caused build failures: CXF-7766

amarkevich:
yep; fixed by 2nd commit

coheigea:
LGTM

PR #487, opened 2018-12-21
Ensure JAXB and javax.activation are not required for JAX-RS

rmannibucau:
Goal is to ensure default java11 experience is still good when not needing JAXB/activation (which is pretty much often these days with JAX-RS)

### DIFF
--- a/core/src/main/java/org/apache/cxf/service/model/EndpointInfo.java
+++ b/core/src/main/java/org/apache/cxf/service/model/EndpointInfo.java
@@ -21,6 +21,7 @@ package org.apache.cxf.service.model;
 
 import javax.xml.namespace.QName;
 
+import org.apache.cxf.ws.addressing.AttributedURIType;
 import org.apache.cxf.ws.addressing.EndpointReferenceType;
 import org.apache.cxf.ws.addressing.EndpointReferenceUtils;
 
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/apache/cxf/service/model/EndpointInfo.java
```diff
@@ -94,7 +95,12 @@ public String getAddress() {
 
     public void setAddress(String addr) {
         if (null == address) {
-            address = EndpointReferenceUtils.getEndpointReference(addr);
+            // address = EndpointReferenceUtils.getEndpointReference(addr); loads jaxb and we want to avoid it
+            final EndpointReferenceType reference = new EndpointReferenceType();
+            final AttributedURIType a = new AttributedURIType();
+            a.setValue("/");
```
coheigea: This should be: a.setValue(address);

rmannibucau: Hmm, sure? address is null here so it defaults to "/" but good catch i forgot the else ;)

coheigea:
ProviderFactory has a method:

 static int comparePriorityStatus(Class<?> cl1, Class<?> cl2) {
        Integer value1 = AnnotationUtils.getBindingPriority(cl1);
        Integer value2 = AnnotationUtils.getBindingPriority(cl2);
        return value1.compareTo(value2);
    }

AnnotationUtils.has: import javax.annotation.Priority;

So I guess AnnotationUtils.getBindingPriority needs to use some reflection as well here.

rmannibucau:
@coheigea have to admit i used CDI to test so commons-annotation was there (and @Priority) but good catch, updated

rmannibucau:
@coheigea fixed it, misread the getClassAnnotation method (thought the first test was on the annotation but it was on the class). Thanks for catching it.

coheigea:
Thanks @rmannibucau

PR #490, opened 2018-12-26
cxf-rt-transports-udp: minor code improvements to improve test stability

amarkevich:


rt/transports/udp/src/main/java/org/apache/cxf/transport/udp/UDPConduit.java
```diff
@@ -150,11 +150,7 @@ public void close(Message msg) throws IOException {
             if (s != null && c != null) {
                 c.getSession().removeAttribute(CXF_MESSAGE_ATTR);
 
-                Queue<ConnectFuture> q = connections.get(s);
```
reta: The change looks logical but I would recommend to revert it please. There is an issue with `computeIfAbsent` (https://bugs.openjdk.java.net/browse/JDK-8161372) which leads to unnecessary locks and contention on JDKs below 9. Since this is pretty low-level component, under certain conditions it could kill the throughput, I have seen that in the wild.

PR #493, opened 2019-01-04
cxf-systests-codegen: update dependency after CXF-7924

amarkevich:


coheigea:
+1 - I had noticed the test failures on Jenkins.

PR #500, opened 2019-01-14
[CXF-7910] Change JAX-WS javax to jakarta artifact dependencies

deki:


deki:
It's just the flaky test failing, so I think this is ready for merge...

parent/pom.xml
```diff
@@ -2854,35 +2878,38 @@
...
+                <dependency>
+                    <groupId>jakarta.xml.soap</groupId>
+                    <artifactId>jakarta.xml.soap-api</artifactId>
+                    <version>1.4.1</version>
                 </dependency>
                 <dependency>
                     <groupId>com.sun.xml.messaging.saaj</groupId>
                     <artifactId>saaj-impl</artifactId>
-                    <version>1.4.0-b03</version>
+                    <version>1.3.23</version>
```
reta: The SAAJ (impl) was one of the problems we run into before, so we have now:
 - `1.5.1` in samples/pom.xml
 - `1.3.23` in parent/pom.xml

Meantime, SAAJ API is 1.4.0.1:
 - `<cxf.osgi.saaj.version>[1.4,2)</cxf.osgi.saaj.version>` in parent/pom.xml

Should we keep `1.4.0-b03` to align API and impl? It seemed to work fine.

distribution/src/main/release/samples/pom.xml
```diff
@@ -261,24 +261,24 @@
                     <version>1.3.1</version>
                 </dependency>
                 <dependency>
-                    <groupId>javax.xml.ws</groupId>
-                    <artifactId>jaxws-api</artifactId>
-                    <version>2.3.0</version>
+                    <groupId>jakarta.xml.ws</groupId>
```
reta: This guy brings `jakarta.annotation-api`, should we get rid of `javax.annotation-api` in this case?

parent/pom.xml
```diff
@@ -2854,35 +2878,38 @@
...
-                    <groupId>javax.xml.ws</groupId>
-                    <artifactId>jaxws-api</artifactId>
-                    <version>2.3.1</version>
-                    <exclusions>
-                        <exclusion>
-                            <groupId>javax.xml.bind</groupId>
-                            <artifactId>jaxb-api</artifactId>
-                        </exclusion>
-                    </exclusions>
+                    <groupId>jakarta.xml.ws</groupId>
```
reta: This guy brings jakarta.annotation-api, should we get rid of javax.annotation-api in this case?

distribution/src/main/release/samples/pom.xml
```diff
@@ -261,24 +261,24 @@
...
+                    <artifactId>jakarta.xml.ws-api</artifactId>
+                    <version>2.3.2</version>
                 </dependency>
                 <dependency>
```
reta: I think we could remove `javax.activation-api`, `jakarta.activation-api` replaces it.

parent/pom.xml
deki: Good catch. For me 1.5.1 also works, but 1.4 is fine as well. @ffang what issues did you have?

distribution/src/main/release/samples/pom.xml
deki: Definitely, my plan was to do it along with Spring upgrade afterwards. But we can also do it now, it's hard to separate these things.

parent/pom.xml
deki: Just checked 1.5.2-b01, also works.

ffang: @deki , sorry, can't recall exactly, but should be OSGi test related
I suggest we use saaj api 1.4 from SMX and use saaj 1.3 impl

Thanks!

PR #502, opened 2019-01-22
SSLUtils: move tests and update due to JDK-8211883

amarkevich:
https://bugs.openjdk.java.net/browse/JDK-8211883 Disable anon and NULL cipher suites
Backported to 7u211, 8u201, 11.0.2


coheigea:
@amarkevich I will just remove the tests, as the server endpoint doesn't support NULL any more

PR #506, opened 2019-01-24
Fix Spring 5 OSGi version constraints

siepkes:
Commit 424591420b052a2b557d3f1dd48609c9858d9809 bumped Spring to version 5 however the OSGi version constraints still only allowed max version 4 of Spring.


reta:
Thanks, @siepkes, it make sense to bump upper limit for OSGi. Regarding Spring 3/4, CXF does no really depend on anything beside Spring core components (we also need to play well with Spring Boot 1.5.x, which is still on Spring 4.3.x).

PR #507, opened 2019-01-30
CXF-7957: Swagger2Feature Doesn't Work With Swagger Versions Above 1.…

jwcarman:
This patch compensates for the Swagger library's change in behavior.  They are now using a Swagger object as the entity in their responses and relying upon the SwaggerSerializers (a MessageBodyWriter<Swagger>) to take care of the serialization (they were previously just converting to a JSON string and using that as the entity).  The JSON serialization causes there to be null values in the JSON, which was breaking our SwaggerParseUtils method (there's an included test case for it).

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -164,7 +164,7 @@
         <cxf.spring.osgi.version>1.2.1</cxf.spring.osgi.version>
         <cxf.spring.ldap.version>2.3.2.RELEASE</cxf.spring.ldap.version>
         <cxf.spring.mock>spring-test</cxf.spring.mock>
-        <cxf.swagger2.version>1.5.21</cxf.swagger2.version><!-- updating to 1.5.18 will cause systest failures -->
+        <cxf.swagger2.version>1.5.21</cxf.swagger2.version>
         <cxf.swagger.v3.version>2.0.6</cxf.swagger.v3.version>
         <cxf.swagger.ui.version>3.20.5</cxf.swagger.ui.version>
         <cxf.tomcat.version>8.5.37</cxf.tomcat.version>
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -164,7 +164,7 @@
         <cxf.spring.osgi.version>1.2.1</cxf.spring.osgi.version>
         <cxf.spring.ldap.version>2.3.2.RELEASE</cxf.spring.ldap.version>
         <cxf.spring.mock>spring-test</cxf.spring.mock>
-        <cxf.swagger2.version>1.5.17</cxf.swagger2.version><!-- updating to 1.5.18 will cause systest failures -->
+        <cxf.swagger2.version>1.5.21</cxf.swagger2.version><!-- updating to 1.5.18 will cause systest failures -->
         <cxf.swagger.v3.version>2.0.6</cxf.swagger.v3.version>
         <cxf.swagger.ui.version>3.20.5</cxf.swagger.ui.version>
         <cxf.tomcat.version>8.5.37</cxf.tomcat.version>
--- a/rt/rs/description-swagger/src/main/java/org/apache/cxf/jaxrs/swagger/Swagger2Feature.java
+++ b/rt/rs/description-swagger/src/main/java/org/apache/cxf/jaxrs/swagger/Swagger2Feature.java
@@ -40,7 +40,6 @@ import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
 
-import io.swagger.jaxrs.listing.SwaggerSerializers;
 import org.apache.cxf.Bus;
 import org.apache.cxf.annotations.Provider;
 import org.apache.cxf.annotations.Provider.Scope;
@@ -64,6 +63,7 @@ import io.swagger.jaxrs.config.DefaultReaderConfig;
 import io.swagger.jaxrs.config.ReaderConfig;
 import io.swagger.jaxrs.config.SwaggerContextService;
 import io.swagger.jaxrs.listing.ApiListingResource;
+import io.swagger.jaxrs.listing.SwaggerSerializers;
 import io.swagger.models.Swagger;
 import io.swagger.models.auth.SecuritySchemeDefinition;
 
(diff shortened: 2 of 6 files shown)


parent/pom.xml
```diff
@@ -164,7 +164,7 @@
         <cxf.spring.osgi.version>1.2.1</cxf.spring.osgi.version>
         <cxf.spring.ldap.version>2.3.2.RELEASE</cxf.spring.ldap.version>
         <cxf.spring.mock>spring-test</cxf.spring.mock>
-        <cxf.swagger2.version>1.5.17</cxf.swagger2.version><!-- updating to 1.5.18 will cause systest failures -->
+        <cxf.swagger2.version>1.5.21</cxf.swagger2.version><!-- updating to 1.5.18 will cause systest failures -->
```
ilgrosso: I'd remove this comment

ilgrosso:
Besides a minor edit request, LGTM, thanks @jwcarman !
Let's see if there is any trouble with tests..

Once merged, I'd also backport to 3.1.x-fixes and 3.2.x-fixes

jwcarman:
I added an entity type check to make sure it's still compatible with 1.5.17 also and not just the newer stuff.

parent/pom.xml
jwcarman: Done.

ilgrosso:
@reta you're absolutely right, thx for noticing

jwcarman:
I didn't figure it'd be a big deal, since the SwaggerSerializers and BaseApiListingResource appear to be designed to be used together (especially now since the entity type is Swagger).  We are already using BaseApiListingResource.

parent/pom.xml
```diff
@@ -164,7 +164,7 @@
         <cxf.spring.osgi.version>1.2.1</cxf.spring.osgi.version>
         <cxf.spring.ldap.version>2.3.2.RELEASE</cxf.spring.ldap.version>
         <cxf.spring.mock>spring-test</cxf.spring.mock>
-        <cxf.swagger2.version>1.5.17</cxf.swagger2.version><!-- updating to 1.5.18 will cause systest failures -->
+        <cxf.swagger2.version>1.5.21</cxf.swagger2.version>
```
reta: :+1:

rt/rs/description-swagger/src/main/java/org/apache/cxf/jaxrs/swagger/Swagger2Feature.java
```diff
@@ -151,6 +152,8 @@ protected void addSwaggerResource(Server server, Bus bus) {
         swaggerResources.add(apiListingResource);
 
         List<Object> providers = new ArrayList<>();
+        providers.add(new SwaggerSerializers());
```
reta: :+1:, we could also get rid of some code in `3.1.x` I think

rt/rs/description-swagger/src/main/java/org/apache/cxf/jaxrs/swagger/openapi/SwaggerToOpenApiConversionFilter.java
```diff
@@ -55,7 +57,8 @@ public void filter(ContainerRequestContext reqCtx) throws IOException {
     @Override
     public void filter(ContainerRequestContext reqCtx, ContainerResponseContext respCtx) throws IOException {
         if (Boolean.TRUE == reqCtx.getProperty(OPEN_API_PROPERTY)) {
-            String swaggerJson = (String)respCtx.getEntity();
+            String swaggerJson = respCtx.getEntity() instanceof String ? (String)respCtx.getEntity()
```
reta: Minor, `String swaggerJson = Json.mapper().writeValueAsString(respCtx.getEntity());` should do it.

rt/rs/description-swagger/src/main/java/org/apache/cxf/jaxrs/swagger/parse/SwaggerParseUtils.java
```diff
@@ -131,41 +131,41 @@ public static UserApplication getUserApplicationFromJson(String json,
...
-                    //"query", "header", "path", "formData" or "body"
-                    String paramType = (String)param.get("in");
-                    ParameterType pType = "body".equals(paramType) ? ParameterType.REQUEST_BODY
-                        : "formData".equals(paramType)
-                        ? ParameterType.FORM : ParameterType.valueOf(paramType.toUpperCase());
-                    Parameter userParam = new Parameter(pType, name);
-
-                    setJavaType(userParam, (String)param.get("type"));
-                    userOpParams.add(userParam);
+                if (oper != null) {
```
reta: Just to understand, the `oper` is null when API has no any endpoints / operations exposed?

jwcarman: When I debugged it, the map that came back had keys for all the HTTP methods, even i they weren't supported by the endpoint.  My guess is that their previous JSON serialization suppressed properties with null values and the new one doesn't.  So, now you'll get a "head" property with a null value.

rt/rs/description-swagger/src/main/java/org/apache/cxf/jaxrs/swagger/openapi/SwaggerToOpenApiConversionFilter.java
jwcarman: I was trying to maintain the exact functionality in a 1.5.17 world if I could.

reta: It is exactly the same: the mapper does nothing for `String` (returns the same string back) but convert to JSON any object (same is done with `?:`)

jwcarman:
I’m fine with it. Was just explaining my motivation. I’ll update when I get
back home later.

rt/rs/description-swagger/src/main/java/org/apache/cxf/jaxrs/swagger/openapi/SwaggerToOpenApiConversionFilter.java
reta: @jwcarman Please disregard my comment, the behavior is not identical (I thought it was).

rt/rs/description-swagger/src/main/java/org/apache/cxf/jaxrs/swagger/parse/SwaggerParseUtils.java
reta: :+1:

reta:
Thanks @jwcarman! I am on it, just adding a few comments to the filter (will do followup commit) and backporting to 3.2.x and 3.1.x as well.

PR #510, opened 2019-02-08
Fix CXF-7966: allows more flexibility in Beanspector

mrulli:
Signed-off Matteo Rulli <matteo.rulli@flairbit.io>


rt/rs/extensions/search/src/main/java/org/apache/cxf/jaxrs/ext/search/Beanspector.java
```diff
@@ -58,31 +58,49 @@ public Beanspector(T tobj) {
...
+                } 
+            } 
+        } 
+        // check type equality for getter-setter pairs 
+        Set<String> pairs = new HashSet<>(getters.keySet()); 
+        pairs.retainAll(setters.keySet()); 
+        for (String accessor : pairs) { 
+            Class<?> getterClass = getters.get(accessor).getReturnType(); 
+            Class<?> setterClass = setters.get(accessor).getParameterTypes()[0]; 
+            if (!setterClass.isAssignableFrom(getterClass)) {
```
reta: Shouldn't it be reversed to `if (!getterClass.isAssignableFrom(setterClass)) {`? Like for example getter I think should always return the subclass (or exact type) of setter. Example here:
```
    class BaseBean {
        public Number getNumber() {
            return 0L;
        }

        public void setNumber(Number val) {
        }
    }
    
    class InheritableBean extends BaseBean {
        public void setNumber(Long val) {
        }
    }
```

The `getNumber` return `Number` which could be assigned from `Long`, however `Long` is not assignable from `Number`. What do you think?

reta:
Thanks for this PR, @mrulli, the problem is understood, I think you solution looks good.

mrulli:
Hello, should we do something to get this PR accepted and merged? Are there any problem or missing actions on our side? Thanks.

reta:
Hi @mrulli, there is one concern expressed through the comment, it would be much appreciated if you clear it out, thank you.

rt/rs/extensions/search/src/main/java/org/apache/cxf/jaxrs/ext/search/Beanspector.java
mrulli: The rationale behind `if (!getterClass.isAssignableFrom(setterClass)) {` is this: setters are neither covariant or contravariant while getters are covariant. So I think the healthy behavior should be the following:

```
public class TestBean {
	public Number getNumber() {
        return 0L;
    }
    public void setNumber(Number val) { }
}

// FAIL (ambiguous setters)
public class InheritableBean1 extends TestBean {
	public void setNumber(Number val) { }
	public void setNumber(Long val) { }
}

// PASS (getters covariance)
public class InheritableBean2 extends TestBean {
	public Long getNumber() {
        return 0L;
    }
}

// PASS (getters covariance and setter override)
public class InheritableBean3 extends TestBean {
	public void setNumber(Number val) { }
	public Long getNumber() {
        return 0L;
    }
}
```

This is provided through the current implementation.

reta: +1, Reasonable, thank you

PR #513, opened 2019-02-20
[CXF-7977] Osgi quickstasrt (keycloak-cxf) fails with wrong protocol 

JiriOndrusek:
Issue https://jira.apache.org/jira/browse/CXF-7977

Make condition to detect https more strict. According to https://github.com/apache/cxf/blob/master/rt/transports/http-undertow/src/main/java/org/apache/cxf/transport/http_undertow/UndertowHTTPServerEngine.java#L277 there has to be present clientAuthentication in case of non http connection.
Test also needed to be enhanced.

### DIFF
--- a/rt/transports/http-undertow/src/main/java/org/apache/cxf/transport/http_undertow/UndertowHTTPServerEngine.java
+++ b/rt/transports/http-undertow/src/main/java/org/apache/cxf/transport/http_undertow/UndertowHTTPServerEngine.java
@@ -390,7 +390,7 @@ public class UndertowHTTPServerEngine implements ServerEngine {
     }
 
     private void retrieveListenerFactory() {
-        if (tlsServerParameters != null) {
+        if (tlsServerParameters != null && tlsServerParameters.getClientAuthentication() != null) {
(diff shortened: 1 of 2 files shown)


JiriOndrusek:
@ffang May I ask for a review? I'm not sure whether condition "tlsServerParameters.getClientAuthentication() != null" is strong enough. Whether it is possible to want https without this parameter.Code from [1] seems that having it is the only way how to force undertow to use ssl. (other option is to check whether at least one property is not null - but it is ugly and error causing solution)

Thank you in advance,
Jiri


[1] https://github.com/apache/cxf/blob/master/rt/transports/http-undertow/src/main/java/org/apache/cxf/transport/http_undertow/UndertowHTTPServerEngine.java#L277

ffang:
Hi @JiriOndrusek 

I don't think your change in that PR is correct.

"tlsServerParameters.getClientAuthentication() != null" isn't a compulsory requirement for  HTTPS protocol

Freeman

JiriOndrusek:
Problem is probably not in cxf. To be sure it needs more investigations

PR #514, opened 2019-02-20
Configure media types in bean for logging intercepters

shadrin-nv:
Need for 
https://issues.apache.org/jira/browse/CXF-7920

### DIFF
--- a/rt/features/logging/src/main/java/org/apache/cxf/ext/logging/LoggingFeature.java
+++ b/rt/features/logging/src/main/java/org/apache/cxf/ext/logging/LoggingFeature.java
@@ -113,7 +113,7 @@ public class LoggingFeature extends AbstractFeature {
         in.setLogMultipart(logMultipart);
         out.setLogMultipart(logMultipart);
     }
-    
+
     public void setVerbose(boolean verbose) {
         setSender(verbose ? new Slf4jVerboseEventSender() : new Slf4jEventSender());
     }
@@ -124,7 +124,7 @@ public class LoggingFeature extends AbstractFeature {
      * Uses, for example:
      * <pre>
      * &lt;bean id="loggingFeature" class="org.apache.cxf.ext.logging.LoggingFeature"&gt;
-     *   &lt;property name="addInBinaryContentMediaTypes" value="application/vnd.oasis.opendocument.spreadsheet;application/zip"/&gt;
+     *   &lt;property name="addInBinaryContentMediaTypes" value="audio/mpeg;application/zip"/&gt;
      * &lt;/bean&gt;
      * </pre>
      * @param mediaTypes list of mediaTypes. symbol ; - delimeter
@@ -139,7 +139,7 @@ public class LoggingFeature extends AbstractFeature {
      * Uses, for example:
      * <pre>
      * &lt;bean id="loggingFeature" class="org.apache.cxf.ext.logging.LoggingFeature"&gt;
-     *   &lt;property name="addOutBinaryContentMediaTypes" value="application/vnd.oasis.opendocument.spreadsheet;application/zip"/&gt;
+     *   &lt;property name="addOutBinaryContentMediaTypes" value="audio/mpeg;application/zip"/&gt;
      * &lt;/bean&gt;
      * </pre>
      * @param mediaTypes list of mediaTypes. symbol ; - delimeter
@@ -154,7 +154,7 @@ public class LoggingFeature extends AbstractFeature {
      * Uses, for example:
      * <pre>
      * &lt;bean id="loggingFeature" class="org.apache.cxf.ext.logging.LoggingFeature"&gt;
-     *   &lt;property name="addBinaryContentMediaTypes" value="application/vnd.oasis.opendocument.spreadsheet;application/zip"/&gt;
+     *   &lt;property name="addBinaryContentMediaTypes" value="audio/mpeg;application/zip"/&gt;
      * &lt;/bean&gt;
      * </pre>
      * @param mediaTypes list of mediaTypes. symbol ; - delimeter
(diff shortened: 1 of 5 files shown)


rt/features/logging/src/main/java/org/apache/cxf/ext/logging/event/DefaultLogEventMapper.java
```diff
@@ -46,17 +44,25 @@
...
+        DEFAULT_BINARY_CONTENT_MEDIA_TYPES.add("application/pdf");
+        DEFAULT_BINARY_CONTENT_MEDIA_TYPES.add("image/png");
+        DEFAULT_BINARY_CONTENT_MEDIA_TYPES.add("image/jpeg");
+        DEFAULT_BINARY_CONTENT_MEDIA_TYPES.add("image/gif");
     }
     private static final String MULTIPART_CONTENT_MEDIA_TYPE = "multipart";
 
+    private final Set<String> binaryContentMediaTypes = new HashSet<>(DEFAULT_BINARY_CONTENT_MEDIA_TYPES);
+
+    public void setBinaryContentMediaTypes(String mediaTypes) {
```
coheigea: Maybe it'd be better to call this "addBinaryContentMediaTypes", as it adds to what's already there, as opposed to setting the values.

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/LoggingFeature.java
```diff
@@ -117,4 +117,50 @@ public void setLogMultipart(boolean logMultipart) {
     public void setVerbose(boolean verbose) {
         setSender(verbose ? new Slf4jVerboseEventSender() : new Slf4jEventSender());
     }
+
+    /**
+     * Set binary media types for Logging In Interceptor.
+     * Contet for these types was not logged.
```
coheigea: "Content" not "Contet" (Same below as well)

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/event/DefaultLogEventMapper.java
shadrin-nv: Correct this

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/LoggingFeature.java
shadrin-nv: Correct this

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/event/DefaultLogEventMapper.java
```diff
@@ -46,17 +44,25 @@
 import org.apache.cxf.ws.addressing.ContextUtils;
 
 public class DefaultLogEventMapper {
-    private static final Set<String> BINARY_CONTENT_MEDIA_TYPES;
+    private static final Set<String> DEFAULT_BINARY_CONTENT_MEDIA_TYPES;
```
andrei-ivanov: Trivial improvement suggestion: since this field seems to want to be immutable, I think it's better to initialize it like this: https://github.com/spring-projects/spring-framework/blob/master/spring-context/src/main/java/org/springframework/format/datetime/DateTimeFormatAnnotationFormatterFactory.java#L43
I know this PR just renames the field, but maybe the initialization can be improved too 🙂

shadrin-nv: Actually, I would remove this set by default. But forcing everyone to tune it is not good.

coheigea: I think what @andrei-ivanov meant here is that this patch could use Collections.unmodifiableSet instead?

shadrin-nv: But, i need modifiable set, in order to be able to customize it.
May be remove DEFAULT_BINARY_CONTENT_MEDIA_TYPES from class and stay only binaryContentMediaTypes varible? And initialize it in constuctor.

andrei-ivanov: Hmm, as far as I see, `DEFAULT_BINARY_CONTENT_MEDIA_TYPES` doesn't get modified, it is copied to `binaryContentMediaTypes` on line 60 and that set is modified.

shadrin-nv: Logically, we don’t need this set at all as a separate instance.
I propose to abandon it. And configure only the editable set at once.
    public DefaultLogEventMapper() {
        binaryContentMediaTypes.add("application/octet-stream");
        binaryContentMediaTypes.add("application/pdf");
        binaryContentMediaTypes.add("image/png");
        binaryContentMediaTypes.add("image/jpeg");
        binaryContentMediaTypes.add("image/gif");
    }

What do you think? Would that be more correct?

coheigea: No I think it's easier just to leave it as it is. I'll make the change to use an unmodifiable set locally.

shadrin-nv: Thanks. Than i close pull request.

coheigea: No, please re-open it, I will merge it and fix up the other thing locally.

shadrin-nv: I apologize for the delay

PR #515, opened 2019-02-21
FileUtils: move test specific getStringFromFile; remove extra space

amarkevich:


core/src/main/java/org/apache/cxf/helpers/FileUtils.java
```diff
@@ -293,67 +290,6 @@ public static File createTempFile(String prefix, String suffix, File parentDir,
         return result;
     }
 
-    public static String getStringFromFile(File location) {
```
coheigea: I'm not sure if it's wise to be removing public methods for a minor patch, maybe we should wait until we move to 3.4.0-SNAPSHOT?

amarkevich: ok

amarkevich: @coheigea WDYT?

PR #516, opened 2019-02-26
[cxf-testutils] ServerLauncher: expose only running server ports

amarkevich:
avoid 'java.io.IOException: CreateProcess error=206, The filename or extension is too long' on Windows


coheigea:
+1.

PR #518, opened 2019-03-05
CXF-7953 re-enable the corba binding and make sure it works with the JDK 11

zhfeng:
https://issues.apache.org/jira/browse/CXF-7953


ffang:
Hi @zhfeng,

Thanks for the patch, glad to see corba binding works again with JDK11.
Just one issue, I can still see CORBADocLitClientTypeTest failure under JDK11. I will exclude it for now so that we will keep investigating it.

Best Regards
Freeman

PR #524, opened 2019-03-18
Avoid to load jaxb when it is useless

rmannibucau:


amarkevich:
cause compilation error; try 'return pClass.cast(value);'

rmannibucau:
@amarkevich thks for the catch, used (T) style as other/nearer casts in this method - just for consistency.

amarkevich:
regarding '(T) style': its used for primitives (L433, L453); 'pClass.cast(...)' used in other places

rmannibucau:
@amarkevich fair enough. Can I ask some deeper review on the "if (pClass == String.class && !adapterHasToBeUsed)" coming next the block I added, if this is important - and somehow covered/coverable in tests - then we should move to the jaxb boolean we have to test if we must try jaxb handling or bypass it cause missing. I'm not fully sure we want to keep that. My understanding is that it is to use @XmlJavaTypeAdapter on method parameter but not sure it is something used or to keep with param converter of JAX-RS.

dkulp:
It doesn't look like we have any test in place that would handle that particular case.   We have XmlAdapters that go from the on-the-wire string to custom types, but nothing that would go from something else on the wire to a string.     That said, I BELIEVE it's something that is supposed to be supported from the specs.

rmannibucau:
@dkulp it is not in JAX-RS spec, do you have another spec in mind?

dkulp:
JAXB in particular.   In any case, I just pushed a test case for it.  :)

rmannibucau:
JAXB has no word about JAX-RS so don't think so as well, it is a CXF specific feature AFAIK
will close this one to open another PR (using my IDE instead of github online edit since it needs more work than I thought)
thanks guys for the feedback

PR #525, opened 2019-03-18
try to skip jaxb when not available in jaxrs

rmannibucau:


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/InjectionUtils.java
```diff
@@ -125,6 +130,19 @@
...
         VALUE_CONTEXTS.add(Application.class.getName());
         VALUE_CONTEXTS.add("javax.ws.rs.sse.Sse");
+
+        boolean useJaxb;
+        try {
+            ClassLoader loader = Thread.currentThread().getContextClassLoader();
+            if (loader == null) {
+                loader = getSystemClassLoader();
+            }
+            loader.loadClass("javax.xml.bind.annotation.adapters.XmlJavaTypeAdapter");
```
reta: You may better use `ClassLoaderUtils.loadClass("javax.xml.bind.annotation.adapters.XmlJavaTypeAdapter", InjectionUtils.class)` since it properly handles the presence of the `SecurityManager`.

rmannibucau: Will do this afternoon but not sure I get it, loadClass does not use the security manager explicitly here (vs setTccl does for instance). Anything Im missing?

reta: That's right, the classloader part (fe `ClassLoader loader = Thread.currentThread().getContextClassLoader();`) may need to run as privileged. That's basically what the  `ClassLoaderUtils.loadClass` helps with, thanks!

rmannibucau:
@reta done and thanks for the catch, completely forgot this point!

PR #526, opened 2019-03-19
CXF-7989: org.apache.cxf.jaxrs.JAXRSInvoker should handle CompletionException

reta:
The Problem
===

In case when CXF is used along with `CompletionStage` / `CompletableFuture`, in most cases the error handling is propagated back through `CompletionException` instance (http://cs.oswego.edu/pipermail/concurrency-interest/2014-August/012911.html). It works but it forces users to introduce own implementations of `ExceptionMapper`s and manually unwrap the `CompletionException` into something more meaningful.

Use cases
===
There are 3 leading use cases which cover most typical usage of the `CompletionStage` / `CompletableFuture`.

1. `@Suspended AsyncResponse`
```
public void get(@PathParam("isbn") String isbn, @Suspended final AsyncResponse resp) {
        client.target("http://localhost:19092/services/catalog/" + isbn)
            .request()
            .rx()
            .get(Book.class)
            .whenComplete((r, ex) -> {
                if (ex != null) {
                    resp.resume(ex);
                } else {
                    resp.resume(r);
                }
            });
    }
```
2. Return `CompletableFuture<Book>`

```
public CompletableFuture<Book> get(@PathParam("isbn") String isbn) {
        return client
            .target("http://localhost:19092/services/catalog/" + isbn)
            .request()
            .rx()
            .get(Book.class)
            .toCompletableFuture();
    }
```

3. Return constructed instance of the `CompletableFuture<Book>` 

```
public CompletableFuture<Book> get(@PathParam("isbn") String isbn) {
    final CompletableFuture<Book> future = new CompletableFuture<Book>();
    future.completeExceptionally(...);
    return future;
}
```

Suggested Solution
===
Instrument `JAXRSInvoker::handleAsyncResponse` to distinguish `CompletionException`, unwrap the cause and use it to select the appropriate `ExceptionMapper` (if any).

@rmannibucau why `JAXRSInvoker::handleAsyncResponse` and not `JAXRSInvoker::checkFutureResponse`, the 1st use case which uses suspended async response does not flow through `JAXRSInvoker::checkFutureResponse` but all three use cases run through `JAXRSInvoker::handleAsyncResponse`

@coheigea @dkulp guys would really appreciate if you could take a quick look, small change but may introduce undesired side effects, thanks a lot.

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/JAXRSInvoker.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/JAXRSInvoker.java
@@ -27,6 +27,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.ResourceBundle;
 import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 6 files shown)


rmannibucau:
@reta agree, your fix is better than my proposal which was ignoring suspended case. Thanks a lot to work on it.

reta:
Thanks for review, @coheigea , sorry I didn't get the whitespace changes thingy, could you please clarify a bit, thanks a lot!

coheigea:
@reta If I apply the patch via "git apply" I get a lot of whitespace warnings:

526.patch:37: trailing whitespace.
            
526.patch:39: trailing whitespace.
                cause = throwable.getCause(); 
526.patch:41: trailing whitespace.
            
526.patch:77: trailing whitespace.
    
526.patch:82: trailing whitespace.
        return CompletableFuture.supplyAsync(() -> { 
warning: squelched 15 whitespace errors
warning: 20 lines add whitespace errors.

reta:
@coheigea done, the patch applies smoothly now, thanks!

PR #529, opened 2019-04-01
CXF-8009: CXF should not rely on ClassUtils for CGLIB proxy checks

reta:
This is a tricky one. So the `ClassUtils.isCglibProxyClass`, which is used by `SpringAopClassHelper`, does check if the class name has `$$` to disambiguate CGLIB proxies from regular classes. The typical examples of CGLIB proxy class names are:
```
SpringProxy$$EnhancerBySpringCGLIB$$d8e73a30
Function$$EnhancerByCGLIB$$53576dae
```
However, the classes generated for lambda expressions do contain `$$` as well, tricking `SpringAopClassHelper` to believe it is a CGLIB proxy, here is an example:
```
org.apache.cxf.common.util.ClassHelperTest$$Lambda$10/846254484@32b260fa
```
It turned out to be difficult to come up with efficient and bullet-proof check (based on class name only) to determine if it is CGLIB generated proxy or not. The cheapest but not 100% reliable way seems to be to check the presence of the `CGLIB` in the class name (this PR).

Alternatively, using the reflection it is possible to achieve more reliable outcomes by inspecting the class, looking for CGLIB generated methods / fields (`CGLIB$BIND_CALLBACKS`). I am not sure it is worth it actually.

@dkulp @coheigea what do you think, guys?

### DIFF
--- a/core/src/main/java/org/apache/cxf/common/util/ClassHelper.java
+++ b/core/src/main/java/org/apache/cxf/common/util/ClassHelper.java
@@ -33,17 +33,47 @@ public class ClassHelper {
     public static final String USE_DEFAULT_CLASS_HELPER = "org.apache.cxf.useDefaultClassHelpers";
 
     static final ClassHelper HELPER;
-    static final ClassHelper DEFAULT_HELPER;
+    static final ClassUnwrapper DEFAULT_UNWRAPPER;
+    static final ClassUnwrapper UNWRAPPER;
+    
+    /**
+     * Default class unwrapper implementation which delegates to the ClassHelper
+     * internal methods.
+     *
+     */
+    private static class DefaultClassUnwrapper implements ClassUnwrapper {
+        private final ClassHelper helper;
+        
+        DefaultClassUnwrapper(ClassHelper helper) {
+            this.helper = helper;
+        }
+        
+        @Override
+        public Class<?> getRealClassFromClass(Class<?> clazz) {
+            return helper.getRealClassFromClassInternal(clazz);
+        }
+        
+        @Override
+        public Class<?> getRealClass(Object o) {
+            return helper.getRealClassInternal(o);
+        }
+        
+        @Override
+        public Object getRealObject(Object o) {
+            return helper.getRealObjectInternal(o);
+        }
+    }
 
     static {
-        DEFAULT_HELPER = new ClassHelper();
-        HELPER = getClassHelper(DEFAULT_HELPER);
+        HELPER = new ClassHelper();
+        DEFAULT_UNWRAPPER = new DefaultClassUnwrapper(HELPER);
+        UNWRAPPER = getClassUnwrapper(DEFAULT_UNWRAPPER);
     }
 
     protected ClassHelper() {
     }
 
-    private static ClassHelper getClassHelper(ClassHelper defaultHelper) {
+    private static ClassUnwrapper getClassUnwrapper(ClassUnwrapper defaultHelper) {
         boolean useSpring = true;
         String s = SystemPropertyAction.getPropertyOrNull("org.apache.cxf.useSpringClassHelpers");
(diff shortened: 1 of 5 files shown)


rmannibucau:
Hi @reta

When working on tomee or cdi based servers i always wondered why this part of the code was not:

1. An extension of the bus allowing to cache the mapping - ClassUnwrapper only?
2. A chain of predicates instead of a single impl which is hard to do right or precise enough for all impls - bytebuddy ones are custom for instance so not manageable ; mixed in an app with spring or a cdi impl you are forced to know all apps internals :(

Guess default can also change since superclass name is often in the proxy name for $$ kind of proxies which enables to keep lambd aand add cglib logic.

Hope it makes sense.

reta:
Hi @rmannibucau ,

Thanks a lot for the review, yeah indeed this is pretty old class, purely Spring-specific. I remember we have discussed this awhile back, when `ClassUnwrapper` / `CdiClassUnwraper` were introduced. Agree that the responsibilities are now half-splitted between `ClassHelper` and `ClassUnwrapper`. 

I think we could do:
1) Convert `SpringAopClassHelper` into `SpringClassUnwrapper` (we would need to add a method
`getRealClassFromClass` to `ClassUnwrapper` to handle the case when instance is not available).
2) Make `SpringClassUnwrapper` logic based on list of predicates with defaults provided (to behave like `SpringAopClassHelper` at least)

Did I get your idea right? Thanks.

rmannibucau:
Hi @reta

1. Sounds ok - we can think about having a getClassFromModel instead, cdi will need Annotated and not Class in some rare cases. For spring model is the class but it would depend the "deployer" - hiw classes are scanned/registered.
2. Almost, the list would be generic. You would get a CompositeClassUnwrapper usable in spring apps but also cdi and standalone apps. Wonder if know proxies should be handled by default or not wrapping user unwrapper hut guess with a composite reusable class taking a list of unwrappers it is not immediately needed.

Hope it is clearer

reta:
@rmannibucau Converted `SpringAopClassHelper` to `SpringClassUnwrapper`, so now `ClassHelper` is completely driven by `ClassUnwrapper`s. I had to preserve the default behavior which uses the system properties, I know that people use it (recall multiple discussions around that). 

I introduced 2 new methods to `ClassUnwrapper`: `getRealClassFromClass` and `getRealObject`. I kept with `getRealClassFromClass` (not the `getClassFromModel` as you suggested) because it looks to me that `ClassUnwrapper` should not really care about underlying model (CDI/Spring), but it could be used to back different models if needed (fe Annotated, BeanDefinition, ...). The `getRealObject` is outlier but really  needed to fully replace the `ClassHelper` functionality (it is a default method, no need to implement).

Regarding  `CompositeClassUnwrapper ` it is absolutelypossible now, I didn't introduce it yet since I just don't have use case at this moment. If PR looks good, we could build on top.

rmannibucau:
Looks good, worse case a default method would solve the model api issue. It can happen since cdi enables to switch class model dynamically but it is rare for the case we are speaking about so probably enough for years ;). Thanks for the change, very appreciated.

reta:
Thanks a lot @rmannibucau, @coheigea do you have any concerns? Thanks guys.

andrei-ivanov:
Just as a comment from a bystander, since Spring Framework classes were used as inspiration, maybe `getTargetClass` and `getTarget` sound better?
Taken from `org.springframework.aop.TargetSource` and `org.springframework.aop.support.AopUtils`

reta:
Thanks @andrei-ivanov, the CXF adopted `getRealXxx` prefix for this particular case, for better or worst, `getTargetXxx` would have worked as well I believe. I don't see strong reasons to prefer one over another, but renaming would break the existing user's code (since all the classes are in use).

coheigea:
@reta No objections from me.

PR #532, opened 2019-04-01
handle WstxLazyException for pretty logging

SmirnovKirill:


rt/features/logging/src/main/java/org/apache/cxf/ext/logging/event/PrettyLoggingFilter.java
```diff
@@ -77,9 +78,9 @@ private String getPrettyMessage(LogEvent event) {
         try {
             StaxUtils.copy(xreader, xwriter);
             xwriter.flush();
-        } catch (XMLStreamException xse) {
+        } catch (XMLStreamException | WstxLazyException e) {
```
dkulp: This would put a hard dependency on woodstox in the logging module.   I believe the better option would be to catch RuntimeException and check the cause to see if the cause is a XMLStreamException.  If not, rethrow the RuntimeException, otherwise treat it like the XMLStreamException.

SmirnovKirill: But can't we just catch any exception without any additional checks? Because code in the catch block doesn't depend on any particular exception class. And it's definitely not good if any exception will go beyond this method because in this case logging will cause crash of the whole request.

SmirnovKirill: I did what you proposed (in the second commit) although I think it would be cleaner and more stable just to catch all possible exceptions.

PR #533, opened 2019-04-02
Drop @PreMatching - no need for it

davidkarlsen:
Signed-off-by: David J. M. Karlsen <david@davidkarlsen.com>

As discussed on mailinglist:
http://cxf.547215.n5.nabble.com/VerifySignatureFilter-phase-td5796123.html

### DIFF
--- a/rt/rs/security/http-signature/src/main/java/org/apache/cxf/rs/security/httpsignature/filters/VerifySignatureClientFilter.java
+++ b/rt/rs/security/http-signature/src/main/java/org/apache/cxf/rs/security/httpsignature/filters/VerifySignatureClientFilter.java
@@ -23,7 +23,6 @@ import javax.ws.rs.Priorities;
 import javax.ws.rs.client.ClientRequestContext;
 import javax.ws.rs.client.ClientResponseContext;
 import javax.ws.rs.client.ClientResponseFilter;
-import javax.ws.rs.container.PreMatching;
 import javax.ws.rs.ext.Provider;
 
 import org.apache.cxf.rs.security.httpsignature.exception.SignatureException;
@@ -32,7 +31,6 @@ import org.apache.cxf.rs.security.httpsignature.exception.SignatureException;
  * RS CXF client Filter which extracts signature data from the context and sends it to the message verifier
  */
 @Provider
-@PreMatching
 @Priority(Priorities.AUTHENTICATION)
 public class VerifySignatureClientFilter extends AbstractSignatureInFilter implements ClientResponseFilter {
 
--- a/rt/rs/security/http-signature/src/main/java/org/apache/cxf/rs/security/httpsignature/filters/VerifySignatureFilter.java
+++ b/rt/rs/security/http-signature/src/main/java/org/apache/cxf/rs/security/httpsignature/filters/VerifySignatureFilter.java
@@ -23,14 +23,12 @@ import javax.ws.rs.BadRequestException;
 import javax.ws.rs.Priorities;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerRequestFilter;
-import javax.ws.rs.container.PreMatching;
 import javax.ws.rs.ext.Provider;
 
 /**
  * RS CXF container Filter which extracts signature data from the context and sends it to the message verifier
  */
 @Provider
-@PreMatching
 @Priority(Priorities.AUTHENTICATION)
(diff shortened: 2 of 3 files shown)


davidkarlsen:
@coheigea

rt/rs/security/http-signature/src/main/java/org/apache/cxf/rs/security/httpsignature/filters/VerifySignatureFilter.java
```diff
@@ -23,14 +23,12 @@
...
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerRequestFilter;
-import javax.ws.rs.container.PreMatching;
 import javax.ws.rs.ext.Provider;
 
 /**
  * RS CXF container Filter which extracts signature data from the context and sends it to the message verifier
  */
 @Provider
-@PreMatching
```
coheigea: I think I'd prefer to keep it for the ContainerRequestFilter case - CXF does this for all of the other ContainerRequestFilter security implementations in JOSE, XML Security, etc.

rt/rs/security/http-signature/src/main/java/org/apache/cxf/rs/security/httpsignature/filters/VerifySignatureClientFilter.java
```diff
@@ -32,7 +31,6 @@
  * RS CXF client Filter which extracts signature data from the context and sends it to the message verifier
  */
 @Provider
-@PreMatching
```
coheigea: You could also get rid of it in CreateSignatureFilter

rt/rs/security/http-signature/src/main/java/org/apache/cxf/rs/security/httpsignature/filters/VerifySignatureFilter.java
davidkarlsen: Hm - that's actually where I need it in order to do:

```
+@Provider
+@Priority(Priorities.AUTHENTICATION)
+public class DynamicVerifySignatureFilter implements ResourceInfoAware, ContainerRequestFilter {
+  private final Logger logger = LoggerFactory.getLogger(DynamicVerifySignatureFilter.class);
+  private final ContainerRequestFilter delegate;
+  private ResourceInfo resourceInfo;
+
+  public DynamicVerifySignatureFilter(ContainerRequestFilter delegate) {
+    this.delegate = delegate;
+  }
+
+  @Override
+  public void filter(ContainerRequestContext requestContext) throws IOException {
+    if (isIgnored()) {
+      logger.warn(
+          "Resource: class: {} method: {} is exempted for signature-check",
+          resourceInfo.getResourceClass(),
+          resourceInfo.getResourceMethod().getName());
+    } else {
+      this.delegate.filter(requestContext);
+    }
+  }
+
+  private boolean isIgnored() {
+    return null
+            != AnnotationUtils.findAnnotation(
+                getResourceInfo().getResourceMethod(), IgnoreSigning.class)
+        || null
+            != AnnotationUtils.findAnnotation(
+                getResourceInfo().getResourceClass(), IgnoreSigning.class);
+  }
+
+  @Override
+  public void setResourceInfo(ResourceInfo resourceInfo) {
+    this.resourceInfo = resourceInfo;
+  }
+
+  private ResourceInfo getResourceInfo() {
+    return new ResourceInfoImpl(PhaseInterceptorChain.getCurrentMessage());
+  }
```

or else the resourceInfo can't be resolved at this stage.
I can however keep on wrapping it in a delegate filter as shown to defer the decision until the matcher has found the target resource - but it adds cruft.

davidkarlsen: What is the reason for having it in @PreMatching as https://docs.jboss.org/resteasy/docs/3.0.21.Final/userguide/html/Interceptors.html seems to suggest that the use-case for PreMatching is mainly for altering the request (which might influence the matcher):

`Pre-matching ContainerRequestFilters are designated with the @PreMatching annotation and will execute before the JAX-RS resource method is matched with the incoming HTTP request. Pre-matching filters often are used to modify request attributes to change how it matches to a specific resource method (i.e. strip .xml and add an Accept header). ContainerRequestFilters can abort the request by calling ContainerRequestContext.abortWith(Response). A filter might want to abort if it implements a custom authentication protocol.`

PR #538, opened 2019-04-09
[CXF-7990]:Fix Infinite loop when service endpoint throws SOAPFaultEx…

jimma:
…ception

### DIFF
--- a/systests/jaxws/src/test/java/org/apache/cxf/systest/jaxws/EchoServiceImpl.java
+++ b/systests/jaxws/src/test/java/org/apache/cxf/systest/jaxws/EchoServiceImpl.java
@@ -1,0 +1,63 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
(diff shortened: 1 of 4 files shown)


systests/jaxws/src/test/java/org/apache/cxf/systest/jaxws/EchoServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+import javax.xml.soap.SOAPConstants;
+import javax.xml.soap.SOAPFactory;
+import javax.xml.soap.SOAPFault;
+import javax.xml.ws.soap.SOAPFaultException;
+
+import org.apache.cxf.ext.logging.Logging;
+
+@WebService(name = "MyEchoService", targetNamespace = "urn:echo")
+@Logging
+public class EchoServiceImpl {
```
coheigea: "implements EchoService"

core/src/main/java/org/apache/cxf/interceptor/FaultOutInterceptor.java
```diff
@@ -25,6 +25,7 @@
 import java.util.logging.Logger;
 
 import javax.xml.stream.XMLStreamWriter;
+import javax.xml.ws.soap.SOAPFaultException;
```
dkulp: This makes core depend on SAAJ which is definitely not what we'd want to do.   Let me take a look at the test case to see if there is a different way to handle this from the jaxws fault interceptors.

dkulp:
I believe the better patch would be:

```
--- a/rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/interceptors/WebFaultOutInterceptor.java
+++ b/rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/interceptors/WebFaultOutInterceptor.java
@@ -181,6 +181,8 @@ public class WebFaultOutInterceptor extends FaultOutInterceptor {
                 //and let the rest of the chain try handling it as is.
                 LOG.log(Level.WARNING, "EXCEPTION_WHILE_WRITING_FAULT", nex);
             }
+        } else if (cause instanceof SOAPFaultException && ((SOAPFaultException)cause).getFault().hasDetail()) {
+            return;
         } else {
             FaultMode mode = message.get(FaultMode.class);
             if (mode == FaultMode.CHECKED_APPLICATION_FAULT) {
```
which moves the check for the detail not the jaxws frontend which is where the handling for the SAAJ based things should be done.

jimma:
Thanks for review @dkulp @coheigea . Change has pushed.

PR #540, opened 2019-04-10
handle signing and digest together

davidkarlsen:
Signed-off-by: David J. M. Karlsen <david@davidkarlsen.com>


davidkarlsen:
I think it needs to be solved like this, due to:
* writerinterceptors are only called for requests with body
* signing needs to happen after any digest
* hence have a filter which does signing if no body, else let the interceptor handle both digest+signing

Also note that digest is not strictly required by the spec - so it it's debatable if it should be optional - and in this case the signing filter should always be active - but I think the current solution is a sane start.

@coheigea feel free to reach out on chat if you want to discuss this more.

a lot of the integration-tests are failing on missing signing configuration.

coheigea:
I'll take care of the failing tests once we've agreed a fix. I think the basic idea sounds good. I wonder though if it's possible for the filter to add the interceptor as a provider dynamically "under the hood", rather than expose it to the user? It's a bit confusing to have two separate ways of adding HTTP Signature, one for an empty request + one for when you want digest as well.

davidkarlsen:
I have investigated a bit, but https://docs.oracle.com/javaee/7/api/javax/ws/rs/container/DynamicFeature.html only applies server-side.

You have http://cxf.apache.org/docs/features.html - but that's for CXF interceptors on the bus - so I'm out of options.

davidkarlsen:
@coheigea PTAL?

davidkarlsen:
@coheigea fixed

davidkarlsen:
@coheigea ping -  want to finish this off so we're ready for a release

davidkarlsen:
Shure. Just ask for help if you need any. Hope we can reach next release.

PR #541, opened 2019-04-12
CXF-8020 use System.currentTimeMillis()

bd2019us:
new Date() is just a thin wrapper around System.currentTimeMillis(). Using System.currentTimeMillis() can help speed up the system


reta:
These are just sample projects, it is very unlikely that `System.currentTimeMillis()` vs `new Date().getTime()` in this particular case would make the difference. Thank you.

PR #542, opened 2019-04-16
CXF-8022: Thread hangs using Reactor Flux when Exception is Thrown

reta:
Indeed, there is an issue in case of error, opening the PR to get some feedback while working on the test cases. According to Reactive Streams spec (https://github.com/reactive-streams/reactive-streams-jvm), the `onError` is the terminal state:

> For a Publisher: When onComplete or onError has been signalled. For a Subscriber: When an onComplete   or onError has been received.

However, `StreamingAsyncSubscriber` does not treat it as that.

@coheigea @rmannibucau @davidkarlsen if you guys can validate my assumptions here, thanks.

### DIFF
--- a/rt/rs/extensions/reactor/src/main/java/org/apache/cxf/jaxrs/reactor/client/ReactorInvokerImpl.java
+++ b/rt/rs/extensions/reactor/src/main/java/org/apache/cxf/jaxrs/reactor/client/ReactorInvokerImpl.java
@@ -25,6 +25,7 @@ import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.stream.StreamSupport;
 
 import javax.ws.rs.HttpMethod;
 import javax.ws.rs.client.Entity;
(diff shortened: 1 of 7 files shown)


rmannibucau:
@reta looks good to me, I would add a test cause it is the kind of thing easy to break while fighting against a concurrency issue for instance

reta:
@rmannibucau absolutely, working on test cases right now, takes a bit more time, thanks!

rt/rs/extensions/reactor/src/main/java/org/apache/cxf/jaxrs/reactor/client/ReactorInvokerImpl.java
```diff
@@ -201,13 +202,15 @@
...
-        return Flux.fromIterable(toIterable(futureResponse, responseType));
+        return Flux.fromStream(() -> 
+            StreamSupport.stream(toIterable(futureResponse, responseType).spliterator(), false));
     }
 
     @Override
     public <T> Flux<T> flux(String name, Class<T> responseType) {
         Future<Response> futureResponse = webClient.async().method(name);
-        return Flux.fromIterable(toIterable(futureResponse, responseType));
+        return Flux.fromStream(() ->
```
reta: So when the `toIterable` fails (fe empty response, or alike), it does thrown the exception from `getFlux` method, which is unexpected. Instead, the exception should be propagated using `onError` callback, following the reactive stream.

PR #543, opened 2019-04-17
[CXF-8025]: Fix Wrapper Style for multilevel xsd:extension elements

jpoje-gtri:
Original support was added in [CXF-2193].  This allows for multiple
levels of xsd:extension in elements while still using wrapper style.

### DIFF
--- a/tools/wsdlto/frontend/jaxws/src/main/java/org/apache/cxf/tools/wsdlto/frontend/jaxws/processor/internal/ProcessorUtil.java
+++ b/tools/wsdlto/frontend/jaxws/src/main/java/org/apache/cxf/tools/wsdlto/frontend/jaxws/processor/internal/ProcessorUtil.java
@@ -357,8 +357,6 @@ public final class ProcessorUtil {
     }
 
     public static List<WrapperElement> getWrappedElement(ToolContext context, QName partElement) {
-        List<WrapperElement> qnames = new ArrayList<>();
-
         ServiceInfo serviceInfo = context.get(ServiceInfo.class);
         SchemaCollection schema = serviceInfo.getXmlSchemaCollection();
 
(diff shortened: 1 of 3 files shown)


jpoje-gtri:
Weird, that test passes for me locally.  I ran the full test suite with no failures on the cxf-3.3.1 tag and a few failures on master that were exactly the same with/without the patch (master from Tuesday).  This was on OSX with java 8 and maven 3.6.  Thoughts?

deki:
Don't worry, sometimes there are flanky tests on Jenkins. Thanks for your PR, looks good to me, but let's wait for another review before merging it.

tools/wsdlto/frontend/jaxws/src/main/java/org/apache/cxf/tools/wsdlto/frontend/jaxws/processor/internal/ProcessorUtil.java
```diff
@@ -368,32 +366,46 @@ public static Node cloneNode(Document document, Node node, boolean deep) throws
...
+        XmlSchemaComplexContentExtension extension = (XmlSchemaComplexContentExtension)schemaContent;
+        QName baseTypeName = extension.getBaseTypeName();
+        XmlSchemaType baseType = schema.getTypeByQName(baseTypeName);
+
+        if (baseType instanceof XmlSchemaComplexType) {
+            XmlSchemaComplexType complexBaseType = (XmlSchemaComplexType)baseType;
+
+            if (complexBaseType.getParticle() == null && complexBaseType.getContentModel() != null) {
+                // continue up the extension ladder
+                qnames.addAll(createWrappedElementsFromExtension(schema, complexBaseType));
```
coheigea: I'd be afraid here that this could lead to a stack overflow exception on a malicious schema. Could we add a reasonable hard-coded limit on the recursion depth?

jpoje-gtri: Seems like a good thing to handle.  If there is interest in this PR's approach from a high level, I don't mind shepherding it to completion.  I was primarily making sure a test case was available but if the fix is accepted that's awesome too.  If this approach is on the right track please give me some feedback for the questions below and I'll get a tweaked commit up this week probably.

Questions:
1. Is there a CXF best practice for handling recursive traversals?  A brief search shows something similar in DepthRestrictingStreamReader, with hardcoded limits throwing a custom exception when max depth is exceeded.
1. What kind of limit makes sense here?

coheigea: I guess we need to add a new optional configuration option to specify the stack depth. It could default to something relatively small like 3. We can always tweak this in the future if required.

jpoje-gtri: Just wanted to drop a note that I have not forgotten about this, it is just going to be a while before I can circle back and knock it out.

coheigea: Any update?

coheigea: I'll fix the stack depth issue myself separately.

PR #545, opened 2019-04-19
If there is no inMessage use the inFaultMessage

spark404:
I observed that when reading an entity from a response of a failed rest call the CXF throws a NullPointer exception.

I tracked this down to specific case where a JAXRS ReaderInterceptor is present on the chain and the response fails. When parsing the response this condition triggers a different path in JAXRSUtils.readFromMessageBodyReader which eventually results in a call to ProviderFactory.createMessageBodyReaderInterceptor which will throw a NullPointerException when there is no inMessage.

This fix tries to load  an inFaultMessage if there is no inMessage available which prevents the NPE and allows the code to execute successfully.

To reproduce create a JAX-RS Proxy client with a ReaderInterceptor. Then trigger a request that returns a 403 status code and try to 'readEntity' on the resulting ForbiddenException response.

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ReaderInterceptorMBR.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ReaderInterceptorMBR.java
@@ -18,9 +18,11 @@
  */
 package org.apache.cxf.jaxrs.impl;
 
-import org.apache.cxf.jaxrs.provider.ProviderFactory;
-import org.apache.cxf.jaxrs.utils.JAXRSUtils;
-import org.apache.cxf.message.Message;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.Reader;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
 
 import javax.ws.rs.ProcessingException;
 import javax.ws.rs.WebApplicationException;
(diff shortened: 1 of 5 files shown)


spark404:
> Thanks for the PR. Could you add a unittest for it?

Sure, i'll see what i can do. Already looked at it and this method wasn't covered yet. So i'll have to write the test for the existing behaviour as well

spark404:
BTW, Build failed due to a java OutOfMemory exception. Any way to restart it?

spark404:
@deki I added the tests you mentioned. The build keeps failing due to issues with the Apache Jenkins. Do you permissions to restart the build?

coheigea:
The tests added in ProviderFactoryTest pass even without the changes in src/main, and so they are not really testing the change. I would have expected the tests to fail with the old code, and pass with the new.

spark404:
@coheigea Yes. I started out by making the tests fail in those unit tests. However I had to setup a pretty large amount of mocks and additional classes. To prove this you basically need to setup a complete entity reader that parses an init stream and eventually calls a method that was initialised wrongly with a null value. I managed to get quite far but the test looked closed to a system test than a unit test.  Thinking about it again i concluded that the actual fix should be that the initialisation of the ReaderInterceptorMBR should just not accept a null value as message. So that is that actual fix with a associated unittest. The other new unittest do indeed not trigger on this error because it happens downstream in other flow, still i wrote them anyway so they might be useful to have. Long story short, the proof is the null check in the ReaderInterceptorMBR. I would be interested in adding a system test to prove that the entire chain works as designed now, but i would ask for some pointers on where to start.

coheigea:
Yes a system test sounds like the way to go. You can add it in systests/jax-rs. There are lots of tests in there that you can copy to try to reproduce the original error.

PR #548, opened 2019-04-25
[CXF-8029] Add redirect support to URIResolver

ivy-arus:


ivy-arus:
No Problem, my very first version had a flag included.
But I wasn’t sure what name to use. I saw that there is a property ‘http.autoredirect’ in the jax-rs implementation (class org.apache.cxf.jaxrs.client.ClientProperties). Should I use this property name?

PR #550, opened 2019-04-29
CXF-8028: try to fix performance for very big document in DOMUtils

gillesdiacon:
CXF-8028: try to fix performance for very big document in DOMUtils.getDomElement() and DOMUtils.getDomDocumentFragment()


dkulp:
I believe this would be much better off using a java.lang.ClassValue to store the method (or null).   The reflection code would only happen once per class (good) and the value directly stored with the class so nothing is in a static that that would prevent garbage collection.

coheigea:
@gillesdiacon - Does @dkulp 's suggestion sounds OK for you? It would be good to get this fix into the next CXF release.

dkulp:
I committed a fix that uses the ClassValue approach.  The test you provided shows a massive speed increase which is awesome.   Major thanks for identifying the issue.

Fix is in 1cc03b9b70bc509d8f8c631563c3964e56e30621

gillesdiacon:
@dkulp: sorry for lack of reaction. Your proposition is very good and better than my static sets :-)

PR #551, opened 2019-05-02
CXF-8033: MP Rest Client 1.3 Implementation

andymc12:
Still a work in progress, but this PR implements the MicroProfile Rest Client 1.3 spec - based on the [1.3-RC1 release candidate](https://github.com/eclipse/microprofile-rest-client/releases/tag/1.3-RC1).


systests/microprofile/pom.xml
```diff
@@ -33,7 +33,7 @@
     <url>http://cxf.apache.org</url>
     <properties>
         <cxf.geronimo.config.version>1.0</cxf.geronimo.config.version>
-        <cxf.microprofile.rest.client.version>1.2.1</cxf.microprofile.rest.client.version>
+        <cxf.microprofile.rest.client.version>1.3-RC1</cxf.microprofile.rest.client.version>
```
coheigea: Maybe these versions could be consolidated a bit? On master I see:

parent/pom.xml ->  <cxf.microprofile.rest.client.version>1.0.1</cxf.microprofile.rest.client.version>
systests/microprofile/client/jaxrs ->  <cxf.microprofile.rest.client.version>1.2.0</cxf.microprofile.rest.client.version>
systests/microprofile/async/jaxrs -> <cxf.microprofile.rest.client.version>1.2.0</cxf.microprofile.rest.client.version>
rt/rs/microprofile/client -> <mp.rest.client.version>1.2.0</mp.rest.client.version>
systests/microprofile ->  <cxf.microprofile.rest.client.version>1.2.1</cxf.microprofile.rest.client.version>

andymc12: Done.  Thanks for the suggestion!

systests/microprofile/pom.xml
```diff
@@ -42,6 +42,7 @@
         <cxf.arquillian.version>1.1.14.Final</cxf.arquillian.version>
         <cxf.slf4j.version>1.7.25</cxf.slf4j.version>
         <cxf.commons.logging.version>1.2</cxf.commons.logging.version>
+        <cxf.jetty.version>9.4.14.v20181114</cxf.jetty.version>
```
coheigea: Can we import some of these properties from the CXF parent pom as well? For example, Jetty, Johnzon, Weld, SLF4J, etc.

andymc12: Hi @coheigea - I'll try to consolidate the Johnson, Weld, SLF4J, etc. into the parent pom.  There is a particular problem with newer versions of Jetty - to workaround it, I think we might need to keep the (downlevel) jetty version specified in the MP systest. The problem is described in issue https://github.com/eclipse/jetty.project/issues/3656 - since it only seems to affect the MP Rest Client TCKs, I didn't want to downgrade it for everybody.

coheigea: Sounds good @andymc12 . Re the Jetty issue, I recently upgraded CXF and worked around the SSL client auth issue - see here: https://github.com/apache/cxf/commit/af6a95f5b7efc141cc55792e10d05d471905cf32#diff-d0e45878043844ffc41aac437e86b602

rt/rs/extensions/providers/pom.xml
```diff
@@ -114,12 +114,18 @@
             <groupId>javax.json</groupId>
             <artifactId>javax.json-api</artifactId>
             <optional>true</optional>
-        </dependency>          
+        </dependency>
+        <dependency>
```
reta: This dependency may go in conflict with `geronimo-jsonb_1.0_spec` (https://github.com/apache/cxf/blob/CXF-7601_microProfileOpenApi/parent/pom.xml#L2213) which we would need for Geronimo OpenAPI integration ...

andymc12: No problem.  It looks like the geronimo-jsonb_1.0_spec dependency works for me too, so I'll switch to that.  Thanks for the review!

systests/microprofile/pom.xml
andymc12: Thanks @coheigea - We will update the MP Rest Client TCK with that change (I verified that it works locally).  If it's ok with you, I'd still like to leave this older version in (for the MP Rest Client tests only) until we release a new version of the TCK with these changes.

PR #552, opened 2019-05-09
JWT Token lifetime incorrectly calculated causing token to expire.

onlinenguyen:
The token lifetime is incorrectly calculated to a negative value because claims expiry time is in second interval and code is setting issued at value in millisecond interval. Consequently causing token to expire.

### DIFF
--- a/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/filters/JwtAccessTokenValidator.java
+++ b/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/filters/JwtAccessTokenValidator.java
@@ -18,7 +18,6 @@
  */
 package org.apache.cxf.rs.security.oauth2.filters;
 
-import java.time.Instant;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/filters/JwtAccessTokenValidator.java
```diff
@@ -41,6 +41,7 @@
 import org.apache.cxf.rs.security.oauth2.provider.OAuthServiceException;
```
reta: Could you please remove `java.time.Instant` import? It is not used, thank you.

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/filters/JwtAccessTokenValidator.java
```diff
@@ -76,8 +77,7 @@ private AccessTokenValidation convertClaimsToValidation(JwtClaims claims) {
         if (claims.getIssuedAt() != null) {
             atv.setTokenIssuedAt(claims.getIssuedAt());
         } else {
-            Instant now = Instant.now();
-            atv.setTokenIssuedAt(now.toEpochMilli());
+            atv.setTokenIssuedAt(OAuthUtils.getIssuedAt());
```
reta: Indeed,  you are right, it should be in seconds, as per https://tools.ietf.org/html/rfc7519
> value representing the number of seconds from 1970-01-01T00:00:00Z UTC until the specified UTC date/time

reta:
@onlinenguyen Thank you for the PR, one minor comment (unused import) to let the build pass, LGTM otherwise.

PR #556, opened 2019-05-23
Set jaxb-xjc back to scope test

dblevins:
Was switched in commit 5544a990e1272bf72fbbb7bb5104409d020328e3

https://github.com/apache/cxf/commit/5544a990e1272bf72fbbb7bb5104409d020328e3#diff-357e4854869b2e21c38b1b437f11095a

This was likely by mistake.  The result is the `maven-war-plugin` and `maven-assembly-plugin` is pulling `jaxb-xjc` and all dependencies such as `ant` into war files, etc.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -178,6 +178,7 @@
         <dependency>
             <groupId>org.glassfish.jaxb</groupId>
             <artifactId>jaxb-xjc</artifactId>
+	    <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.osgi</groupId>


coheigea:
@ffang Can you take a look?

ffang:
With this change, we also need add  jaxb-xjc dependency in test scope for java11, otherwise java11 build would be broken, I did it on master, so this change in core/pom.xml can be merged.

Thanks!

dkulp:
Why is jaxb-xjc a dependency at all?   All the tests in core pass without it.

ffang:
> Why is jaxb-xjc a dependency at all? All the tests in core pass without it.

Hi @dkulp,

Yes, you are correct, actually we don't need it, removed and we can close this PR now.

Thanks!
Freeman

ffang:
Hi @dblevins,

Thanks for your contribution, found this issue in the first place.
Please let us know do you have further concern about it.

Freeman

PR #558, opened 2019-05-28
CXF-8049: Enclosing class path entries that contain whitespaces in double quotes

vadakr:


ffang:
@vadakr thanks for the contribution!
The patch is good, just a few checkstyle correction is needed

core/src/main/java/org/apache/cxf/common/util/Compiler.java
```diff
@@ -85,6 +85,31 @@ public void setClassPath(String s) {
...
+                    correctedEntries.add("\"" + tok + "\"");
+                } else {
+                    correctedEntries.add(tok);
+                }
+            }
+
+            return String.join(File.pathSeparator, correctedEntries);
+        }
+
+        return "";
```
coheigea: It could just return javaClasspath here

ffang: +1
@coheigea , sorry I missed your comment and merged the PR, I will make another commit to address accordingly.

Cheers
Freeman

PR #561, opened 2019-06-01
[CXF-8053] portable feature to avoid jaxws for jaxrs apps

rmannibucau:
This PR aims at enabling users to rely on CXF features without the need of jaxws API in the classpath (targets JAXRS users and java >= 9).

### DIFF
--- a/core/src/main/java/org/apache/cxf/databinding/stax/StaxDataBindingFeature.java
+++ b/core/src/main/java/org/apache/cxf/databinding/stax/StaxDataBindingFeature.java
@@ -25,31 +25,44 @@ import org.apache.cxf.Bus;
 import org.apache.cxf.endpoint.Client;
 import org.apache.cxf.endpoint.Server;
 import org.apache.cxf.feature.AbstractFeature;
+import org.apache.cxf.feature.AbstractPortableFeature;
 import org.apache.cxf.interceptor.AbstractInDatabindingInterceptor;
 import org.apache.cxf.interceptor.Interceptor;
 import org.apache.cxf.message.Message;
 
 public class StaxDataBindingFeature extends AbstractFeature {
-
+    private Portable delegate = new Portable();
 
     @Override
     public void initialize(Client client, Bus bus) {
-        removeDatabindingInterceptor(client.getEndpoint().getBinding().getInInterceptors());
+        delegate.initialize(client, bus);
     }
 
     @Override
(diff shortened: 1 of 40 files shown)


rmannibucau:
retest this please

dkulp:
I'm wondering if it would make sense to insert a layer between AbstractFeature and the actual features like:  
```
public abstract class AbstractDelegatingFeature<T extends AbstractPortableFeature> extends AbstractFeature {
    protected final T delegate;
    
    protected  AbstractDelegatingFeature(T d) {
        delegate = d;
    }
    
    @Override
    public void initialize(Server server, Bus bus) {
        delegate.initialize(server, bus);
    }

    @Override
    public void initialize(Client client, Bus bus) {
        delegate.initialize(client, bus);
    }

    @Override
    public void initialize(InterceptorProvider interceptorProvider, Bus bus) {
        delegate.initialize(interceptorProvider, bus);
    }

    @Override
    public void initialize(Bus bus) {
        delegate.initialize(bus);
    }

    @Override
    protected void initializeProvider(InterceptorProvider interceptorProvider, Bus bus) {
        delegate.doInitializeProvider(interceptorProvider, bus);
    }

}
```

and then each of the subclasses would look something like:
```
public class GZIPFeature extends AbstractDelegatingFeature<GZIPFeature.Portable> {
    public GZIPFeature() {
        super(new Portable());
    }
...
}
```

and a lot of the duplicate code in each of the subclasses goes away. 

Thoughts?

rmannibucau:
I'd say it depends the future of cxf features. If AbstractFeature/Feature get deprecated it is probably saner to not simplify that. If not you are right.
I am indeed biased toward portability and to "drop" - in 3.4? - AbstractFeature as a central extension point since future of cxf is java >= 11 but it probably require a vote.

Side note: if we want to take time to discuss it I am ok to take the refactor task myself once decided (or not) later but can be good to merge before it requires much work since it impacts almost all features.

Hope it makes sense

dkulp:
AbstractFeature cannot change (and any of its subclasses) until at CXF 4.0 as it would affect the public API and class structures and break compile time compatibility for a majority of CXF users.   Thus, I'd definitely lean toward making it "cleaner".  It also provides a better model to help users that may have their own AbstractFeature subclasses to migrate to the new model a bit quicker/easier.

rmannibucau:
@dkulp thanks to clarify that quickly, I updated the PR

PR #566, opened 2019-07-11
CXF-8066: Support Doclet API (JDK13+)

reta:
The problem
===
Our `java2wadl` plugin uses some part of Javadoc API to extract the documentation pieces from the
Java classes / methods / parameters. Since OpenJDK13, this API has been removed completely [2], [3]. To support JDK13 and above, we essentially have to rewrite the implementation completely. However, we would face another blocker here: how to support JDK8-11 and JDK13+ at the same time?

The solution
===
Use multi-release JAR. This is exactly the use case this feature has been designed for. So what it means 
is that `java2wadl` has additional source folder `src/main/java13` with JDK13 implementation. It would be it however there is another subtle complication. 

The complication 
===
For multi-release JAR, the java2wadl has to be build with 2 JDKs (ideally), pre-JDK13 and JDK13, or 
alternatively by latest JDK13 twice, once for `src/main/java` (using `--release 8`) and once for `src/main/java13` (equivalent to `--release 13`). The issue is that Javadoc-related API comes from `tools.jar`, which is not in scope of javac's `--release` flag [4], it really needs to be built by JDK8. As the workaround, we could used `--release 9` to mitigate this issue (since `tools.jar` was merged into JDK9+). This is not ideal (see please the next paragraph) but we get JDK13 builds back on track.

The impact
===
As of now, since our release jobs are using JDK8, there is no impact on release artifacts BUT `java2wadl` 
will fail when used with JDK13+ea builds. When JDK13 is out, we could rely on toolchain plugin to build `java2wadl` with 2 JDKs (`JDK8` and `JDK13`).

I think this is the first real precedent when we have to deal with large features / API removal from JDK, whereas reflection served as very well before. But this is certainly not the last one. 

[2] https://bugs.openjdk.java.net/browse/JDK-8215608
[3] https://jdk.java.net/13/release-notes 
[4] https://bugs.openjdk.java.net/browse/JDK-8199325

### DIFF
--- a/maven-plugins/java2wadl-plugin/pom.xml
+++ b/maven-plugins/java2wadl-plugin/pom.xml
@@ -208,5 +208,56 @@
                 </plugins>
             </build>
         </profile>
+        
+        <profile>
+            <id>jdk13</id>
+              <activation>
+                  <jdk>[13,)</jdk>
+              </activation>
+              <build>
+                  <pluginManagement>
+                      <plugins>
+                          <plugin>
+                              <groupId>org.apache.maven.plugins</groupId>
+                              <artifactId>maven-compiler-plugin</artifactId>
+                              <executions>
+                                  <execution>
+                                    <id>default-jdk13-compile</id>
+                                    <goals>
+                                        <goal>compile</goal>
+                                    </goals>
+                                    <configuration>
+                                        <forceJavacCompilerUse>true</forceJavacCompilerUse>
+                                        <compileSourceRoots>
+                                            <compileSourceRoot>${project.basedir}/src/main/java13</compileSourceRoot>
+                                        </compileSourceRoots>
+                                        <outputDirectory>${project.build.outputDirectory}/META-INF/versions/13</outputDirectory>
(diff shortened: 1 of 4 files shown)


maven-plugins/java2wadl-plugin/pom.xml
```diff
@@ -208,5 +208,56 @@
...
+                                        <forceJavacCompilerUse>true</forceJavacCompilerUse>
+                                        <compileSourceRoots>
+                                            <compileSourceRoot>${project.basedir}/src/main/java13</compileSourceRoot>
+                                        </compileSourceRoots>
+                                        <outputDirectory>${project.build.outputDirectory}/META-INF/versions/13</outputDirectory>
+                                    </configuration>
+                              </execution>
+                          </executions>
+                          <configuration>
+                              <release>9</release>
```
reta: Workaround (for now, no impact on release artifacts) for `tools.jar` (https://bugs.openjdk.java.net/browse/JDK-8199325)

PR #567, opened 2019-07-11
[CXF-8073] Better error reporting on oauth2's OAuthServiceException

rcsilva83:


rcsilva83:
> 
> 
> I don't think the "state" should be returned, it doesnt appear in the spec for example:
> https://tools.ietf.org/html/rfc6749#section-5.2
> 
> Secondly, I don't think this is the correct solution. Instead of adding a toString, instead we could modify the OAuthServiceExceptionMapper to properly map the OAuthServiceException to the correct JSON output.

Hi!

The problem I'm trying to fix is when the client can't authenticate with the Oauth server. No JSON serialization is done in this scenario. So I don't think that modifying the OAuthServiceExceptionMapper would solve this issue.

It the "toString" solution is ok for you, I can rip off the "state" field from it.

PR #569, opened 2019-07-23
CXF-8079 Update to introduce the wsat transaction example

zhfeng:
https://issues.apache.org/jira/browse/CXF-8079

### DIFF
--- a/distribution/src/main/release/samples/ws_transaction/README.md
+++ b/distribution/src/main/release/samples/ws_transaction/README.md
@@ -1,6 +1,10 @@
 CXF Web Service Transaction Demo
 ================================
 This demo shows how to use the JTA->WSAT->JTA bridge in the two differnet web services by using the apache-cxf, spring-boot and narayana.
+**NOTE**:  the following dependencies are licensed under LGPL and thus may have additional restrictions beyond the Apache License.
+
+ - [Narayana](https://github.com/jbosstm/narayana/blob/master/LICENSE)
+ - [Hibernate](https://github.com/hibernate/hibernate-orm/blob/master/lgpl.txt)
 
 Buiding and running the demo
 ----------------------------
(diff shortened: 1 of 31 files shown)


distribution/src/main/release/samples/ws_transaction/README.md
```diff
@@ -0,0 +1,40 @@
...
+================
+It can wrap the local transaction and create a bridge between the JTA and WSAT transaction. 
+From the client side, we use the *JaxWSTxOutboundBridgeHandler* to create a mapping of the JTA and Subordinate WSAT
+also the *EnabledWSTXHandler* to propagate the WSAT transaction in the SOAP message headers.
+
+WSAT->JTA bridge in the server side
+================
+From the server side, we use the *JaxWSSubordinateHeaderContextProcessor* to import the Subordinate WSAT transaction from the outside
+and the *JaxWSHeaderContextProcessor* to resume the WSAT transaction and the *OptionalJaxWSTxInboundBridgeHandler* to create the bridge
+the WSAT and JTA.
```
dkulp: I think we should add a note to the README to mention that this example uses dependencies that are licensed under the LGPL and thus may have additional restrictions beyond the Apache License.    Explicitly listing the transaction manager (and hibernate) as having the potentially problematic license in the READ is likely good enough.    Users can then decide to use or not use the source in this example.

distribution/src/main/release/samples/ws_transaction/client/pom.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
dkulp: The apache license header needs to go in here.

zhfeng:
Thanks @dkulp for reviewing and I will make changes as your suggestion.

distribution/src/main/release/samples/ws_transaction/client/pom.xml
zhfeng: yeah, I added it.

zhfeng:
@dkulp can you take a look at this PR and is there anything I have to do with the changes ?
Thanks !

PR #570, opened 2019-07-29
[CXF-8080] Ensure stages from async methods are completed

andymc12:
This fix ensures that we use the specified executor service to complete the CompletionStage from the MP Rest Client async methods. It also adds a test case based off of Manish Kumar's [reproducer example](https://gist.github.com/manish2aug/b4133b9fbedc2175e15b303a0682b552).

We may a similar change in OpenLiberty to handle this case, but I neglected to contribute that change back to CXF until now (apologies).


andymc12:
Thanks for the review @reta!  There were three failures in the Jenkins build, but I've seen those same failures (intermittently) in other builds that did not contain these changes, so I think it should be safe to merge.

PR #574, opened 2019-08-22
CXF-8096 Adding LoggingFeature causes blocking read

kkrisz1:
https://issues.apache.org/jira/browse/CXF-8096


gnodet:
Closing: superseded by #2957 which provides a clean fix for CXF-8096 using a TeeInputStream approach that avoids blocking reads on streaming responses.

PR #575, opened 2019-08-27
CXF-8094 Register additional classes for WADL

yurkom:
It is useful in case when classes are provided by 3rd party and cannot be annotated by @XmlSeeAlso.


yurkom:
Hi @reta , it works for our case.

PR #577, opened 2019-09-03
cxf-rt-rs-client: use ArrayList in ClientProxyImpl

amarkevich:


rmannibucau:
Doesnt it consumes more memory and impacts apps negatively under load? (Gc)

amarkevich:
1. getPathParamValues
allocate 10 continuous elements by default; its enough for most cases without re-allocations;
list.set is O(1) for ArrayList
2. handleMultipart
allocate exactly expected parameters size

In general ArrayList uses less memory than LinkedList

rmannibucau:
2. Agree
1. ArrayList is not sized as far as i can see - can't be without visiting parameters in current state - so assumption is random, no? A max can be stored building the request and propagated (plain int) or linkedlist sounds saner since we dont control param count here and it can be a lot. Wdyt?

amarkevich:
its complex to count PathParam count because
a) class level annotations [0..nc]
b) method level annotations [0..nm]
c) BeanParam with annotations [0..nb]
Total [0..(nc+nm+nb)]
0 is the best case for LinkedList.
Add operation for LinkedList cause memory allocation

Checked using https://github.com/DimitrisAndreou/memory-measurer
```
    public static void main(String[] arguments) {
        ArrayList<Object> al = new ArrayList<>();
        LinkedList<Object> ll = new LinkedList<>();
        System.out.println("size\tArrayList\tLinkedList");
        for (int i = 0; i < 11; ++i) {
            System.out.println(i + "\t" + MemoryMeasurer.measureBytes(al) + "\t" + MemoryMeasurer.measureBytes(ll));
            al.add(new Object());
            ll.add(new Object());
        }
    }

```
Result:

> size	ArrayList	LinkedList
> 0	40	32
> 1	96	72
> 2	112	112
> 3	128	152
> 4	144	192
> 5	160	232
> 6	176	272
> 7	192	312
> 8	208	352
> 9	224	392
> 10	240	432

and ArrayList structure inclined to cpu cache hit

rmannibucau:
@amarkevich yes, but you only test with <= 11 objects which is not true for the code which can have let say 17 parameters which means 32 cells for the arraylist.

amarkevich:
we're talking about PathParam only; usually its count less than 10
> 17	400	712
`ObjectGraphMeasurer.measure`:
> 17	Footprint{Objects=19, References=23, Primitives=[int x 2]}	Footprint{Objects=35, References=53, Primitives=[int x 2]}

rmannibucau:
I know but ~25 happens, think about bean binding and search endpoints. And more you have params more the diff will be impacting
17 being the worse case i think

amarkevich:
AFAIK search uses QueryParam and handled by `handleQueries` method

rmannibucau:
Ok, ran a comparison on a recent JVM - didn't recheck that since years, I get something close to your test:

    size	array	linked	diff
    0	40	32	-8
    1	136	112	-24
    2	192	192	0
    3	248	272	24
    4	304	352	48
    5	360	432	72
    6	416	512	96
    7	472	592	120
    8	528	672	144
    9	584	752	168
    10	640	832	192
    11	720	912	192
    12	776	992	216
    13	832	1072	240
    14	888	1152	264
    15	944	1232	288
    16	1024	1312	288
    17	1080	1392	312
    18	1136	1472	336
    19	1192	1552	360
    20	1248	1632	384
    21	1304	1712	408
    22	1360	1792	432
    23	1464	1872	408
    24	1520	1952	432
    25	1576	2032	456
    26	1632	2112	480
    27	1688	2192	504
    28	1744	2272	528
    29	1800	2352	552
    30	1856	2432	576
    31	1912	2512	600
    32	1968	2592	624

So long story short, for 0 or 1 parameter, linked list is better, for 2, there is no real different (except linked list would be a very little bit slower and then arraylist is better and moreover more stable in terms of allocations so guess this is actually a very good change, sorry for the false warning.

Side note: here is what I used to generate ^^ if anyone wants to give a try on openj9 or another VM version (used zulu 202):

    import static jdk.nashorn.internal.ir.debug.ObjectSizeCalculator.getObjectSize;
    
    import java.lang.reflect.Field;
    import java.util.ArrayList;
    import java.util.LinkedList;
    import java.util.List;
    
    public class Tmp {
        public static void main(final String[] args) {
            System.setProperty("java.vm.name", "Java HotSpot(TM) tmp"); // to reuse nashorn object size evaluator
    
            final List<Object> linked = new LinkedList<>();
            final List<Object> array = new ArrayList<>();
            System.out.println("size;array;linked");
            csv(0, array, linked);
            for (int i = 0; i < 32; i++) {
                final String value = "test" + i;
                linked.add(value);
                array.add(value);
                csv(i + 1, array, linked);
            }
            if (!array.equals(linked)) {
                throw new IllegalStateException("Something went wrong: \n" + array + "\n" + linked);
            }
        }
    
        private static void csv(final int size, final List<?> array, final List<?> linked) {
            System.out.print(size);
            System.out.print(';');
            System.out.print(getObjectSize(array));
            System.out.print(';');
            System.out.println(getObjectSize(linked));
        }
    }

PR #581, opened 2019-09-28
CXF-8125

bbodnar:
Implementation-proposal for CXF-8125


bbodnar:
Will be recreated in a new fork.

PR #584, opened 2019-10-01
CXF-8128: Fill address field in case of responses

kkrisz1:
Fill address in case of responses to be able to find out which URI does belong to the logged response.


ffang:
Hi @kkrisz1 ,

First thanks for the contribution.

With your change, org.apache.cxf.jaxws.logging.SOAPLoggingTest is broken. Previously it assume LogEvent.getAddress return "null", but with your change the LogEvent.getAddress expects to return something.

Could you please update org.apache.cxf.jaxws.logging.SOAPLoggingTest accordingly also?

Freeman

kkrisz1:
hi @ffang ,

Thank you for you comment.
Test is fixed.

-Krisz

PR #585, opened 2019-10-02
cxf-rt-rs-client: set Accept header to wildcard for void response

amarkevich:


rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/ClientProxyImpl.java
```diff
@@ -477,16 +477,18 @@ protected void checkResponse(Method m, Response r, Message inMessage) throws Thr
         List<MediaType> accepts = getAccept(headers);
         if (accepts == null) {
             List<MediaType> produceTypes = ori.getProduceTypes();
```
coheigea: LGTM but here "produceTypes" could be in the else statement instead.

mlschechter:
@coheigea - Can you please let me know when this will be released? We just implemented a media type whitelisting feature, and the bug this fixes is stopping us from deploying.

Thanks!

coheigea:
Realistically with the holiday season - the second week of January I'd say.

mlschechter:
That's great - thanks!

mlschechter:
Thanks for the release, @coheigea - everything's working great!

PR #588, opened 2019-10-14
CXF-8129 - Update to Woodstox 6.2.x

coheigea:


core/pom.xml
```diff
@@ -189,18 +189,6 @@
             <scope>provided</scope>
             <optional>true</optional>
         </dependency>
-        <dependency>
```
reta: Just curious @coheigea if we should use this dependency vs shaded one. It seems like Woodstox uses this library anyway (https://github.com/FasterXML/woodstox/blob/df8d1648cd8bf5c4e622f00156d62ef3fd88d7c5/pom.xml#L42), may be we could keep it the old way (plus, it would stay optional). What do you think?

reta: I see, thanks @coheigea

dkulp:
With the MSV change... does this mean CXF would no longer work with Woodstox 4/5 or would only the schema validation stuff not work?     I guess that might be my biggest concern.   If they have camel or something loaded first that would pull in a Woodstox 5, do we have a way to to allow it to still work?

coheigea:
I raised an issue in Woodstox about the MSV shading: https://github.com/FasterXML/woodstox/issues/103

PR #589, opened 2019-10-18
CXF-8133 - Disable JAX-RS HTTP method overriding by default

coheigea:
@rmannibucau WDYT


rmannibucau:
@coheigea :+1:  from me

PR #595, opened 2019-11-07
Build behind firewall

cunningt:
I'm having trouble building cxf behind a firewall and would like to add a property so that I can change the URL address to something my proxy will accept


cunningt:
Fixes https://issues.apache.org/jira/browse/CXF-8145

PR #596, opened 2019-11-10
CXF-8149 - Reduce synchronization in AbstractJXBProvider

scsosna99:
Previous version did what I would consider a lot of gratuitous synchronization which can be reduced in a number of ways.

First, use ConcurrentHashMap instead of straight HashMap for managing for tracking the contexts created for classes and packages.  At that point, we only need to do serialization when the collections don't already contain a context for the class or package.

Second, move to a "ConcurrentHashSet" for collectionContextClasses.  Underlying this is a ConcurrentHashMap and is used by ConcurrentHashMap as the KeySet, so we can create one of these for the collectionContextClasses.  At that point, can again reduce synchronizations required.

These changes were made locally at my company to address these synchronizations substantially slowing down some of our web services, and reducing synchronizations substantially improved throughput by - in most cases - reducing synchronizations and letting threads not block each other.


scsosna99:
nevermind, additional testing showed that ConcurrentHashMap causes more blocks than what I had done originally, I still think there's too many synchronizations occurring but need to revisit my solution

PR #597, opened 2019-11-12
CXF-8149 Reduce synchronization in AbstractJXBProvider

amarkevich:
rework #596 
Note for Java 8: ConcurrentHashMap.computeIfAbsent(k,f) locks bin when k present [1]

1. https://bugs.openjdk.java.net/browse/JDK-8161372


amarkevich:
@scsosna99 could you recheck the case using this impl?

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/provider/AbstractJAXBProvider.java
```diff
@@ -506,38 +505,41 @@ public JAXBContext getJAXBContext(Class<?> type, Type genericType) throws JAXBEx
...
-            if (context == null) {
-                Class<?>[] classes;
-                if (extraClass != null) {
-                    classes = new Class[extraClass.length + 1];
-                    classes[0] = type;
-                    System.arraycopy(extraClass, 0, classes, 1, extraClass.length);
-                } else {
-                    classes = new Class[] {type};
-                }
+        final AtomicReference<JAXBException> jaxbException = new AtomicReference<>();
```
reta: I am wondering if there is any particular reason to use `AtomicReference`, the code block seems to be local to the function, no sharing or race conditions should be observed.

amarkevich: `JAXBException extends Exception` -> should be handled inside Function or catched+rethrown in the main code for the same behaviour

amarkevich: `AtomicReference` just a object holder here

scsosna99:
> @scsosna99 could you recheck the case using this impl?

Our biggest contention appears to be in getCollectionContext(), where manipulating the collection blocks continually.  The solution that we've come up with is to use a ConcurrentHashMap and call putIfAbsent, which substantially reduces blockage

 I wanted to find a solution that kept this a Set - and had a better chance of passing your scrutiny - but blocking continues, synchronizing threads, impacting performance.  I thought using newKeySet() would get me what I needed but didn't; also tried ReentrantReadWriteLock to allow n-read/1-write.

Your changes would affect other collections which aren't really impacting me, though I thought changing them to ConcurrentHashMap's would help reduce synchronization.

Chances are that all these changes give some general improvement but our hack works better.  Our one-off change is 2.6.11 running against Java7 (don't chastise me, I just started here!) so the JDK bug referenced above would need more research.  Longer-term goal is moving our services out of the monolith which will end up being Spring Boot and new problems arise....

Thanks all, but I'm going to drop this for now.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/provider/AbstractJAXBProvider.java
reta: Since this is JAXB-related code, using `javax.xml.ws.Holder` would be more lightweight option I think

amarkevich: PR updated

scsosna99:
So I had my team member rerun his performance tests and the original changes I made are much better performing than originally thought.  Looks like @amarkevich has taken the lead on this and doesn't appear I need to restart my pull request.  Is this true?

amarkevich:
@reta any objections?

dkulp:
My main concern is using the javax.xml.ws.Holder object.   That would introduce a direct dependency from the JAX-RS frontend onto the JAX-WS API.   With Java8, that's not a huge deal as its in the JDK, but with Java11+, it is a problem.   My simple suggestion would be to create a private inner "holder" class of some sort just for that.

amarkevich:
@dkulp is initial impl using AtomicReference fit the requirements?

dkulp:
Yep.  Replacing Holder with AtomicReference would work perfectly for this.

reta:
@dkulp That is right, I suggested to remove AtomicReference since this is sychronization primitive, this code snippet does not need sychronization. The small holder object would be simpler and lightweight I thinkm

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/provider/AbstractJAXBProvider.java
```diff
@@ -517,7 +516,7 @@ public JAXBContext getClassContext(Class<?> type) throws JAXBException {
         return getClassContext(type, type);
     }
     protected JAXBContext getClassContext(Class<?> type, Type genericType) throws JAXBException {
-        final Holder<JAXBException> jaxbException = new Holder<>();
+        final JAXBException[] jaxbException = new JAXBException[] {null};
```
reta: Nice trick :+1:

dkulp: I agree.  Nice trick.  👍

PR #598, opened 2019-11-12
Add a property for tarLongFileMode

cunningt:
Add a property for tarLongFileMode so we can build cxf with posix tarLongFileMode


cunningt:
https://issues.apache.org/jira/browse/CXF-8152

PR #603, opened 2019-11-20
[CXF-8161] fix memory leak and thread leak in JMSDestination

steingebein:
Solves problems of https://issues.apache.org/jira/browse/CXF-8161:

- JMSDestination#restartConnection() is synchronized: when multiple poller-threads run into exception handling, the restart will only run sequential. (otherwise there will be race conditions in restartConnection)
- JMSDestination#restartConnection() now checks if the PollingMessageListenerContainer is running before restart. When multiple poller-threads trigger the restart, it will only starts one new PollingMessageListenerContainer.
- PollingMessageListenerContainer stops the Executor after exception handling, resources will be released

### DIFF
--- a/rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSDestination.java
+++ b/rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSDestination.java
@@ -128,14 +128,22 @@ public class JMSDestination extends AbstractMultiplexDestination implements Mess
         Session session = null;
         try {
             ExceptionListener exListener = new ExceptionListener() {
-                public void onException(JMSException exception) {
-                    if (!shutdown) {
+                private boolean restartTriggered;
+
+                public synchronized void onException(JMSException exception) {
+                    if (!shutdown && !restartTriggered) {
                         LOG.log(Level.WARNING, "Exception on JMS connection. Trying to reconnect", exception);
-                        restartConnection();
+                        new Thread(new Runnable() {
+                            @Override
+                            public void run() {
+                                restartConnection();
+                            }
+                        }).start();
+                        restartTriggered = true;
                     }
                 }
             };
-            
+
             PollingMessageListenerContainer container;
             if (!jmsConfig.isOneSessionPerConnection()) {
                 connection = JMSFactory.createConnection(jmsConfig);
@@ -174,7 +182,7 @@ public class JMSDestination extends AbstractMultiplexDestination implements Mess
         }
     }
 
-    protected void restartConnection() {
+    protected synchronized void restartConnection() {
         int tries = 0;
         do {
             tries++;
@@ -215,14 +223,11 @@ public class JMSDestination extends AbstractMultiplexDestination implements Mess
     }
 
 
-
     /**
      * Convert JMS message received by ListenerThread to CXF message and inform incomingObserver that a
      * message was received. The observer will call the service and then send the response CXF message by
      * using the BackChannelConduit
      *
-     * @param message
-     * @throws IOException
      */
     public void onMessage(javax.jms.Message message) {
         ClassLoaderHolder origLoader = null;
--- a/rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/util/PollingMessageListenerContainer.java
+++ b/rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/util/PollingMessageListenerContainer.java
@@ -152,6 +152,7 @@ public class PollingMessageListenerContainer extends AbstractMessageListenerCont
                         safeRollBack();
                     }
                 } catch (Throwable e) {
+                    safeRollBack();
                     handleException(e);
                 }
             }
@@ -201,7 +202,6 @@ public class PollingMessageListenerContainer extends AbstractMessageListenerCont
     }
     
     protected void handleException(Throwable e) {
-        running = false;
         JMSException wrapped;
         if (e  instanceof JMSException) {
             wrapped = (JMSException) e;
@@ -239,10 +239,7 @@ public class PollingMessageListenerContainer extends AbstractMessageListenerCont
 
     @Override
     public void stop() {
-        LOG.fine("Shuttting down " + this.getClass().getSimpleName());
-        if (!running) {
-            return;
-        }
+        LOG.fine("Shutting down " + this.getClass().getSimpleName());
         running = false;
         super.stop();
     }
--- a/rt/transports/jms/src/test/java/org/apache/cxf/transport/jms/JMSDestinationTest.java
+++ b/rt/transports/jms/src/test/java/org/apache/cxf/transport/jms/JMSDestinationTest.java
@@ -31,6 +31,7 @@ import javax.jms.Destination;
 import javax.jms.InvalidClientIDException;
 import javax.jms.JMSException;
 import javax.jms.Queue;
+import javax.jms.Session;
 import javax.jms.Topic;
 
 import org.apache.activemq.util.ServiceStopper;
@@ -44,10 +45,14 @@ import org.apache.cxf.transport.Conduit;
 import org.apache.cxf.transport.MessageObserver;
 import org.apache.cxf.transport.MultiplexDestination;
 import org.apache.cxf.transport.jms.util.ResourceCloser;
+import org.awaitility.Awaitility;
 
 import org.junit.Ignore;
 import org.junit.Test;
 
+import static org.easymock.EasyMock.expect;
+import static org.easymock.EasyMock.niceMock;
+import static org.easymock.EasyMock.replay;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
@@ -57,12 +62,12 @@ public class JMSDestinationTest extends AbstractJMSTester {
     private static final class FaultyConnectionFactory implements ConnectionFactory {
         private final ConnectionFactory delegate;
         private final AtomicInteger latch;
-        
+
         private FaultyConnectionFactory(ConnectionFactory delegate, int faults) {
             this.delegate = delegate;
             this.latch = new AtomicInteger(faults);
         }
-        
+
         @Override
         public Connection createConnection() throws JMSException {
             if (latch.getAndDecrement() == 0) {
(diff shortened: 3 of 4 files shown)


rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/util/PollingMessageListenerContainer.java
```diff
@@ -210,6 +210,8 @@ protected void handleException(Throwable e) {
             wrapped.addSuppressed(e);
         }
         this.exceptionListener.onException(wrapped);
+        // stop the executor and the current thread after handling of the exception.
```
reta: If I understood the idea behind current implementation correctly, we have the problem with `running` flag. The first thing `restartConnection` does is to deactivate the existing `jmsListener` by calling shutdown. Unfortunately, method short-circuits because the `running` flag is unset, the executor is not being shut down properly. We could fix it by introducing `shutdown` flag or just calling `super.stop();`:
```
@Override
    public void stop() {
        LOG.fine("Shuttting down " + this.getClass().getSimpleName());
        if (!running) {
            super.stop();
            return;
        }
        running = false;
        super.stop();
    }
```

Wdyt?

rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSDestination.java
```diff
@@ -174,9 +174,9 @@ public void onException(JMSException exception) {
         }
     }
 
-    protected void restartConnection() {
+    protected synchronized void restartConnection() {
```
reta: To be fair, I am not sure this is the right thing. Just looking over the code, I think we could run into the deadlock pretty easy. For example, `JMSDestination::activate()` executes `restartConnection()` in `unnamed` thread, whereas the `ExceptionListener` will be executed in the context of `PollingMessageListenerContainer` thread pool. If the the exception happens in one of the pollers while the `restartConnection()` is still in progress (I think it is very possible), we would have an issue.

steingebein: I don't think we're getting a deadlock. The first thread calling `restartConnection()` (`unnamed` or poller-thread) will do the restart - other threads will wait/block until `restartConnection()` is finished. The `ExceptionListener` is only called from the poller threads, these will wait until `restartConnection()` is finished.
So all threads get blocked until the restart ist successful. Without a connection the threads cann't do anything useful - I think this is ok when they are blocked.

One Problem can occur: when `restartConnection()` runs in `createTargetDestinationListener()` to  `container.start()` another `ThreadPool` will be created and the threads are blocked - every time until restart is successful. 
I have no idea how to solve this problem, but I think it's not critical because at this point we already have a successful JMS connection. 

BTW:
When you use the flag  jmsConfig.isOneSessionPerConnection = true, the Exception Handling will not work. The `ExceptionHander `is alwas null and `PollingMessageListenerContainer#handleException()` run into a NullPointerException, so I ignored this case.

I think the interaction between `JMSDestination`, `PollingMessageListenerContainer` and the poller-threads is very complicated and fragile. I think for future versions there must be a rewrite of the code.

rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/util/PollingMessageListenerContainer.java
steingebein: Yes. I had two reasons for my solution:
* If you call `stop()` before the Exception handling is finished, the poller thread will be shutdown and the Exception handling will be interrupted immediately. (`AbstractMessageListenerContainer#stop()` calls `shutdownNow()`)
* `JMSDestination#restartConnection()` checks the running-Flag now, so we need to unset it before `ExceptionListener` calls `restartConnection()`

If it's ok for you I change it to 
   ```
 @Override
    public void stop() {
        LOG.fine("Shutting down " + this.getClass().getSimpleName());
        running = false;
        super.stop();
    }
```
and to `this.stop()` without comment in `handleException`?

File: rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSDestination.java (hunk 2 above)
Comment (reta) [reply rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSDestination.java]: Sorry for delayed response @steingebein, I think you are right, no risk of deadlock, just tons of blocked threads, did some tests to verify that, we should be "safe" there. The `synchronized` would indeed help there. 

I was trying to reconstruct the flows, and came to the same conclusions as you: the code is quite complicated and fragile.

File: rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/util/PollingMessageListenerContainer.java (hunk 1 above)
Comment (reta) [reply rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/util/PollingMessageListenerContainer.java]: :+1: , should be a bit simpler overall I think, thanks!

Comment (steingebein) [reply rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/util/PollingMessageListenerContainer.java]: done

File: rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSDestination.java (hunk 2 above)
Comment (steingebein) [reply rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSDestination.java]: I have one suggestion to improve the code:

Move `sychnronized` from `JMSDestination#restartConnection()` to `PollingMessageListenerContainer#handleException()`, because `PollingMessageListenerContainer` starts multiple threads and needs the synchronisation. So `sychnronized` belongs more into this class.

To complete this change, I have to add `!jmsListener.isRunning() &&` in `JMSDestination#activate()`:
```
  if (!jmsListener.isRunning() && !jmsConfig.isOneSessionPerConnection()) {
                // If first connect fails we will try to establish the connection in the background
                new Thread(new Runnable() {
                    @Override
                    public void run() {
                        restartConnection();
                    }
                }).start();
            }
```

But I'm not sure if it makes the code better...

Comment (reta) [reply rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSDestination.java]: I think with the current implementation, it could be an option to consider, but in general, the `ExceptionListener` is a black box for `PollingMessageListenerContainer`, it is actually `JMSDestination` which makes things worst: it tries to reconnect within the callback. Between these two patches, I would personally prefer to stick with your first idea, `synchronized` over `restartConnection`. Thank you.

**Comment by reta** [comment]
Hi @steingebein, sorry for being quiet for a while, I was trying to understand the issue / solution more profoundly while also committing a number of test cases (in master, will try to backport to 3.3.x / 3.2.x shortly). 

I think the solution we have with shutting down the thread pool from one of its threads (inside `onException` callback) kind of works but it does not look right (I think you would agree). On the faulty path, we always stack at `awaitTermination` ... followed by interruption. 

Now, obviously, the question is what the alternative. I think a more reliable option would be to spin off the thread for `restartConnection`:
```
ExceptionListener exListener = new ExceptionListener() {
                public void onException(JMSException exception) {
                    if (!shutdown) {
                        LOG.log(Level.WARNING, "Exception on JMS connection. Trying to reconnect", exception);
                        if (jmsListener == null) {
                            restartConnection();
                        } else {
                            new Thread(new Runnable() {
                                @Override
                                public void run() {
                                    restartConnection();
                                }
                            }).start();
                        }
                    }
                }
            };
```

It is definitely far from ideal, but taking the complexity of the `JMSDestination` & related components, it looks like an acceptable tradeoff. Yes, we may create a number of short-lived threads at peak, but only one would be doing useful work in fact. With this approach, we eliminate self-blocking at a cost of wasting more resources. It also means we should not call `stop` in `onException` but let `JMSDestination` to clean things up (it does inside the `deactivate`).

This is obviously a patch, not a proper solution, the thoughtful redesign would make more sense in the long run. What do you think? Thanks for your patience.

**Comment by steingebein** [comment]
Hi @reta,

I think your idea is better than the current pull request. It moves the fix completly into `JMSDestination` and it will work even if you replace or change `PollingMessageListenerContainer` 

But I think it comes up with one problem: We have to revert my change in `restartConnection()`, which checks `jmsListener.isRunning()` before restarting a connection. Otherwise `restartConnection()` does nothing (`PollingMessageListenerContainer.running` is not unset).
It will cause every short-lived thread from `ExceptionListener` to stop and restart the connection. You get `jmsConfig.getConcurrentConsumers()` reconnects.

Perhaps it's better to restart only once per `PollingMessageListenerContainer`/`ExceptionListener`-instance:
```
ExceptionListener exListener = new ExceptionListener() {
                private boolean restartTriggered = false;
                public synchronized void onException(JMSException exception) {
                    if (!shutdown && !restartTriggered) {
                        LOG.log(Level.WARNING, "Exception on JMS connection. Trying to reconnect", exception);
                        new Thread(new Runnable() {
                             @Override
                             public void run() {
                                 restartConnection();
                             }
                        }).start();
                        restartTriggered = true;
                    }
                }
            };
```
changes:
- removed your if-condition, because `jmsListener == null` is always false when `onException` is called
- `restartTriggered` prevents multiple restarts from the `PollingMessageListenerContainer`-threads
- `synchronized onException` because it's called from multiple threads

reta:
:+1: , sounds great, thank you @steingebein! My appologies, did you push your changes? Thanks!

steingebein:
hi @reta, 
I changed the code and added a test case, please review it.

reta:
hey @steingebein, thanks a lot for the great work, I did two post-merge changes:
 - put back `running = false;`, so the polling threads would stop progressing, I believe this make sense
 - refactored your test case to use test doubles as other tests from `JMSDestinationTest` suite (no mocks)
Please let me know if you have any concerns or I missed something, thank you!

steingebein:
Hi @reta, thank you for review & finish this pull request.

`running = false;` makes the poller threads stopping/failing faster, so I think that's better.

Your test changes: my test case was intended to test multiple poller-threads failing, so they call `handleException` parallel. In your test case the error handling runs not parallel, because there's only one poller thread. If you remove `sychronized` and my changes in `ExceptionListener`, your test will not fail.
I created another branch with following changes to the test case:
- `concurrentConsumer` is set 10
- `FaultyConnection` will fail 5 times (more poller-threads will fail)
- verify that there was only one call to restartConnection() 

See [https://github.com/steingebein/cxf/commit/0dbf6c046b2b321d03474003104beae732c5ebf4]

If you like these changes I can create another pull request.

reta:
@steingebein thanks a lot, the test case looks good, yeah, could you please create a PR for it? Thanks!

steingebein:
@reta pull request created: https://github.com/apache/cxf/pull/614

PR #604, opened 2019-11-21
[CXF-8162] JWE with multiple recipients does not work for AES CBC Encryption

frelibert:
Fixed:
- System.arraycopy src position
- exposed AesCbcContentEncryptionAlgorithm to make cek and iv reusable for all recipients

### DIFF
--- a/rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwe/AesCbcContentEncryptionAlgorithm.java
+++ b/rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwe/AesCbcContentEncryptionAlgorithm.java
@@ -1,0 +1,82 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.cxf.rs.security.jose.jwe;
+
+import java.security.spec.AlgorithmParameterSpec;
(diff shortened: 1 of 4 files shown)


rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwe/AesCbcHmacJweEncryption.java
```diff
@@ -19,66 +19,50 @@
...
     protected byte[] getActualCek(byte[] theCek, String algoJwt) {
         return doGetActualCek(theCek, algoJwt);
     }
+    
     protected static byte[] doGetActualCek(byte[] theCek, String algoJwt) {
-        int size = getFullCekKeySize(algoJwt) / 2;
+        int size = AesCbcContentEncryptionAlgorithm.getFullCekKeySize(algoJwt) / 2;
         byte[] actualCek = new byte[size];
-        System.arraycopy(theCek, size, actualCek, 0, size);
+        System.arraycopy(theCek, 0, actualCek, 0, size);
```
coheigea: I think this change may not be correct. The spec says that for AES_128_CBC_HMAC_SHA_256 the input key must be 32 bytes, and the test in JweJsonProducerTest.testMultipleRecipientsA128CBCHS256GivenCek is passing in a 16 byte key hence the error:

"The input key K is 32 octets long."
https://tools.ietf.org/html/rfc7518#section-5.2.3

frelibert:
Thank you for the feedback.
You were right. The problem with the array indexoutofbound was due to an invalid inputKey.
I reverted that change and I've added a check for proper keysize to have better exceptionhandling.

rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwe/AesCbcContentEncryptionAlgorithm.java
```diff
@@ -0,0 +1,82 @@
...
+
+import javax.crypto.spec.IvParameterSpec;
+
+import org.apache.cxf.rs.security.jose.jwa.AlgorithmUtils;
+import org.apache.cxf.rs.security.jose.jwa.ContentAlgorithm;
+
+public class AesCbcContentEncryptionAlgorithm extends AbstractContentEncryptionAlgorithm {
+    
+    static final Map<String, String> AES_HMAC_MAP;
+    static final Map<String, Integer> AES_CEK_SIZE_MAP;
```
coheigea: I'd prefer to make these private and add getter methods. Either that or make them unmodifiable maps.

frelibert:
Ok, thanks.
I had committed another fix (for the private property issue) but I see you fixed that yourself by now.
I'll remove my forked branch

coheigea:
@frelibert thanks for the PR!

PR #605, opened 2019-11-26
CXF-8163: javax.ws.rs.core.Response#getCookies() fails when cookie has no value

reta:
The RFC-6265 has the clarifications regarding empty cookie values, both in grammar  [1] and interpretation [2]. The approach adopted by other Java libraries (RESTEasy, Jersey, JDK11's HttpClient, ...) is to return an empty string as the value of the cookie.

[1] https://tools.ietf.org/html/rfc6265#section-4.1.1
[2] https://tools.ietf.org/html/rfc6265#section-5.2

@rmannibucau


rmannibucau:
Hi @reta, looks ok to me

PR #608, opened 2019-12-03
CXF-8171 Support multiple data fields in SSE client

aonnikov:


reta:
Thank you for this fix, @aonnikov , indeed the multiple `data` fields is legitimate case (https://www.w3.org/TR/2009/WD-eventsource-20090421/#event-stream-interpretation)

PR #612, opened 2019-12-06
CXF-8177 CXF-8178 ECDH Algorithm Fixes

frelibert:
Fixed:
- [CXF-8177] Support in API for encrypting/decrypting encryptionkey with ECDH Direct Key Agreement
- [CXF-8177] Support in API for encrypting/decrypting encryptionkey with ECDH Key Wrapping
- [CXF-8177] Support in API for encrypting/decrypting content with ECDH and AESCBC
- [CXF-8178] DerivedKey algorithm for ECDH Key Agreement with Key Wrapping according to the RFC 

I added a testsuite I used myself in another project where I compare JWS and JWE of multiple libraries to see whether or not they are compatible with each other using algorithms as defined in RFC 7518 (JWA).

Quite some tests returned errors before my changes. Now they all pass. Some of the tests could be fixed from the clientside perspective using other classes to set key and content encryptionprovider. Others could not be fixed at all. 
I prefer anyway to use one setup for all which makes it a lot easier from the client's perspective. See the testclasses for this.
I updated operations in JweUtils for this purpose as some assumptions were made in that class that are only compatible with a subset of the algorithms that are actually supported by the codebase.
I also added better support for the case that the given JWK object does not define an algorithm (which is optional, according to the spec).


frelibert:
I replayed my changes to split CXF-8177 and CXF-8178 into 2 separate commits and I did some cleanups as well. Should be fine now.

PR #614, opened 2019-12-16
[CXF-8161] test case with parallel error handling in JMSDestination and PollingMessageListenerContainer

steingebein:
Changes in test case from https://github.com/apache/cxf/pull/603


reta:
Thanks @steingebein !

PR #619, opened 2020-01-04
CXF-8188: Inject into ClientHeadersFactory

andymc12:
Implements [CXF-8188](https://issues.apache.org/jira/browse/CXF-8188) - injecting `@Context` and `@Inject` into custom `ClientHeadersFactory` implementations.  

This tests JAX-RS injection.  The MP Rest Client 2.0 TCK will test CDI injection.

### DIFF
--- a/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/CDIFacade.java
+++ b/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/CDIFacade.java
@@ -1,0 +1,39 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.cxf.microprofile.client.cdi;
+
+import java.util.Optional;
+
+public final class CDIFacade {
+
+    private CDIFacade() {
+    }
+
+    public static <T> Optional<T> getInstanceFromCDI(Class<T> clazz) {
+        T t;
+        try {
+            t = CDIUtils.getInstanceFromCDI(clazz);
+        } catch (ExceptionInInitializerError | NoClassDefFoundError | IllegalStateException ex) {
+            // expected if no MP Config implementation is available
+            t = null;
+        }
+        return Optional.ofNullable(t);
+    }
+}
--- a/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/CDIFacade.java
+++ b/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/CDIFacade.java
@@ -56,17 +56,17 @@ public final class CDIFacade {
         }
     }
 
-    public static <T> Optional<T> getInstanceFromCDI(Class<T> clazz, Bus b) {
+    public static <T> Optional<Instance<T>> getInstanceFromCDI(Class<T> clazz, Bus b) {
         return nullableOptional(() -> CDIUtils.getInstanceFromCDI(clazz, b));
     }
 
-    public static <T> Optional<T> getInstanceFromCDI(Class<T> clazz) {
+    public static <T> Optional<Instance<T>> getInstanceFromCDI(Class<T> clazz) {
         return nullableOptional(() -> CDIUtils.getInstanceFromCDI(clazz));
     }
 
     private static <T> Optional<T> nullableOptional(Callable<T> callable) {
         if (!CDI_AVAILABLE) {
-            return Optional.ofNullable(null);
+            return Optional.empty();
         }
 
         T t;
--- a/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/CDIFacade.java
+++ b/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/CDIFacade.java
@@ -20,18 +20,60 @@
 package org.apache.cxf.microprofile.client.cdi;
 
 import java.util.Optional;
+import java.util.concurrent.Callable;
+
+import org.apache.cxf.Bus;
+
 
 public final class CDIFacade {
+
+    private static final boolean CDI_AVAILABLE;
 
     private CDIFacade() {
     }
 
+    static {
+        boolean b;
+        try {
+            Class.forName("javax.enterprise.inject.spi.BeanManager");
+            b = true;
+        } catch (Throwable t) {
+            b = false;
+        }
+        CDI_AVAILABLE = b;
+    }
+
+    public static Optional<Object> getBeanManager(Bus b) {
+        return nullableOptional(() -> CDIUtils.getCurrentBeanManager(b));
+    }
+
+    public static Optional<Object> getBeanManager() {
+        try {
+            return nullableOptional(() -> CDIUtils.getCurrentBeanManager());
+        } catch (Throwable t) {
+            t.printStackTrace();
+            return Optional.ofNullable(null);
+        }
+    }
+
+    public static <T> Optional<T> getInstanceFromCDI(Class<T> clazz, Bus b) {
+        return nullableOptional(() -> CDIUtils.getInstanceFromCDI(clazz, b));
+    }
+
     public static <T> Optional<T> getInstanceFromCDI(Class<T> clazz) {
+        return nullableOptional(() -> CDIUtils.getInstanceFromCDI(clazz));
+    }
+
+    private static <T> Optional<T> nullableOptional(Callable<T> callable) {
+        if (!CDI_AVAILABLE) {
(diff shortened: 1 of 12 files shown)


systests/microprofile/client/jaxrs/src/test/java/org/apache/cxf/systest/microprofile/rest/client/InjectClientHeadersFactory.java
```diff
@@ -0,0 +1,42 @@
...
+public class InjectClientHeadersFactory implements ClientHeadersFactory {
+
+    @Context
+    Request jaxrsRequest;
+
+    @Override
+    public MultivaluedMap<String, String> update(MultivaluedMap<String, String> incomingHeaders,
+                                                 MultivaluedMap<String, String> clientOutgoingHeaders) {
+
+        System.out.println("InjectClientHeadersFactory update");
```
reta: Debug leftovers?  ;-)

reta:
@andymc12 please correct me if I am wrong, if we consider CDI-based deployments, the instances of the `MicroProfileClientProxyImpl` (and others) should be created as CDI beans, right?

systests/microprofile/client/jaxrs/src/test/java/org/apache/cxf/systest/microprofile/rest/client/InjectClientHeadersFactory.java
andymc12: Good catch - I'll remove that.  Thanks!

andymc12:
> if we consider CDI-based deployments, the instances of the MicroProfileClientProxyImpl (and others) should be created as CDI beans, right?

@reta yes, it is done indirectly by creating the client proxy instance using the `CxfTypeSafeClientBuilder` from the `RestClientBean` / `RestClientExtension`.

reta:
Excellent, thanks @andymc12 , so here is my follow up question / concern: why won't we make `CxfTypeSafeClientBuilder` CDI-aware (using extension or alike) so the proxies are going to be managed by `BeanManager` and as such, could use `@Inject` annotation directly? In this case, the integration should work in all scenarios (managed, standalone, ...). Does it make sense?

andymc12:
@reta I think CDI is already managing the proxy instances.  The extension manages the creation of the proxy (it has to provide additional config via the MP annotations, like `@RegisterProvider`, `@RegisterClientHeaders`, etc.), but once it is created, then it turns the proxy over to CDI to manage.

The problem with the ClientHeadersFactory instance is that it is associated with the client interface proxy by annotation, not injected.  The other problem is that we want the ClientHeadersFactory to be injected with any JAX-RS contextual info too.  

Does that makes sense?  Thanks again for your review!

reta:
hey @andymc12, sorry for the delay, finally found the time to finish the review:
 - `CDIUtils` and `CDIFacade`: we already use reflection-base `BeanManager` detection in `CDIInterceptorWrapper::createWrapper`, it think we should stick to one way to do it
 - may be `MicroProfileClientFactoryBean` is a better place for dealing with CDI & `ClientHeadersFactory`, which could be passed to `MicroProfileClientProxyImpl` in a similar way as `CDIInterceptorWrapper`. The `BeanManager` detection logic (whichever we pick), could be reused for both of them.

What do you think?

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/CDIUtils.java
```diff
@@ -0,0 +1,56 @@
...
+        } catch (NoSuchElementException ex) {
+            // expected if ClientHeadersFactory is not managed by CDI
+            t = null;
+        }
+        return t;
+    }
+
+    @SuppressWarnings("unchecked")
+    private static <T> T findBean(Class<T> clazz) {
+        BeanManager beanManager = CDI.current().getBeanManager();
```
rmannibucau: this is a slow operation, it should likely not be done at runtime but the BM should be bound (at least boundable) to CXF bus as an extension probably

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/CDIUtils.java
```diff
@@ -0,0 +1,56 @@
...
+        Bean<?> bean = beanManager.getBeans(clazz).iterator().next();
+        CreationalContext<?> ctx = beanManager.createCreationalContext(bean);
```
rmannibucau: if !beanManager.isNormalScope(bean.getScope()) then ctx must be released to not leak

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/CDIUtils.java
andymc12: in some cases this is called before the bus is created, but when the bus is available, my latest change will use the BeanManager associated with the bus before calling `CDI.current().getBeanManager()`.

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/CDIUtils.java
andymc12: @rmannibucau That makes sense.  When does the ctx need to be released?  Can I release it as soon as I have a reference to the bean, or do I need to wait until it is out of scope?  Thanks for the review!

andymc12:
@reta
> CDIUtils and CDIFacade: we already use reflection-base BeanManager detection in CDIInterceptorWrapper::createWrapper, it think we should stick to one way to do it

Agreed.  I updated the CDIInterceptorWrapper to use CDIFacade/CDIUtils.

> may be MicroProfileClientFactoryBean is a better place for dealing with CDI & ClientHeadersFactory, which could be passed to MicroProfileClientProxyImpl in a similar way as CDIInterceptorWrapper. The BeanManager detection logic (whichever we pick), could be reused for both of them.

My latest changes reuse the BeanManager detection logic.  I'm not sure if moving the ClientHeadersFactory logic into MicroProfileClientFactoryBean will work since that is when the client proxy is being created.  It's possible the the rest client instance is created in one JAX-RS context (or outside of a JAX-RS request entirely) and then used in a different JAX-RS context.  So then the injection would need to occur on each request, right?

Thanks again!

reta:
Thanks @andymc12 , regarding

> My latest changes reuse the BeanManager detection logic. I'm not sure if moving the ClientHeadersFactory logic into MicroProfileClientFactoryBean will work since that is when the client proxy is being created. It's possible the the rest client instance is created in one JAX-RS context (or outside of a JAX-RS request entirely) and then used in a different JAX-RS context. So then the injection would need to occur on each request, right?

you are right, it might happen, but probably the same would apply to `CDIInterceptorWrapper`: it will pick up the interceptors from the different context. I don't know what is the best option to be fair, just an opinion if we have a single place to make the decisions (the factory bean), it would be much easier to reason about the outcome (proxy) and maintain it. The uncertainty with contexts may cause issues but I would leave the decision to you, I think you have a better view of the picture.

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/CDIUtils.java
rmannibucau: It will trigger dependent predestroys so out of scope I think.

andymc12: I added a new commit that will release the contexts (for all scopes) when the client instance is closed.  Can you take one more peek at my changes?  If all looks good, I think we can merge.  Thanks again!

rmannibucau: Works since there is one factory per bus
Side note being: why cdi utility does not return an Instance{value,runnable} and you use the runnable as a close callback to call the release method. Main difference is then the storage can be an instance list (not set ;)) in the factory instead of a static method which would leak if anything happens wrong in the app lifecycle.

andymc12: I like it!  I've just made those changes.  One last (hopefully) look?  Thanks again!

rmannibucau: Second param should be sthg like:

     beanManager.isNormalScope(bean.getScope()) ? () -> {} : context::release

Normal scope lifecycle is already handled by the container.

Otherwise perfect for me.

andymc12: Shouldn't it be the other way around?   We should call release if it _is_ a normal scope, right?

rmannibucau: If it is not. If it is, it is implicitly called, when not (kind of caller managed) it is your responsability.

andymc12: Those changes are in - as well as updated unit tests.  Thanks again!

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/CDIFacade.java
```diff
@@ -0,0 +1,81 @@
...
+        return nullableOptional(() -> CDIUtils.getInstanceFromCDI(clazz, b));
+    }
+
+    public static <T> Optional<Instance<T>> getInstanceFromCDI(Class<T> clazz) {
+        return nullableOptional(() -> CDIUtils.getInstanceFromCDI(clazz));
+    }
+
+    private static <T> Optional<T> nullableOptional(Callable<T> callable) {
+        if (!CDI_AVAILABLE) {
+            return Optional.ofNullable(null);
```
reta: Minor, `Optional.empty()` would be the thing you are looking for :)

PR #620, opened 2020-01-06
CXF-7996: TCK: Resolves FormParam-related failures

andymc12:
Should fix:
allParamsInParamTest_from_standalone
allParamsOnFieldTest_from_standalone
formParamInParamTest_from_standalone
formParamOnFieldTest_from_standalone

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/JAXRSUtils.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/JAXRSUtils.java
@@ -19,6 +19,8 @@
 
 package org.apache.cxf.jaxrs.utils;
 
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -89,6 +91,7 @@ import org.apache.cxf.common.util.PropertyUtils;
 import org.apache.cxf.common.util.ReflectionUtil;
 import org.apache.cxf.common.util.StringUtils;
 import org.apache.cxf.helpers.DOMUtils;
+import org.apache.cxf.helpers.IOUtils;
 import org.apache.cxf.interceptor.Fault;
 import org.apache.cxf.io.CacheAndWriteOutputStream;
 import org.apache.cxf.io.ReaderInputStream;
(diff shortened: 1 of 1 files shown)


andymc12:
I see two Jenkins failures in different JMS buckets - I don't think these are related to this change though, as I don't think JAX-RS is even involved in these buckets.

reta:
@andymc12 the JMS test failures should not be related, one of them is flaky for sure, second one became flaky recently, I know why, will address it, thanks.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/JAXRSUtils.java
```diff
@@ -1878,4 +1891,16 @@ public static String logMessageHandlerProblem(String name, Class<?> cls, MediaTy
...
+    // copy the input stream so that it is not inadvertently closed
+    private static InputStream copyAndGetEntityStream(Message m) {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        try {
+            IOUtils.copy(m.getContent(InputStream.class), baos);
+        } catch (IOException e) {
+            throw ExceptionUtils.toInternalServerErrorException(e, null);
+        }
+        final byte[] copiedBytes = baos.toByteArray();
+        m.setContent(InputStream.class, new ByteArrayInputStream(copiedBytes));
```
reta: I am wondering if we have to close the source (origin) `InputStream` here, since we replace it (dangling?) with the copy , what do you think @andymc12 ?

andymc12: Good question.  I'm running that suggestion through our automated tests and the TCK - on the surface, I think it should be fine, but just want to be sure.  The change I added looks like this:
```
InputStream origInputStream = m.getContent(InputStream.class);
        try {
            IOUtils.copy(origInputStream, baos);
        } catch (IOException e) {
            throw ExceptionUtils.toInternalServerErrorException(e, null);
        } finally {
            try {
                origInputStream.close();
            } catch (Throwable t) { /* AutoFFDC */ }
        }
```

I'm out of the office tomorrow, so I may not have the results until Monday.

reta: :+1: thanks a lot!

andymc12: I see Dan already merged it (thanks Dan!), but just to followup, the change had no ill effects in our automated testing - the normal functional tests and TCK all passed.

reta: Yep, I also rerun TCK (https://issues.apache.org/jira/browse/CXF-7996) and these failure are gone, thanks @andymc12 !

PR #622, opened 2020-01-08
subclasses of declared exception should be thrown

flaviodonze:
if a service throws "ServerExecption" but ends up throwing a sub-class
of ServerException e.g. ValidationException, the ResponseExceptionMapper
resulting Exception is lost.

Additional information, Resteasy also does it this way:
https://github.com/resteasy/Resteasy/blob/cc424c27f0e9b5ccbc712fd114b7c9abb43e6ba5/resteasy-client-microprofile/src/main/java/org/jboss/resteasy/microprofile/client/ExceptionMapping.java


coheigea:
@andymc12 Should this be merged to 3.3.x-fixes as well?

andymc12:
@coheigea yes, I'll back port that now.  Thanks!

PR #623, opened 2020-01-11
CXF-8159: Update to Swagger 2.1.1 / 1.6.0

deki:


deki:
@reta am I missing something? my tests look good so far

PR #624, opened 2020-01-13
getBytes uses the platform charset by default.

NaridaL:
As far as I can see, the only place where the schemas are loaded into the cache is loadSchemasIntoCache, where they are extracted from an XML Document. As such, they won't have an associated encoding. The XMLSource defaults to UTF-8, so this blows up if you have special characters and, for example, windows-1252 as the platform charset.


rt/rs/description/src/main/java/org/apache/cxf/jaxrs/model/wadl/WadlGenerator.java
```diff
@@ -32,6 +32,7 @@
 import java.lang.reflect.TypeVariable;
 import java.net.URI;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
```
coheigea: Please remove the ByteArrayOutputStream import which is no longer required, to fix a checkstyle problem.

NaridaL: Done.

PR #626, opened 2020-01-19
CXF-7949: Jakarta Servlet API

deki:


amarkevich:
@deki local build passed using java version "1.8.0_241"

deki:
thanks for adapting the test @amarkevich

PR #630, opened 2020-01-24
CXF-8201 add additional check for file scheme in setUriParts

jgallimore:
This additional check enables path appends for file URIs to work as it has done in previous versions of CXF, for example, with this patch:

UriBuilder.fromUri("file:///~/calendar").path("extra").build() will return file:///~calendar/extra as opposed to file:///~/calendar.

This has been discussed on the mailing list here: https://mail-archives.apache.org/mod_mbox/cxf-dev/202001.mbox/%3cCAGRgoZhuKXoYW_JtxyDH=u7Qqqcbeo6-9BXa0dxB8L+iGB4J4g@mail.gmail.com%3e

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/UriBuilderImpl.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/UriBuilderImpl.java
@@ -39,6 +39,7 @@ import javax.ws.rs.core.PathSegment;
 import javax.ws.rs.core.UriBuilder;
 import javax.ws.rs.core.UriBuilderException;
 
+import org.apache.cxf.common.util.CollectionUtils;
 import org.apache.cxf.common.util.PropertyUtils;
 import org.apache.cxf.common.util.StringUtils;
 import org.apache.cxf.jaxrs.model.URITemplate;
@@ -651,9 +652,10 @@ public class UriBuilderImpl extends UriBuilder implements Cloneable {
         } else {
             schemeSpecificPart = uri.getSchemeSpecificPart();
         }
-        if (scheme != null && host == null && port == -1
-                && queryIsEmpty() && !sspMatchessPath(uri) && userInfo == null
-                && uri.getSchemeSpecificPart() != null) {
+        if (scheme != null && host == null && port == -1 && userInfo == null
+                && CollectionUtils.isEmpty(query)
+                && uri.getSchemeSpecificPart() != null
+                && !schemeSpecificPartMatchesUriPath(uri)) {
             schemeSpecificPart = uri.getSchemeSpecificPart();
         }
         String theFragment = uri.getFragment();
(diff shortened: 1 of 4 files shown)


jgallimore:
retest this please

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/UriBuilderImpl.java
```diff
@@ -651,8 +651,8 @@ private void setUriParts(URI uri) {
         } else {
             schemeSpecificPart = uri.getSchemeSpecificPart();
         }
-        if (scheme != null && host == null && (query == null || query.isEmpty()) && userInfo == null
-            && uri.getSchemeSpecificPart() != null) {
+        if (scheme != null && !"file".equalsIgnoreCase(scheme) && host == null && (query == null || query.isEmpty())
```
reta: Hi @jgallimore, thanks for the PR. I think CXF should not assume any intrinsic knowledge about the URI scheme and handle it in a specific way. For example,  what about handling `jar:file`? Or any other custom schemes?  

Although the reply on spec list (https://www.eclipse.org/lists/jaxrs-dev/msg00774.html) explicitly mentions that JAX-RS context is REST-over-HTTP (the CXF case), we could certainly make `UriBuilder` more flexible if the general solution could be found. 

Hope it makes sense, thank you.

jgallimore:
@reta Thanks for the feedback. The conversation continues on the spec list - there's a note about UriBuilder referencing rfc-3986 - https://tools.ietf.org/html/rfc3986. I've pushed a further commit to try and make this more generic by checking the path and raw scheme specific element.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/UriBuilderImpl.java
```diff
@@ -661,6 +662,14 @@ private void setUriParts(URI uri) {
         }
     }
 
+    private boolean queryIsEmpty() {
+        return query == null || query.isEmpty();
```
reta: We have a helper method for it,  `CollectionUtils.isEmpty(query)` from `org.apache.cxf.common.util`

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/UriBuilderImpl.java
```diff
@@ -661,6 +662,14 @@ private void setUriParts(URI uri) {
...
+    }
+
+    private boolean sspMatchessPath(final URI uri) {
+        return uri.getRawSchemeSpecificPart() != null && uri.getRawSchemeSpecificPart().equals("//" + uri.getPath());
```
reta: It would be good to check if `uri.getPath()` is not null, otherwise `"//" + uri.getPath()` becomes something like `//null`

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/UriBuilderImpl.java
```diff
@@ -661,6 +662,14 @@ private void setUriParts(URI uri) {
         }
     }
 
+    private boolean queryIsEmpty() {
+        return query == null || query.isEmpty();
+    }
+
+    private boolean sspMatchessPath(final URI uri) {
```
reta: `Matchess` is misspelled, `schemeSpecificPartMatchesUriPath` would be probably better name (despite being long).

reta:
@jgallimore thanks, I think we can get it in like that, a few comments please to address, thank you

jgallimore:
@reta Thank you again for the feedback. I've hopefully addressed those items!

reta:
@jgallimore thank you!

PR #631, opened 2020-01-24
fix OSGi test after Karaf upgrade

amarkevich:


amarkevich:
its Hudson specific failure - I'm unable to reproduce locally and try to sync failed tests with passing org.apache.cxf.osgi.itests

PR #632, opened 2020-01-25
avoid checked exception in @PostConstruct method

amarkevich:


rt/transports/http-jetty/src/main/java/org/apache/cxf/transport/http_jetty/JettyHTTPServerEngine.java
```diff
@@ -994,27 +993,19 @@ protected void retrieveListenerFactory() {
...
-                int cp = ((ServerConnector)connector).getPort();
-                if (port != cp) {
-                    throw new IOException("Error: Connector port " + cp + " does not match"
-                                + " with the server engine port " + port);
-                }
+    private void checkConnectorPort() {
+        if (null != connector) {
+            int cp = ((ServerConnector)connector).getPort();
+            if (port != cp) {
+                throw new IllegalStateException("Error: Connector port " + cp + " does not match"
```
reta: May be `UncheckedIOException` would be better replacement for `IOException`?

PR #633, opened 2020-01-28
[CXF-8202] Update to OSGi Core 6.0.0

amarkevich:


parent/pom.xml
```diff
@@ -247,7 +247,8 @@
         <cxf.oauth.bundle.version>20100527_1</cxf.oauth.bundle.version>
         <cxf.oro.bundle.version>2.0.8_6</cxf.oro.bundle.version>
         <cxf.osgi.saaj.version>[1.4,2)</cxf.osgi.saaj.version>
-        <cxf.osgi.version>5.0.0</cxf.osgi.version>
+        <cxf.osgi.version>6.0.0</cxf.osgi.version>
+        <cxf.osgi.compendium.version>5.0.0</cxf.osgi.compendium.version>
```
jbonofre: Why not upgrading to 6.0.0 as well using the updated coordinates ?

parent/pom.xml
```diff
@@ -1320,7 +1321,7 @@
             <dependency>
                 <groupId>org.osgi</groupId>
                 <artifactId>org.osgi.compendium</artifactId>
```
jbonofre: We should better use the updated coordinates (osgi.cmpn).

amarkevich: done

parent/pom.xml
amarkevich: changed to artifact osgi.cmpn

PR #635, opened 2020-01-29
[CXF-8206] Update to saaj-impl 1.5.1

amarkevich:


reta:
@amarkevich could you please sync up with @ffang on that, we wanted to switch to saaj-impl 1.5.1 as part of https://issues.apache.org/jira/browse/CXF-7910 (specifically, https://fisheye.apache.org/changelog/cxf?cs=49d34cf3a1583932779d39766b055bf49359da53) but he expressed the concerns  with this specific version, so we decided to stick to 1.3.x (for impl), thanks

amarkevich:
ok

PR #637, opened 2020-02-06
Cxf 8208 - Handle exceptions when looking up swagger ui resources

natechadwick:
CXF-8208 add exception handler so that failed swagger ui resource lookups don't fail the request completely with invalid uri errors.

### DIFF
--- a/rt/rs/description-swagger-ui/src/main/java/org/apache/cxf/jaxrs/swagger/ui/SwaggerUiResourceLocator.java
+++ b/rt/rs/description-swagger-ui/src/main/java/org/apache/cxf/jaxrs/swagger/ui/SwaggerUiResourceLocator.java
@@ -50,8 +50,14 @@ public class SwaggerUiResourceLocator {
         if (resourcePath.startsWith("/")) {
             resourcePath = resourcePath.substring(1);
         }
+        URL ret;
 
-        return URI.create(swaggerUiRoot + resourcePath).toURL();
+        try {
+            ret = URI.create(swaggerUiRoot + resourcePath).toURL();
+        } catch (Exception ex) {
+            throw new MalformedURLException(ex.getMessage());
+        }
+        return ret;
     }
 
     /**
--- a/rt/rs/description-swagger-ui/src/main/java/org/apache/cxf/jaxrs/swagger/ui/SwaggerUiResourceLocator.java
+++ b/rt/rs/description-swagger-ui/src/main/java/org/apache/cxf/jaxrs/swagger/ui/SwaggerUiResourceLocator.java
@@ -20,6 +20,7 @@
 package org.apache.cxf.jaxrs.swagger.ui;
 
 import java.io.IOException;
+import java.lang.IllegalArgumentException;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
@@ -54,7 +55,7 @@ public class SwaggerUiResourceLocator {
 
         try {
             ret = URI.create(swaggerUiRoot + resourcePath).toURL();
-        } catch (Exception ex) {
+        } catch (IllegalArgumentException ex) {
             throw new MalformedURLException(ex.getMessage());
         }
         return ret;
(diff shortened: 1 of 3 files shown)


rt/rs/description-swagger-ui/src/test/java/org/apache/cxf/jaxrs/swagger/ui/SwaggerUIResourceLocatorTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.net.MalformedURLException;
+
+import org.junit.Test;
+
+
+
+public class SwaggerUIResourceLocatorTest {
+
+    @Test
```
amarkevich: use @Test(expected = MalformedURLException .class)  instead of inner try-catch

rt/rs/description-swagger-ui/src/main/java/org/apache/cxf/jaxrs/swagger/ui/SwaggerUiResourceLocator.java
```diff
@@ -50,8 +50,14 @@ public URL locate(String resourcePath) throws MalformedURLException {
         if (resourcePath.startsWith("/")) {
             resourcePath = resourcePath.substring(1);
         }
+        URL ret;
 
-        return URI.create(swaggerUiRoot + resourcePath).toURL();
+        try {
+            ret = URI.create(swaggerUiRoot + resourcePath).toURL();
+        } catch (Exception ex) {
```
reta: The `URI.create` may thrown only `IllegalArgumentException` exception, could you please use it instead of generic `Exception`? Thank you.

natechadwick: I can do that, but the way that the code is currently implemented, the only caller is looking for a MalformedURLException if there is any problem in this method, and the request chain gets nuked if any other exception is thrown which is kinda catastrophic.  Was thinking that any exception from this method should raise the MalformedURLException so the caller just returns false. Future proofing.  If you still want the specific exception caught I can edit and resubmit.

rt/rs/description-swagger-ui/src/test/java/org/apache/cxf/jaxrs/swagger/ui/SwaggerUIResourceLocatorTest.java
natechadwick: got it

rt/rs/description-swagger-ui/src/main/java/org/apache/cxf/jaxrs/swagger/ui/SwaggerUiResourceLocator.java
reta: Thanks for the response, it is fine to raise `MalformedURLException` as you did, the `URI.create` does not do anything beside parsing and throwing `IllegalArgumentException`:
```
        try {
            return URI.create(swaggerUiRoot + resourcePath).toURL();
        } catch (IllegalArgumentException ex) {
            throw new MalformedURLException(ex.getMessage(), ex);
        }
```
Thank you.

reta:
@natechadwick thank you, LGTM

rt/rs/description-swagger-ui/src/main/java/org/apache/cxf/jaxrs/swagger/ui/SwaggerUiResourceLocator.java
amarkevich: To avoid additional knowledge about internal RuntimeException rising a regular constructor with checked URISyntaxException can be used:
```
        try {
            return new URI(swaggerUiRoot + resourcePath).toURL();
        } catch (URISyntaxException ex) {
            throw new MalformedURLException(ex.getMessage());
        }
```

PR #638, opened 2020-02-06
upgrade guava, CVE-2018-10237

davidkarlsen:
Upgrade guava to avoid being marked for CVE-2018-10237 - although CXF might not directly be affected by the issue.
Signed-off-by: David Karlsen <david@davidkarlsen.com>

---
This change is [Reviewable](https://reviewable.io/reviews/apache/cxf/638)

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -119,7 +119,7 @@
         <cxf.geronimo.transaction.version>3.1.4</cxf.geronimo.transaction.version>
         <cxf.geronimo.openapi.version>1.0.12</cxf.geronimo.openapi.version>
         <cxf.glassfish.json.version>1.0.4</cxf.glassfish.json.version>
-        <cxf.guava.version>20.0</cxf.guava.version>
+        <cxf.guava.version>28.2-jre</cxf.guava.version>
         <cxf.hamcrest.version>1.3</cxf.hamcrest.version>
         <cxf.hazelcast.version>3.12.3</cxf.hazelcast.version>
         <cxf.hibernate.em.version>5.4.8.Final</cxf.hibernate.em.version>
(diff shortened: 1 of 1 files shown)


parent/pom.xml
```diff
@@ -119,7 +119,7 @@
         <cxf.geronimo.transaction.version>3.1.4</cxf.geronimo.transaction.version>
         <cxf.geronimo.openapi.version>1.0.12</cxf.geronimo.openapi.version>
         <cxf.glassfish.json.version>1.0.4</cxf.glassfish.json.version>
-        <cxf.guava.version>20.0</cxf.guava.version>
+        <cxf.guava.version>28.2-jre</cxf.guava.version>
```
reta: @davidkarlsen Unfortunately `guava` is problematic to update :( The OSGi bundles are failing to deploy with Guava > 20.0, we need to update the affected dependencies before that:
```
 [caused by: Unable to resolve org.apache.servicemix.bundles.opensaml/3.4.3.1: missing requirement [org.apache.servicemix.bundles.opensaml/3.4.3.1] osgi.wiring.package; filter:="(&(osgi.wiring.package=com.google.common.base)(version>=19.0.0)(!(version>=21.0.0)))"
```
:(

coheigea: @jbonofre - Could we have new opensaml bundles that allow a wider guava range?

coheigea: Also, note that for WSS4J we need to add the following to the feature: install -s mvn:com.google.guava/failureaccess/1.0.1

reta: :+1: thanks @coheigea

andrei-ivanov: Maybe this should be updated directly to 29.0?

coheigea: +1

PR #641, opened 2020-02-17
Prevent RMI registry rebind

jgallimore:
Similar to https://github.com/apache/cxf/pull/640, but also allow different port numbers for the server and registry, if desired.


rt/management/src/main/java/org/apache/cxf/management/jmx/MBServerConnectorFactory.java
```diff
@@ -169,18 +174,71 @@ public void run() {
...
             LOG.info("JMX connector server stopped: " + connectorServer);
         }
     }
 
+    /*
+     * Better to use the internal API than re-invent the wheel.
+     */
+    @SuppressWarnings("restriction")
+    private class JmxRegistry extends sun.rmi.registry.RegistryImpl {
+        static final String LOOKUP_NAME = "jmxrmi";
```
coheigea: I'm just wondering what happens if we define a URL that doesn't end with jmxrmi....shouldn't we take the lookup name from the serviceUrl?

jgallimore: Good point. Let's see if we can do that in a smarter way.

jgallimore: Hopefully done in the latest commit.

coheigea: Looks good thanks

coheigea:
@jgallimore I am happy to merge this if you can't see a way around the sun dependency. Could you also raise a PR with ActiveMQ to port the change you made to pick up the look up name from the connectorPath variable in ManagementContext.

jgallimore:
I can't see a way around the sun dependency at the moment. I've done the PR for ActiveMQ: https://github.com/apache/activemq/pull/465

rt/management/src/main/java/org/apache/cxf/management/jmx/MBServerConnectorFactory.java
```diff
@@ -169,18 +176,85 @@ public void run() {
...
+            portEnd = indexNotOfNumber(serverPart, portStart);
+            if (portEnd > portStart) {
+                final String portString = serverPart.substring(portStart, portEnd);
+                port = Integer.parseInt(portString);
+            }
+        }
+        return port;
+    }
+
+    public static String getBindingName(final JMXServiceURL jmxServiceURL) {
```
coheigea: Could we make this protected? And add a test for it?

jgallimore:
Sorry that was a mistake, I'll revert that.

PR #642, opened 2020-02-18
[CXF-8213] Add Micrometer metric support for JAX-WS

shark300:


integration/spring-boot/autoconfigure/pom.xml
```diff
@@ -74,6 +74,16 @@
             <artifactId>spring-boot-autoconfigure-processor</artifactId>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
```
reta: I think this dependency is not need, or there are reasons to include it?

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/MetricsProperties.java
```diff
@@ -0,0 +1,86 @@
...
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.cxf.spring.boot.autoconfigure;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties("management.metrics")
```
reta: We should probably better use `cxf.management.metrics` (or just `cxf.metrics`) in order to not conflict / confuse with Spring Boot Actuator bundled properties.

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/MicrometerMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,121 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.annotation.Order;
+import org.springframework.web.servlet.DispatcherServlet;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.config.MeterFilter;
+
+@Configuration
+@AutoConfigureAfter({MetricsAutoConfiguration.class, SimpleMetricsExportAutoConfiguration.class})
```
reta: `@AutoConfigureAfter(MetricsAutoConfiguration.class)` should be sufficient, no?

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/MicrometerMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,121 @@
...
+@ConditionalOnWebApplication(type = ConditionalOnWebApplication.Type.SERVLET)
+@ConditionalOnClass({DispatcherServlet.class, MetricsFeature.class})
```
reta: So we need basically 2 conditions here:
1) The metrics module is added (`MetricsProvider.class` should be present)
2) The JAX-WS frontend module is added (may me `JaxWsServerFactoryBean.class` should be present)
What do you think?

integration/spring-boot/autoconfigure/pom.xml
shark300: I'll check it, but probably yes, good catch!

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/MetricsProperties.java
shark300: I see your pont, but as I see, all of Spring Boot related metrics are under management.metrics too. 
https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html#production-ready-metrics-meter
As a developer, I'll search it here. Otherwise, someone should update the documentation too (I don't have a write access to this one).

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/MicrometerMetricsAutoConfiguration.java
shark300: As I see, SimpleMetricsExportAutoConfiguration provides a SimpleMeterRegistry bean, and those extends MeterRegistry and it used by MetricsProvider too.

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/MicrometerMetricsAutoConfiguration.java
shark300: It's ok for me. I'll try it.

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/MicrometerMetricsAutoConfiguration.java
reta: May be you could just have `@ConditionalOnBean(SimpleMeterRegistry.class)` instead?

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/MetricsProperties.java
reta: Fair enough, it is better to follow the Spring's recommendation, https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html#per-meter-properties, so fe we could have metrics under `management.metrics.cxf` and able to control/configure all or part of them, does it make sense?

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/MicrometerMetricsAutoConfiguration.java
shark300: There is a little bit magic inside the Spring Boot: https://github.com/spring-projects/spring-boot/issues/11977#issuecomment-364498154 Due to this, MicrometerMetricsAutoConfiguration should configure after SimpleMetricsExportAutoConfiguration.

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/MetricsProperties.java
shark300: `management.metrics.cxf` is OK for me.

wilkinsona: There are some [guidelines about this](https://docs.spring.io/spring-boot/docs/2.2.5.RELEASE/reference/htmlsingle/#boot-features-custom-starter-configuration-keys) in Spring Boot's documentation. As per those guidelines, please don't use `management.*` for third-party properties. +1 for properties that begin `cfx.`.

shark300: Hi @wilkinsona I overlooked this important thing, thanks. I'll moving that.

junhuhdev:
How would adding support to JAX-RS look like? Could we reuse parts of the code?

shark300:
> How would adding support to JAX-RS look like? Could we reuse parts of the code?
In theoretically yes you can. But I have a lack of knowledge about JAX-RS. Due to this I don't want to add any code related to that.

reta:
@junhuhdev The idea is to shape this PR so the foundation is ready for JAX-RS (very likely without introducing JAX-RS support directly as per discussion with @shark300 on the ticket). I should be back on it in a few days to finish up the review.

junhuhdev:
Ok thanks. Looking forward to testing this implementation. 

Btw is support JAX-RS on the roadmap?

reta:
@junhuhdev certainly, https://issues.apache.org/jira/browse/CXF-8252

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/provider/jaxws/SpringBasedTimedAnnotationProvider.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.cxf.metrics.micrometer.provider.TimedAnnotationProvider;
+import org.apache.cxf.service.Service;
+import org.apache.cxf.service.invoker.MethodDispatcher;
+import org.apache.cxf.service.model.BindingOperationInfo;
+import org.springframework.core.annotation.MergedAnnotationCollectors;
+import org.springframework.core.annotation.MergedAnnotations;
+
+import io.micrometer.core.annotation.Timed;
+
+public class SpringBasedTimedAnnotationProvider implements TimedAnnotationProvider {
```
reta, line 38: I think we could strip `jaxws` package, this provider does not really JAX-WS related, only depends on `cxf-core`, right?

integration/spring-boot/autoconfigure/src/test/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/provider/jaxws/CustomTimed.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.springframework.core.annotation.AliasFor;
+
+import io.micrometer.core.annotation.Timed;
+
+@Target({ElementType.ANNOTATION_TYPE, ElementType.TYPE, ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
+@Timed
+public @interface CustomTimed {
```
reta, line 35: What is the purpose of this annotation?

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/provider/jaxws/SpringBasedTimedAnnotationProvider.java
shark300: You are right, I'll move it into `org.apache.cxf.spring.boot.autoconfigure.micrometer.provider`

integration/spring-boot/autoconfigure/src/test/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/provider/jaxws/CustomTimed.java
shark300: This annotation helps to test meta annotations. The Spring's AnnotationCollector finds this annotations too. I didn't want to reduce feature set of original Spring implementation.

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/MicrometerMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Bean
+    public JaxwsTags jaxwsTags() {
+        return new JaxwsTags();
+    }
+
+    @Bean
+    @ConditionalOnMissingBean(ExceptionClassProvider.class)
+    public ExceptionClassProvider exceptionClassProvider() {
```
reta, line 79: I would suggest to remove `ExceptionClassProvider`  and `JaxwsFaultCodeProvider` altogether and replace it with something like `TagsCustomizer`. 
```
interface TagsCustomizer() {
    Iterable<Tag> getAdditionalTags(Exchange ex);
}
```

There could many of them (fe, could be one for JAX-WS and one for JAX-RS) and `MicrometerMetricsContext` could apply the available customizers before emitting the final set of tags. 

```
Supplier<Iterable<Tag>> tags = () -> this.tagsProvider.getTags(ex) + <all tags from customizers>
```

What do you think?

integration/spring-boot/autoconfigure/src/test/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/provider/jaxws/CustomTimed.java
reta: > This annotation helps to test meta annotations.
So you have one in tests, it should be sufficient to test meta annotations but no need to introduce it into main codebase, correct?

shark300: Yes, exactly.

integration/spring-boot/starter-jaxws/pom.xml
```diff
@@ -54,6 +55,10 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
```
reta, line 58: This should be optional dependency, correct? (not everyone would need actuator)

integration/spring-boot/starter-jaxws/pom.xml
```diff
@@ -69,6 +74,11 @@
             <artifactId>cxf-rt-frontend-jaxws</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
```
reta, line 77: This should be optional dependency as well

rt/features/metrics/src/main/java/org/apache/cxf/metrics/micrometer/provider/jaxws/JaxwsTags.java
```diff
@@ -0,0 +1,140 @@
...
+import org.apache.cxf.common.util.StringUtils;
+import org.apache.cxf.message.Exchange;
+import org.apache.cxf.message.Message;
+import org.apache.cxf.service.model.BindingOperationInfo;
+
+import io.micrometer.core.instrument.Tag;
+
+import static java.util.Optional.ofNullable;
+
+public class JaxwsTags {
```
reta, line 36: We could split this class into 2,  `StandardTags` (under `org.apache.cxf.metrics.micrometer.provider` package) and `JaxwsTags` with JAX-WS specifics, like FaultCode, etc. Along with customizers we could rely on `StandardTags` only in the `MicrometerMetricsContext` where JAX-WS specific tags are going to be injected by customizers.

integration/spring-boot/starter-jaxws/pom.xml
shark300: Yes it should. I've removed that and moved it into integration test's pom.xml. It uses that.

integration/spring-boot/starter-jaxws/pom.xml
shark300: I've changed it to compile scope.

rt/features/metrics/src/main/java/org/apache/cxf/metrics/micrometer/provider/jaxws/JaxwsTagsProvider.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.cxf.message.Message;
+import org.apache.cxf.metrics.micrometer.provider.ExceptionClassProvider;
+import org.apache.cxf.metrics.micrometer.provider.TagsProvider;
+
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+
+import static java.util.Optional.ofNullable;
+
+public class JaxwsTagsProvider implements TagsProvider {
```
reta, line 32: If we apply the idea with customizers (to replace narrow scoped `ExceptionClassProvider` and `JaxwsFaultCodeProvider`), we could rename this class to `StandardTagsProvider` and keep it independent from JAX-WS.

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/MicrometerMetricsAutoConfiguration.java
shark300: Fair point, it leaves open some possibilities.

rt/features/metrics/src/main/java/org/apache/cxf/metrics/micrometer/MicrometerMetricsContext.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private void record(TimingContext timingContext, Exchange ex) {
+        Set<Timed> annotations = timedAnnotationProvider.getTimedAnnotations(ex);
+        Timer.Sample timerSample = timingContext.getTimerSample();
+        Supplier<Iterable<Tag>> tags = () -> {
+            Iterable<Tag> defaultTags = this.tagsProvider.getTags(ex);
+            List<Tag> additionalTags = new ArrayList<>();
+            tagsCustomizers
+                    .forEach(tagsCustomizer -> tagsCustomizer.getAdditionalTags(ex).forEach(additionalTags::add));
+            return Iterables.concat(defaultTags, additionalTags);
```
reta, line 96: Bringing whole Guava for using `Iterables` seems too much, could you please rewrite it using standard library? Fe:

```
            return () -> Stream.concat(
                    StreamSupport.stream(defaultTags.spliterator(), false), 
                    StreamSupport.stream(additionalTags.spliterator(), false))
                .iterator();
```

rt/features/metrics/pom.xml
```diff
@@ -34,14 +34,33 @@
...
         </dependency>
 
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-core</artifactId>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.guava</groupId>
```
reta, line 44: Please remove this dependency, not need really (see please the suggested alternative)

rt/features/metrics/src/main/java/org/apache/cxf/metrics/micrometer/MicrometerMetricsContext.java
```diff
@@ -0,0 +1,138 @@
...
+        if (timingContext == null) {
+            startAndAttachTimingContext(request);
+        }
+    }
+
+    @Override
+    public void stop(long timeInNS, long inSize, long outSize, Exchange ex) {
+        Message request = ex.getInMessage();
+        TimingContext timingContext = TimingContext.get(request);
+        if (timingContext == null) {
```
reta, line 75: The flow with nullable `timingContext` concerns a bit. It should not happen, but assuming it does, we are going to record effectively zeroed samples (since start and stop time would be mostly equal), which basically would screw / mess the metric. I would suggest to skip the `record` call in this case, or if we could, provide our own sample (since we know the time it took, `timeInNS`), what do you think?

reta:
@shark300 thanks a lot for working on that, it looks great, I just have a few comments, looks very good otherwise.

rt/features/metrics/src/main/java/org/apache/cxf/metrics/micrometer/MicrometerMetricsContext.java
shark300: Fixed, thank you. (Un)fortunately, I have not been working with Iterables for 6 years. :)

rt/features/metrics/pom.xml
shark300: Fixed.

rt/features/metrics/src/main/java/org/apache/cxf/metrics/micrometer/MicrometerMetricsContext.java
shark300: Good catch!
Yes, it should not happen.
I've checked `Timer#Sample` class. On the one hand we can use `timeInNS` to calculate "original" timestamp from current time and this one. On the other hand, we have to create a custom implementation of `Clock#monotonicTime` for creating a "start" `Timer#Sample`. As I see, this is used internally and it has a mock implementation for testing. 
In my opinion we should skip the record call in this case, and log this case as a warning.

shark300:
@reta Thank you, and I'm sorry because fixing review comments takes much time.

rt/features/metrics/src/main/java/org/apache/cxf/metrics/micrometer/MicrometerMetricsContext.java
reta: 👍 agree

reta:
@shark300 not a problem at all, this is really nice feature, your contribution is greatly appreciated.

rt/features/metrics/src/main/java/org/apache/cxf/metrics/micrometer/MicrometerMetricsContext.java
shark300: Yes, sure.

shark300: Done.

shark300:
@reta I'm not familiar with this project, but I have some questions: who is responsible for documentation update and squashing commits?

reta:
@shark300 maintainers usually squash commits when merging to master (and sometimes cherry-pick to maintenance branches), the documentation is also updated by maintainers.

reta:
@shark300 I have no outstanding comments or concerns left, I just need to take some time to verify / play with this feature in different scenarios before merging it in, sounds good? Thanks again for your work.

shark300:
@reta Thank you for your work too :)

distribution/src/main/release/samples/jaxws_spring_boot/pom.xml
```diff
@@ -34,6 +35,15 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
```
reta, line 38: The `spring-boot-starter-actuator` is missed, sample project fails on startup (no registry):
```
        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-actuator</artifactId>
        </dependency>
```

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/MicrometerMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,144 @@
...
+        micrometerMetricsProperties.setAutoTimeRequests(server.isAutoTimeRequests());
+        micrometerMetricsProperties.setRequestsMetricName(server.getRequestsMetricName());
+
+        return new MicrometerMetricsProvider(registry, tagsProvider, tagsCustomizers, timedAnnotationProvider,
+                micrometerMetricsProperties);
+    }
+
+    @Bean
+    @Order(0)
+    public MeterFilter metricsSoapServerUriTagFilter() {
```
reta, line 137: I think we could strip `soap` from the bean name, may be name it as `metricsMaxAllowedServerUriTagsFilter`

distribution/src/main/release/samples/jaxws_spring_boot/pom.xml
shark300: You are right, I forgot to update this sample after I modified poms.

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/MicrometerMetricsAutoConfiguration.java
shark300: I rename it to `cxfMetricsMaxAllowedServerUriTagsFilter`

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/provider/SpringBasedTimedAnnotationProvider.java
```diff
@@ -0,0 +1,97 @@
...
+import io.micrometer.core.annotation.Timed;
+
+public class SpringBasedTimedAnnotationProvider implements TimedAnnotationProvider {
+
+    private ConcurrentHashMap<HandlerMethod, Set<Timed>> timedAnnotationCache = new ConcurrentHashMap<>();
+
+    @Override
+    public Set<Timed> getTimedAnnotations(Exchange ex) {
+        HandlerMethod handlerMethod = new HandlerMethod(ex);
+        return timedAnnotationCache.computeIfAbsent(handlerMethod, method -> {
```
reta, line 45: Sorry, a few issues I have run into with this method. 
1. The `computeIfAbsent` is known to introduce unnecessary locking on JDK8 in case when entry already exists (see please https://bugs.openjdk.java.net/browse/JDK-8161372). Since we are going to call `getTimedAnnotations` for each invocation, it is better to avoid these kind if issues if possible.
2. We allocate the empty sets every time for each method that does not have `@Timed` annotation, we could use `Collections.emptySet()` singleton instead.

The suggested modifications are below, hope they make sense (the same would apply to `DefaultTimedAnnotationProvider`)

```
    public Set<Timed> getTimedAnnotations(Exchange ex) {
        HandlerMethod handlerMethod = new HandlerMethod(ex);

        final Set<Timed> exists = timedAnnotationCache.get(handlerMethod);
        if (exists  != null) {
            return exists;
        }

       return timedAnnotationCache.computeIfAbsent(handlerMethod, method -> {
            Set<Timed> timed = findTimedAnnotations(method.getMethod());
            if (timed.isEmpty()) {
                timed = findTimedAnnotations(method.getBeanType());
            }
            return timed.isEmpty() ? Collections.emptySet() : timed;;
        });
    }

```

reta:
@shark300 may I ask you please to:

1. Merge with latest master (seems like some checks are failing,  `CodeQL / Analyze (java)`)
2. Run `mvn clean verify -Dmaven.test.skip=true -DskipITs=true` on this PR, there are some checkstyle violations (it should be covered by Jenkins job but it seems like it was not run for some reasons).

I also created the documentation template (https://cwiki.apache.org/confluence/display/CXF20DOC/Micrometer) and would ask your help please to review once ready (if you don't mind).

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/provider/SpringBasedTimedAnnotationProvider.java
shark300: Fixed.

shark300:
@reta 
PMD and Checkstyle is OK locally.

reta (approved):

PR #643, opened 2020-02-24
CXF-8215: SSE breaks Pipeline Processing

reta:
The JAX-RS Specification 2.1 contains the chapter 9 for "Server-Sent Events", specifically section 9.5 "Pipeline Processing" states:

> For compatibility purposes, implementations MUST initiate processing of an SSE response when either the first message is sent or when the resource method returns, whichever happens first. The initial SSE response, which may only include the HTTP headers, is processed using the standard JAX-RS pipeline as described in Appendix C. Each subsequent SSE event may include a different payload and thus require the use of a specific message body writer. Note that since this use case differs slightly from the normal JAX-RS pipeline, implementations SHOULD NOT call entity interceptors on each individual event (1).

In case of CXF SSE processing, the `JAXRSOutInterceptor` is not called and as such, the container response processing filters are also ignored.

### DIFF
--- a/rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/interceptor/SseInterceptor.java
+++ b/rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/interceptor/SseInterceptor.java
@@ -1,0 +1,82 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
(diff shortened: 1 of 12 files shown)


rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/interceptor/SseInterceptor.java
```diff
@@ -0,0 +1,69 @@
...
+    public SseInterceptor() {
+        super(Phase.POST_LOGICAL);
+    }
+    public SseInterceptor(String phase) {
+        super(phase);
+    }
+
+    public void handleMessage(Message message) {
+        if (!isRequestor(message) && message.get(SseInterceptor.class) == null) {
+            message.put(SseInterceptor.class, this);
```
maggu2810: Is this `put` and the above check if `get` is equal to null responsible that it is not called twice?

reta: Yes, it shouldn't be called twice, but it serves as a guard.

PR #644, opened 2020-02-27
[CXF-8225] Structure send method of Slf4jEventSender

bmaehr:
By extracting code into subroutines it is easier to override parts e.g. to make used Logger dependent of called method

### DIFF
--- a/rt/features/logging/src/main/java/org/apache/cxf/ext/logging/slf4j/Slf4jEventSender.java
+++ b/rt/features/logging/src/main/java/org/apache/cxf/ext/logging/slf4j/Slf4jEventSender.java
@@ -38,26 +38,10 @@ public class Slf4jEventSender implements LogEventSender {
 
     @Override
     public void send(LogEvent event) {
-        String cat = "org.apache.cxf.services." + event.getPortTypeName().getLocalPart() + "." + event.getType();
-        Logger log = LoggerFactory.getLogger(cat);
+        Logger log = getLogger(event);
         Set<String> keys = new HashSet<>();
         try {
-            put(keys, "Type", event.getType().toString());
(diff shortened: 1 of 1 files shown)


rt/features/logging/src/main/java/org/apache/cxf/ext/logging/slf4j/Slf4jEventSender.java
```diff
@@ -38,26 +38,10 @@
...
-            if (event.getServiceName() != null) {
-                put(keys, "ServiceName", localPart(event.getServiceName()));
-                put(keys, "PortName", localPart(event.getPortName()));
-                put(keys, "PortTypeName", localPart(event.getPortTypeName()));
-            }
-            if (event.getFullContentFile() != null) {
-                put(keys, "FullContentFile", event.getFullContentFile().getAbsolutePath());
-            }
-            put(keys, "Headers", event.getHeaders().toString());
+            fillMDC(event, keys);
```
reta: Since you are improving it, could we make `fillMDC` return keys instead of mutating them?

```
Set<String> keys = fillMDC(event);
```

Thanks!

bmaehr: I can do it and thought about it (because it would be cleaner), but it has a little side effect: If there is an exception in the middle of the method fillMDC now the set MDCs are correctly reseted. If I return the keys, then not.

reta: Very valid point, we could separate prepare and populate (`prepareMDC`/`populateMDC`) steps but it looks like too much at this moment. LGTM, thank you!

PR #645, opened 2020-03-01
CXF-8222 OIDC Refresh Response: ID Token iat Claim MUST represent the time that the new ID Token is issued

amarkevich:


rt/rs/security/sso/oidc/src/main/java/org/apache/cxf/rs/security/oidc/idp/IdTokenResponseFilter.java
```diff
@@ -78,7 +78,12 @@ private String getProcessedIdToken(ServerAccessToken st) {
                 idToken.setAudience(st.getClient().getClientId());
                 idToken.setAuthorizedParty(st.getClient().getClientId());
                 // if this token was refreshed then the cloned IDToken might need to have its
-                // issuedAt and expiry time properties adjusted if it proves to be necessary
+                // issuedAt and expiry time properties adjusted
+                if (null == st.getResponseType()) {
+                    final Long iat = st.getIssuedAt();
+                    idToken.setExpiryTime(iat + (idToken.getExpiryTime() - idToken.getIssuedAt()));
+                    idToken.setIssuedAt(iat);
+                }
```
coheigea: Could you explain the logic here? Why set issued at / expiry only if st.getResponseType() is null?

amarkevich: `st.getResponseType() == null` in case of request with param `grant_type=refresh_token`

coheigea: Is there a way instead we can tell that grant_type=refresh_token was used in the call?

PR #646, opened 2020-03-02
CXF-8229 OAuth 2.0 Dynamic Client Registration: Client Update Request

amarkevich:


systests/rs-security/src/test/java/org/apache/cxf/systest/jaxrs/security/oidc/OIDCDynamicRegistrationTest.java
```diff
@@ -237,17 +212,63 @@ public void testRegisterClientInitialAccessTokenCodeGrantTls() throws Exception
         assertEquals(200, wc.delete().getStatus());
     }
 
+    @org.junit.Test
+    public void testUpdateClient() throws Exception {
```
coheigea: Could you add a test to make sure it's not possible to "put" (or delete actually, GET is already covered) without specifying an access token?

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/utils/OAuthUtils.java
```diff
@@ -429,6 +429,6 @@ public static SignatureAlgorithm getClientSecretSignatureAlgorithm(Properties si
     }
 
     public static String convertListOfScopesToString(List<String> registeredScopes) {
-        return String.join(", ", registeredScopes);
+        return String.join(" ", registeredScopes);
```
coheigea: What's the reason behind this change?

systests/rs-security/src/test/java/org/apache/cxf/systest/jaxrs/security/oidc/OIDCDynamicRegistrationTest.java
amarkevich: ok

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/utils/OAuthUtils.java
amarkevich: https://tools.ietf.org/html/rfc7591#section-4.1.2

>    o  Client Metadata Name: "scope"
>    o  Client Metadata Description: Space-separated list of OAuth 2.0
>       scope values

PR #639, opened 2020-02-10
Incompatibility with recent Brave versions causes NullPointerException and hides root cause

timpeeters:
I believe there is an incompatibility between `cxf-integration-tracing-brave` and recent versions of `brave-instrumentation-http`.
CXF master currently depends on Brave version 5.6.9 while the newest version is 5.9.4.

As of version 5.7, `brave-instrumentation-http` throws a `NullPointerException` when invoking `brave.http.HttpClientHandler.create(HttpTracing, HttpClientAdapter)` with a `null` value for the `HttpClientAdapter` parameter.
Which is exactly what `cxf-integration-tracing-brave` does in case of an exception/SOAP fault.

This results in the following stack trace and hides the original SOAP fault from the user:

```
java.lang.NullPointerException: adapter == null
	at brave.http.HttpClientHandler.create(HttpClientHandler.java:64)
	at org.apache.cxf.tracing.brave.AbstractBraveClientProvider.stopTraceSpan(AbstractBraveClientProvider.java:126)
	at org.apache.cxf.tracing.brave.BraveClientStartInterceptor.handleFault(BraveClientStartInterceptor.java:55)
	at org.apache.cxf.phase.PhaseInterceptorChain.unwind(PhaseInterceptorChain.java:499)
```

Users of Spring Cloud Greenwich SR3 or higher are impacted.
Dependency graph from left to right with indication of the impacted versions:

| Spring Boot | Spring Cloud      | Sleuth        | Brave | Status     |
| ----------- | ----------------- | ------------- | ----- | ---------- |
| 2.1         | Greenwich.RELEASE | 2.1.0.RELEASE | 5.6.1 | Works      |
| 2.1         | Greenwich.SR1     | 2.1.1.RELEASE | 5.6.1 | Works      |
| 2.1         | Greenwich.SR2     | 2.1.2.RELEASE | 5.6.5 | Works      |
| 2.1         | Greenwich.SR3     | 2.1.3.RELEASE | 5.7.0 | **Broken** |
| 2.1         | Greenwich.SR4     | 2.1.6.RELEASE | 5.8.0 | **Broken** |
| 2.1         | Greenwich.SR5     | 2.1.7.RELEASE | 5.9.2 | **Broken** |
| 2.2         | Hoxton.RELEASE    | 2.2.0.RELEASE | 5.9.0 | **Broken** |
| 2.2         | Hoxton.SR1        | 2.2.1.RELEASE | 5.9.0 | **Broken** |

This pull requests:
[...]

PR #647, opened 2020-03-06
Bumps to latest brave

adriancole:
This is mainly to help someone on support who's trying to use latest stuff and having trouble aligning versions.

cc @timpeeters as he did a great job updating this last time in #639

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -89,8 +89,8 @@
         <cxf.atmosphere.version>2.5.2</cxf.atmosphere.version>
         <cxf.bcprov.version>1.64</cxf.bcprov.version>
         <cxf.brave.reporter.version>2.12.1</cxf.brave.reporter.version>
-        <cxf.brave.version>5.9.4</cxf.brave.version>
-        <cxf.brave.zipkin.version>2.19.3</cxf.brave.zipkin.version>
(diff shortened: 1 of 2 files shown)


parent/pom.xml
```diff
@@ -89,8 +89,8 @@
         <cxf.atmosphere.version>2.5.2</cxf.atmosphere.version>
         <cxf.bcprov.version>1.64</cxf.bcprov.version>
         <cxf.brave.reporter.version>2.12.1</cxf.brave.reporter.version>
-        <cxf.brave.version>5.9.4</cxf.brave.version>
-        <cxf.brave.zipkin.version>2.19.3</cxf.brave.zipkin.version>
+        <cxf.brave.version>5.10.1</cxf.brave.version>
+        <cxf.brave.zipkin.version>2.20.1</cxf.brave.zipkin.version>
```
coheigea: According to the parent pom for 5.10.1, the Zipkin version is still 2.19.3, so I guess we should stay aligned with that: https://repo.maven.apache.org/maven2/io/zipkin/brave/brave-parent/5.10.1/brave-parent-5.10.1.pom

adriancole: it doesn't matter, but no problem reverting to that.

PR #648, opened 2020-03-06
CXF-8233: Wrong response code in case of error

kkrisz1:
https://issues.apache.org/jira/browse/CXF-8233


gnodet:
Closing this PR — it only adds a test to reproduce the bug but contains no production fix. The unrelated whitespace changes also make it harder to review. If CXF-8233 (wrong response code for `@Oneway` endpoints when a `ContainerRequestFilter` throws an exception) is still reproducible, it should be reimplemented as a fresh PR with both the fix and a clean test. Thank you for the contribution!

PR #649, opened 2020-03-09
CXF-8235 Handle null continuation in AsyncResponseImpl

WhiteCat22:
instead of throwing a NullPointerException

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/AsyncResponseImpl.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/AsyncResponseImpl.java
@@ -302,6 +302,11 @@ public class AsyncResponseImpl implements AsyncResponse, ContinuationCallback {
     private void initContinuation() {
         ContinuationProvider provider =
             (ContinuationProvider)inMessage.get(ContinuationProvider.class.getName());
+        if (provider == null) {
+            throw new IllegalArgumentException(
+                "Continuation not supported. " 
+                + "Please ensure that all servlets and servlet filters support async operations");
+        }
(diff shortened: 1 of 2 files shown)


WhiteCat22:
Please backport to 3.3.X. 
And 3.1.X (I realize this is a stretch)

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/AsyncResponseImpl.java
```diff
@@ -302,6 +302,11 @@ public synchronized void handleTimeout() {
     private void initContinuation() {
         ContinuationProvider provider =
             (ContinuationProvider)inMessage.get(ContinuationProvider.class.getName());
+        if (provider == null) {
+            throw new IllegalArgumentException(
```
reta: Very minor, I would probably go with `IllegalStateException` (seems to fit better to this particular issue) but `IllegalArgumentException` is also fine.

bmhm: The filters are a kind of argument, aren't they? When defining a filter in `web.xml` or via `@WebFilter`… So yes, both are probably fine.

Anyway, the important thing here is that users see an exception with a somewhat meaningful message. Without this fix, if you use e.g. Apache Shiro `shiro-servlet:1.4.0`, you can only see a `NullPointerException` with a CXF stack trace, which is not helpful at all. See the stack trace here: https://github.com/bmhm/shiro-openliberty-npe and the issue for Apache SHIRO here: https://issues.apache.org/jira/projects/SHIRO/issues/SHIRO-735.

PR #650, opened 2020-03-11
CXF-8237 OAuth 2.0 Authorization Server Metadata / OpenID Provider Metadata

amarkevich:
1. https://tools.ietf.org/html/rfc8414#section-3
2. https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata


rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/services/WellKnownService.java
```diff
@@ -0,0 +1,38 @@
...
+
+import javax.ws.rs.Path;
+
+public class WellKnownService {
+
+    public static final String WELL_KNOWN_PATH = ".well-known";
+
+    private AuthorizationMetadataService service;
+
+    @Path(".well-known")
```
coheigea: Use WELL_KNOWN_PATH here instead.

PR #651, opened 2020-03-13
[CXF-8236] add support of signature challenges in the STSClient

SergiusMohr:


rt/ws/security/src/main/java/org/apache/cxf/ws/security/trust/AbstractSTSClient.java
```diff
@@ -1491,6 +1491,16 @@ protected SecurityToken createSecurityToken(Element el, byte[] requestorEntropy)
...
+                } else if ("SignChallenge".equals(ln)) {
+                    el = DOMUtils.getFirstElement(el);
+                    if ("Challenge".equals(el.getLocalName())) {
+                        // maybe another implementation of the return object is more useful. 
+                        // We need to transport only two values: 
+                        // challengeValue and a marker for the kind of response (SignChallenge response)
+                        SecurityToken token = new SecurityToken(DOMUtils.getContent(el));
+                        token.setTokenType("SignChallenge");
+                        return token;
+                    }
```
coheigea, line 1503: I think we could have a new Challenge class which is referenced in SecurityToken. Challenge could have an enum describing the type, maybe just SIGNATURE for now.

rt/ws/security/src/main/java/org/apache/cxf/ws/security/trust/STSClient.java
```diff
@@ -87,6 +93,76 @@ public SecurityToken requestSecurityToken(
...
+     * <wst:Challenge>4451658898</wst:Challenge> </wst:SignChallengeResponse>
+     * </wst:RequestSecurityTokenResponse>
+     * 
+     * @param action
+     * @param challengeValue the challenge value that was received by first
+     *                       issue-response
+     * @return SecurityToken
+     * @throws Exception
+     */
+    public SecurityToken requestSecurityTokenResponse(String action, String challengeValue) throws Exception {
```
coheigea, line 115: I think I'd prefer to remove this method altogether. Instead, refactor "issue" so that it can work for both cases. The only difference should be "RequestSecurityTokenResponse" instead of "RequestSecurityToken" for the normal issue case. Maybe the writing out of the Challenge itself could be delegated to the Challenge class itself.

Then in SecurityToken.requestSecurityToken, I think we could handle the SignChallenge case automatically. If the SecurityToken that is returned from the STS contains a SignChallenge then invoke again on the STS with the challenge.

gnodet:
Closing this stale PR — the review feedback from @coheigea (proper `Challenge` class, automatic challenge handling inside `issue()` flow, need for tests) was never addressed. If WS-Trust 1.4 SignChallenge support is still needed, it should be reimplemented from scratch following the design guidance from the review. Thank you for the contribution!

PR #652, opened 2020-03-16
CXF-8243 JwtAccessTokenValidator using the JWK Set URL

amarkevich:


rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/filters/JwsJwksJwtAccessTokenValidator.java
```diff
@@ -0,0 +1,106 @@
...
+
+import org.apache.cxf.jaxrs.client.WebClient;
+import org.apache.cxf.rs.security.jose.jaxrs.JsonWebKeysProvider;
+import org.apache.cxf.rs.security.jose.jwk.JsonWebKey;
+import org.apache.cxf.rs.security.jose.jwk.JsonWebKeys;
+import org.apache.cxf.rs.security.jose.jwk.PublicKeyUse;
+import org.apache.cxf.rs.security.jose.jws.JwsHeaders;
+import org.apache.cxf.rs.security.jose.jws.JwsSignatureVerifier;
+import org.apache.cxf.rs.security.jose.jws.JwsUtils;
+
```
coheigea: Can you add some javadoc explaining what the new validator is doing?

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/filters/JwsJwksJwtAccessTokenValidator.java
```diff
@@ -0,0 +1,106 @@
...
+        final Set<String> kids = new HashSet<>();
+        for (JsonWebKey jwk : getJsonWebKeys().getKeys()) {
+            if (PublicKeyUse.ENCRYPT != jwk.getPublicKeyUse()) {
+                final String kid = jwk.getKeyId();
+                kids.add(kid);
+                final JwkHolder h = new JwkHolder(jwk);
+                if (keyId.equals(kid)) {
+                    jwkHolder = h;
+                } else {
+                    jsonWebKeys.putIfAbsent(kid, h);
```
coheigea: I find it hard to follow the logic in this method. Why is the JwkHolder added to jsonWebKeys for kid if the key ids don't match (and then removed on line 77?)

systests/rs-security/src/test/java/org/apache/cxf/systest/jaxrs/security/oidc/filters/OIDCFiltersTest.java
```diff
@@ -111,45 +101,6 @@ public void testClientCodeRequestFilter() throws Exception {
         assertEquals(returnedBook.getId(), 123L);
     }
 
-    @org.junit.Test
```
coheigea: Why is it being removed?

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/filters/JwsJwksJwtAccessTokenValidator.java
amarkevich: added new jwk only; at L77 old keys removed
this case covered in JwsJwksJwtAccessTokenValidatorTest

systests/rs-security/src/test/java/org/apache/cxf/systest/jaxrs/security/oidc/filters/OIDCFiltersTest.java
amarkevich: Initially this test was introduced exactly for AT signature validation using JWKS but its more complex to configure and do not support key rotation

PR #654, opened 2020-03-22
CXF-8242: Stop blocking executor thread on microprofile rest asynchronous call

reta:
Replacing CXF's future implementation with Java 8 `CompletableFuture` delegate.

### DIFF
--- a/rt/frontend/jaxws/src/test/java/org/apache/cxf/jaxws/JaxwsClientCallbackTest.java
+++ b/rt/frontend/jaxws/src/test/java/org/apache/cxf/jaxws/JaxwsClientCallbackTest.java
@@ -1,0 +1,162 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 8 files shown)


reta:
@andymc12 @coheigea guys if don't mind to take a look, it simplifies callback logic a lot

rt/frontend/jaxws/src/test/java/org/apache/cxf/jaxws/JaxwsClientCallbackTest.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+    
+    @After
+    public void tearDown() throws Exception {
+        executor.shutdown();
+        executor.awaitTermination(5, TimeUnit.SECONDS);
+    }
+    
+    @Test
+    public void testHandleResponseCallback() throws Exception {
```
andymc12: Is it possible for the tests to inherit from the ClientCallbackTest?  There might be subtle differences that I'm missing, but it might be nice to consolidate a lot of this test code in the base test class and then have the other tests (Jaxws, Jaxrs, MP, etc.) just offer their own sub-class instance of the callback.

reta: Thanks @andymc12, makes perfect sense, I will consolidate them in order to reduce the duplication of test code.

reta: Hm ... turned out to be more complicated than I thought. This is a test code and each callback is in different module (cfx-core, cxf-rt-client, cxf-mp-client, cxf-jarws-frontend). I could move all of them the `systests` but the tests really belong to the its module (where callbacks live).

andymc12: No worries. I wondered if that would be an issue.  And I agree with you that the right place for these tests is in the module with the runtime code that they test.  Thanks for putting this together!

reta: :+1:

PR #655, opened 2020-03-24
CXF-8248 - workaround to get the root cause when the hostname contains an underscore

zerikv:


gnodet:
Closing this stale PR. The fix is a self-described workaround with a potential race condition (checking the `exception` field immediately after async `execute()`). The HC4 transport (`http-hc`) is also legacy, superseded by the HC5 transport (`http-hc5`). If this issue persists with HC5, it should be reported as a new JIRA issue. Thank you for the contribution!

PR #656, opened 2020-03-26
Consolidate getTargetMethod calls

andymc12:
I was looking at getting the target method in a new interceptor and realized that there are a few different places in the code that is essentially doing the same thing.  So I thought it would be better for maintenance to consolidate that function in one place.

Feedback appreciated.  If nobody objects, I'd like to back port this to 3.3.x too.

Thanks!

### DIFF
--- a/core/src/main/java/org/apache/cxf/validation/AbstractValidationInterceptor.java
+++ b/core/src/main/java/org/apache/cxf/validation/AbstractValidationInterceptor.java
@@ -107,7 +107,7 @@ public abstract class AbstractValidationInterceptor extends AbstractPhaseInterce
         Message inMessage = message.getExchange().getInMessage();
         Method method = null;
         if (inMessage != null) {
-            method = MessageUtils.getTargetMethod(inMessage, null);
+            method = MessageUtils.getTargetMethod(inMessage).orElse(null);
         }
         if (method == null) {
             method = message.getExchange().get(Method.class);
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/cxf/validation/AbstractValidationInterceptor.java
```diff
@@ -108,15 +107,7 @@ protected Method getServiceMethod(Message message) {
...
-            method = (Method)inMessage.get("org.apache.cxf.resource.method");
-            if (method == null) {
-                BindingOperationInfo bop = inMessage.getExchange().getBindingOperationInfo();
-                if (bop != null) {
-                    MethodDispatcher md = (MethodDispatcher)
-                        inMessage.getExchange().getService().get(MethodDispatcher.class.getName());
-                    method = md.getMethod(bop);
-                }
-            }
+            method = MessageUtils.getTargetMethod(inMessage, null);
```
reta: Minor, may be introducing the overloaded method `MessageUtils.getTargetMethod(inMessage)` (which calls the original one with `null` supplier) would be a bit more cleaner, no need to guess what second argument is. Alternatively, could be `Optional<Method>` so the caller can raise exceptions or get along with `null`. What do you think?

reta:
Thanks @andymc12 !

core/src/main/java/org/apache/cxf/validation/AbstractValidationInterceptor.java
andymc12: Good idea!  I'll give the `Optional<Method>` approach a try.  Thanks!

PR #657, opened 2020-04-03
[CXF-8151] Upgrade to Bean Validation 2.x and Hibernate Validator 6.1.0

deki:
along with Swagger Core 2.1.2

### DIFF
--- a/osgi/karaf/features/src/main/resources/features.xml
+++ b/osgi/karaf/features/src/main/resources/features.xml
@@ -314,7 +314,7 @@
         <bundle start-level="35">mvn:org.apache.cxf/cxf-rt-rs-service-description-swagger-ui/${project.version}</bundle>
         <bundle start-level="40">mvn:org.apache.cxf/cxf-rt-rs-service-description-swagger/${project.version}</bundle>
         <bundle start-level="35" dependency="true">mvn:com.fasterxml.jackson.datatype/jackson-datatype-jsr310/${cxf.jackson.version}</bundle>
-        <bundle start-level="10" dependency="true">mvn:javax.validation/validation-api/${cxf.validation.api.version}</bundle>
+        <bundle start-level="10" dependency="true">mvn:jakarta.validation/jakarta.validation-api/${cxf.validation.api.version}</bundle>
         <bundle start-level="35" dependency="true">mvn:org.apache.commons/commons-lang3/${cxf.commons-lang3.version}</bundle>
         <bundle start-level="30" dependency="true">mvn:org.javassist/javassist/${cxf.javassist.version}</bundle>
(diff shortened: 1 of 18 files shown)


ilgrosso:
@deki Why not moving Apache BVal instead (latest version: 2.0.3)?

deki:
@ilgrosso CXF is not using BVal, is it?

ilgrosso:
@deki exactly: I was thinking why we don't switch to BVal rather than upgrading Hibernate Validator, as in its latest version BVal is implementing Bean Validation 2.x anyway.

deki:
Ok got it. Well I like the idea but it's not just drop-in replacement, for example I see references to org.hibernate.validator classes in samples

ilgrosso:
It should be only in [a class](https://github.com/apache/cxf/blob/master/osgi/itests/src/test/java/org/apache/cxf/osgi/itests/jaxrs/BookStore.java) even though AFAICT there should never be need to explicitly call a given impl. 

Anyway, I understand such a change might not be completely on-scope here.

deki:
I'd suggest we try to finish this one and try the BVal changes afterwards.

osgi/karaf/features/src/main/resources/features.xml
```diff
@@ -314,7 +314,7 @@
         <bundle start-level="35">mvn:org.apache.cxf/cxf-rt-rs-service-description-swagger-ui/${project.version}</bundle>
         <bundle start-level="40">mvn:org.apache.cxf/cxf-rt-rs-service-description-swagger/${project.version}</bundle>
         <bundle start-level="35" dependency="true">mvn:com.fasterxml.jackson.datatype/jackson-datatype-jsr310/${cxf.jackson.version}</bundle>
-        <bundle start-level="10" dependency="true">mvn:jakarta.validation/jakarta.validation-api/${cxf.validation.api.version}</bundle>
+        <bundle start-level="10" dependency="true">mvn:javax.validation/validation-api/1.1.0.Final</bundle>
```
reta, line 317: @amarkevich It would be better to override manifest instead, 
```
<bundle start-level="10" dependency="true">mvn:jakarta.validation/jakarta.validation-api/${cxf.validation.api.version}$overwrite=merge&amp;javax.validation.constraints;version="[1.1,3)"</bundle>
```
I have sent the PR against 1.6.x branch as well (https://github.com/swagger-api/swagger-core/pull/3497) but it may take some time.

amarkevich: added

reta: :+1: Thanks!

reta:
@deki @amarkevich I think we could merge it like that guys, and keep Swagger Core 1.6.x on the radar, thoughts?

PR #659, opened 2020-04-16
CXF-8259: Blocked thread of JMS transport during jboss shutdown

reta:
CC @ffang 
Discussion thread: https://github.com/apache/cxf/pull/603/files#diff-fcece95462c2633e642f31c76583896c


reta:
CC @steingebein

steingebein:
Hi @reta,

I think it's ok to remove the synchronized keyword from `restartConnection`, because `onException` in `ExceptionListener` is also synchronized. 
The fix from https://github.com/apache/cxf/pull/603 should not be affected.

Greetings

PR #660, opened 2020-04-17
[CXF-8263] Support SSEs in MP Rest Client 2.0

andymc12:
Hi All - if you have time, could you review my changes for SSE support in MP Rest Client 2.0?  I developed this at the same time as the spec/API changes so there may need to be some tweaks to this implementation if the MP community requests changes to the spec/API.  

The relevant MP API issue is https://github.com/eclipse/microprofile-rest-client/issues/11 and my proposed spec/API/TCK changes are at https://github.com/eclipse/microprofile-rest-client/pull/266

Thanks!

### DIFF
--- a/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseMessageBodyReader.java
+++ b/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseMessageBodyReader.java
@@ -1,0 +1,69 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.cxf.microprofile.client.sse;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.concurrent.ForkJoinPool;
+
+import javax.ws.rs.Produces;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.GenericType;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.ext.MessageBodyReader;
+import javax.ws.rs.ext.Providers;
+import javax.ws.rs.sse.InboundSseEvent;
+
+import org.apache.cxf.jaxrs.impl.tl.ThreadLocalProviders;
+import org.reactivestreams.Publisher;
+
+@Produces(MediaType.SERVER_SENT_EVENTS)
+public class SseMessageBodyReader implements MessageBodyReader<Publisher<?>> {
+
+    @Context
+    Providers providers;
+
+    @Override
+    public boolean isReadable(Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
+        return Publisher.class.isAssignableFrom(type) && MediaType.SERVER_SENT_EVENTS_TYPE.isCompatible(mediaType);
+    }
+
+    @Override
+    public Publisher<?> readFrom(Class<Publisher<?>> type, Type genericType,
+            Annotation[] annotations, MediaType mediaType, MultivaluedMap<String, String> httpHeaders,
+            InputStream entityStream) throws IOException, WebApplicationException {
+        Providers providersImpl = (providers instanceof ThreadLocalProviders) ? ((ThreadLocalProviders)providers).get() 
+                                                                              : providers;
+        SsePublisher publisher = new SsePublisher(entityStream, ForkJoinPool.commonPool(), providersImpl);
+        if (genericType instanceof ParameterizedType) {
+            Type typeArgument = ((ParameterizedType)genericType).getActualTypeArguments()[0];
+            if (typeArgument.equals(InboundSseEvent.class)) {
+                return publisher;
+            }
+
+            return new SseTypeSafeProcessor<Object>(new GenericType<Object>(typeArgument), publisher);
+        }
+        return null;
+    }
+}
--- a/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseMessageBodyReader.java
+++ b/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseMessageBodyReader.java
@@ -23,7 +23,7 @@ import java.io.InputStream;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
-import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.ExecutorService;
 
 import javax.ws.rs.Produces;
 import javax.ws.rs.WebApplicationException;
@@ -35,7 +35,10 @@ import javax.ws.rs.ext.MessageBodyReader;
 import javax.ws.rs.ext.Providers;
 import javax.ws.rs.sse.InboundSseEvent;
 
+import org.apache.cxf.jaxrs.ext.MessageContext;
+import org.apache.cxf.jaxrs.impl.ProvidersImpl;
 import org.apache.cxf.jaxrs.impl.tl.ThreadLocalProviders;
+import org.apache.cxf.microprofile.client.Utils;
 import org.reactivestreams.Publisher;
 
 @Produces(MediaType.SERVER_SENT_EVENTS)
@@ -43,6 +46,9 @@ public class SseMessageBodyReader implements MessageBodyReader<Publisher<?>> {
 
     @Context
     Providers providers;
+
+    @Context
+    private MessageContext mc;
 
     @Override
     public boolean isReadable(Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
@@ -53,9 +59,10 @@ public class SseMessageBodyReader implements MessageBodyReader<Publisher<?>> {
     public Publisher<?> readFrom(Class<Publisher<?>> type, Type genericType,
             Annotation[] annotations, MediaType mediaType, MultivaluedMap<String, String> httpHeaders,
             InputStream entityStream) throws IOException, WebApplicationException {
-        Providers providersImpl = (providers instanceof ThreadLocalProviders) ? ((ThreadLocalProviders)providers).get() 
-                                                                              : providers;
-        SsePublisher publisher = new SsePublisher(entityStream, ForkJoinPool.commonPool(), providersImpl);
+        ProvidersImpl providersImpl = (ProvidersImpl) (providers instanceof ThreadLocalProviders
+            ? ((ThreadLocalProviders)providers).get() : providers);
+        ExecutorService executor = Utils.getExecutorService(mc);
+        SsePublisher publisher = new SsePublisher(entityStream, executor, providersImpl);
         if (genericType instanceof ParameterizedType) {
             Type typeArgument = ((ParameterizedType)genericType).getActualTypeArguments()[0];
             if (typeArgument.equals(InboundSseEvent.class)) {
--- a/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SsePublisher.java
+++ b/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SsePublisher.java
@@ -22,8 +22,8 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
-import java.util.LinkedList;
 import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicBoolean;
 
@@ -38,7 +38,7 @@ public class SsePublisher implements Publisher<InboundSseEvent> {
     final Executor executor;
     final BufferedReader br;
     final Providers providers;
-    final List<SseSubscription> subscriptions = new LinkedList<>();
+    final List<SseSubscription> subscriptions = new CopyOnWriteArrayList<>();
     final AtomicBoolean isStarted = new AtomicBoolean(false);
 
     SsePublisher(InputStream is, Executor executor, Providers providers) {
@@ -80,7 +80,7 @@ public class SsePublisher implements Publisher<InboundSseEvent> {
                         line = br.readLine();
                     }
                     for (SseSubscription subscription : subscriptions) {
-                        subscription.fireComplete();
+                        subscription.complete();
                     }
                 } catch (IOException ex) {
                     for (SseSubscription subscription : subscriptions) {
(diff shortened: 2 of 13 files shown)


rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SsePublisher.java
```diff
@@ -0,0 +1,98 @@
...
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+
+public class SsePublisher implements Publisher<InboundSseEvent> {
+
+    final Executor executor;
+    final BufferedReader br;
+    final Providers providers;
+    final List<SseSubscription> subscriptions = new LinkedList<>();
+    AtomicBoolean isStarted = new AtomicBoolean(false);
```
reta, line 42: Should be `final` as well?

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SsePublisher.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    private void start() {
+        if (isStarted.compareAndSet(false, true)) {
+            executor.execute(() -> {
+                try (BufferedReader br2 = br) {
+                    SseEventBuilder builder = new SseEventBuilder(providers);
+                    String line = br.readLine();
+                    while (line != null && !subscriptions.isEmpty()) {
+                        if (line.startsWith("data: ")) {
```
reta, line 65: This is one is a bit tricky. The space after is `:` is optional (https://www.w3.org/TR/2015/REC-eventsource-20150203/#event-stream-interpretation), both `data: ` and `data:` are valid event encodings.

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseSubscription.java
```diff
@@ -0,0 +1,115 @@
...
+    @Override
+    public void request(long n) {
+        if (n < 1) {
+            throw new IllegalArgumentException("Only postive values are valid - passed-in " + n);
+        }
+        requested.addAndGet(n);
+        synchronized (buffer) {
+            InboundSseEvent bufferedEvent = null;
+            while (delivered.get() < requested.get()
+                   && (bufferedEvent = ((LinkedList<InboundSseEvent>) buffer).pollFirst()) != null) {
```
reta, line 59: May be just declare `buffer` a `LinkedList` instead of `Queue` (it is private member anyway) so to eliminate casting?

rmannibucau:
Hi @andymc12 , instead of using org.reactivestreams, why not using java >= 9 Flow? It will avoid a dependency already replaced and you don't want in most cases (when you dont use SSE or are on tjava LTS, ie java 11 today).

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SsePublisher.java
andymc12: Thanks for the link - it sounds like we can probably check for "data:" and then left-trim the resulting string.

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SsePublisher.java
andymc12: Good catch - will change.

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseSubscription.java
andymc12: That was my original design, but Checkstyle prevents using `LinkedList` as a declared field.  I suppose I could add comments to disable Checkstyle in the code.

andymc12:
@rmannibucau I think that the MicroProfile 4.0 spec (of which MP Rest Client 2.0 will be a child) declares a minimum Java level of Java 8. I agree with you that the Flow API would be preferable, so I've asked the question on the MP mailing list if we could increase the minimum Java level.  Hopefully we can use Java 11.

In the MP Rest Client 2.0 spec text, I added wording that implementations may also support Java 9 Flow API.  Assuming that wording is approved, I would like to support it in CXF.

Thanks for the feedback!

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SsePublisher.java
reta: Right but it applies to all attributes: `id`, `data`, `comment`, ...

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseSubscription.java
reta: :+1: would be probably better since there are too many casts ...

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseSubscription.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @Override
+    public void request(long n) {
+        if (n < 1) {
+            throw new IllegalArgumentException("Only postive values are valid - passed-in " + n);
+        }
+        requested.addAndGet(n);
+        synchronized (buffer) {
+            InboundSseEvent bufferedEvent = null;
+            while (delivered.get() < requested.get()
```
reta, line 59: I think the `completed` condition is missed here (unexpected, due to error fe), no `onNext` calls are expected in this case.

reta:
@andymc12 It seems like Reactive Streams is a good choice for now (and with JDK11 baseline, `Flow` API could be the replacement) but I would suggest to run Reactive Streams TCK tests against `SsePublisher` (https://github.com/reactive-streams/reactive-streams-jvm/tree/v1.0.3/tck), there are a number of requirements the implementation should follow (if would be more than happy to help with that).

rmannibucau:
@reta i didnt see the need of reactivestreams in MP (and it should stay like that to stay aligned on java). Issue starting to use it is to mix reactive styles accross impl and have to manage this dep, this is why i'd prefer to not use it where not needed (and SSE does not need it ;)).

reta:
@rmannibucau I found Reactive Streams to be good abstraction, every single reactive stream implementation supports it (so interop would be the least concern here I believe), afaik even MP does have own spec based of Reactive Streams (https://github.com/eclipse/microprofile-reactive-streams-operators). Certainly SSE impl does not need Reactive Streams, but I think it is helpful.

rmannibucau:
@reta hmm, MP reactive-streams-operators was supposed to redefine (with the same pitfall that it is no more aligned on what people use on the JVM). Seems it hasnt be done but the spec is not yet really part of MP stack I guess, so still a kind of big sandbox ;). 
To be very concrete, today you have 3 kind of (java) users:

1. JVM ones (java >= 11 in general),
2. RxJava ones (they don't care of these abstractions cause they lack too much "helpers",
3. java 8 guys doing what they can (often reactive streams).

Indeed you can bridge between all these impl, even with the standard servlet or jaxrs reactive API (what this PR does) but it just adds noise in apps and is something to maintain in time, this is why I generally try to care to never add a dependency I don't need, and here it is not really needed.

A quick note on the interop which is not always "perfect" (got a few issue with different netty libs if netty was not aligned upfront), can be seen as minor but still.

Then, to be even more concrete, the interesting part of the reactive programming on producer side is the threading and this PR uses common pool which completely breaks the scalability of apps until tuned...on the JVM which prevents the tuning of the app if used anywhere else. So whatever is done, it will need to use a dedicated executor - either in the MP API or through cxf properties/extensions but not the default CXF one which is sized for other kind of requests.

Then in terms of API you just reimpl it completely (cause reactivestreams is not an impl as Rx can be) so importing reactivestreams for 3 interfaces with 2-3 methods is very luxury IMHO.

I'd prefer a module next to the default one enabling to unwrap as a reactive impl the cxf impl rather than the opposite which would enable to embrace all cases and drop reactivestreams when java 8 will be deprecated for example.

Hope it makes sense.

reta:
@rmannibucau It certainly makes sense (implementation wise the PR needs some work, the ForkJoin pool was also on my list but I thought to have TCK tests first). Yeah, you seem to have more insights into "real" state of MP specs,  if there is a better alternative to align with directions the MP is going,  I am definitely fine with that (keeping in mind we have to support Java 8). Thanks for the thoughts!

andymc12:
Hi @reta @rmannibucau  - sorry for the delay on this.  

* Wrt Java Flow API - I agree that that would be preferable to reactive streams.  I discussed the idea of prereq-ing Java 11 with the MicroProfile community and have been pretty much shot down.  The current MP Reactive Streams Operators uses reactive streams and will continue to do so in MP 4.0.  I'm not giving up hope that we can change their mind though, because the 4.0 release is being delayed due to IP working group issues - so that is more time to plead my case. :-). But for now, I'd like to go ahead with using reactive streams to be consistent with the rest of the MP community.

* Wrt to the reactive streams TCK - I really like that idea.  I'm thinking that it might be preferable to add the RS TCK to the MP Rest Client TCK so that all implementations of MP Rest Client 2.0 must implement a compliant Publisher for InboundSseEvents.  As you both have probably guessed, my initial implementation in this PR is not compliant.  I'm working on both the TCK in the MP project and the CXF implementation in parallel.  

* Wrt the ForkJoinPool - good catch!  I'll plan to use the ExecutionService specified by the RestClientBuilder API, which defaults to the ForkJoinPool if none is specified.  Wdyt?

Thanks for the feedback!

andymc12:
Ok, all the changes are in - and it is passing the Reactive Streams TCK.  I'm thinking I'll still leave the PR in draft mode until the MP Rest Client PR is merged - that way I can publish a milestone build of MP Rest Client that contains the updated TCK.  Once the updated TCK is available from Maven, then I can update the parent pom file and officially verify that this passes both the Reactive Streams TCK and the MP Rest Client TCK.

andymc12:
@reta - this is now passing the Reactive Streams TCK (which is pulled in as part of the MP Rest Client TCK in 2.0-m1).  Can you take another pass at reviewing the changes?  Thanks!

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ProvidersImpl.java
```diff
@@ -62,4 +62,8 @@ public ProvidersImpl(Message m) {
     private Type getGenericType(Class<?> type, Type genericType) {
         return genericType == null ? type : genericType;
     }
+
+    public Message getMessage() {
```
reta, line 66: Probably not a good idea, exposing CXF internals (see please suggestion for `SseMessageBodyReader`)

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseMessageBodyReader.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.cxf.message.Message;
+import org.apache.cxf.microprofile.client.Utils;
+import org.reactivestreams.Publisher;
+
+@Produces(MediaType.SERVER_SENT_EVENTS)
+public class SseMessageBodyReader implements MessageBodyReader<Publisher<?>> {
+
+    @Context
+    Providers providers;
+
```
reta, line 49: CXF has `MessageContext` contextual extension to get the message context:
```
@Context
private MessageContext messageContext;
```

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseMessageBodyReader.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    public Publisher<?> readFrom(Class<Publisher<?>> type, Type genericType,
+            Annotation[] annotations, MediaType mediaType, MultivaluedMap<String, String> httpHeaders,
+            InputStream entityStream) throws IOException, WebApplicationException {
+        ProvidersImpl providersImpl = (ProvidersImpl) (providers instanceof ThreadLocalProviders
+            ? ((ThreadLocalProviders)providers).get() : providers);
+        Message message = providersImpl.getMessage();
+        ExecutorService executor = Utils.getExecutorService(message);
```
reta, line 62: And you could use `messageContext` here to check the correct executor :-)

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseMessageBodyReader.java
andymc12: Excellent suggestion - thanks!

andymc12:
@reta thanks for the review. I implemented your `MessageContext` idea.  Please let me know if you have any other concerns or suggestions.

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseSubscription.java
```diff
@@ -0,0 +1,135 @@
...
+            fireError(new IllegalArgumentException("Only positive values may be requested - passed-in " + n));
+            return;
+        }
+        requested.addAndGet(n);
+        synchronized (buffer) {
+            InboundSseEvent bufferedEvent = null;
+            while (delivered.get() < requested.get()
+                   && (bufferedEvent = buffer.pollFirst()) != null) {
+                InboundSseEvent finalEvent = bufferedEvent;
+                delivered.updateAndGet(l -> {
```
reta, line 67: Sorry, the usage `updateAndGet` was "bothering" me and now I understand why ... I believe we should rethink the approach, here is the problem:
```
    /* ...
     * @param updateFunction a side-effect-free function
     * ...
     */
    public final long updateAndGet(LongUnaryOperator updateFunction) {
```
The `updateFunction` must be `a side-effect-free function` (because it could be applied multiple times) but in current implementation is not. I have observed interesting races between `request` and `fireEvent`, which always are scheduled on different threads, when the value of `delivered` was changed and `updateFunction`  was applied multiple time, calling `onNext` several times.

Does it make sense to you?

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseSubscription.java
```diff
@@ -0,0 +1,135 @@
...
+            return;
+        }
+        if (n < 1) {
+            fireError(new IllegalArgumentException("Only positive values may be requested - passed-in " + n));
+            return;
+        }
+        requested.addAndGet(n);
+        synchronized (buffer) {
+            InboundSseEvent bufferedEvent = null;
+            while (delivered.get() < requested.get()
```
reta, line 64: Probably needs `!canceled.get() && !completed.get()` as well ...

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseSubscription.java
andymc12: It does (and sorry for the slow turnaround) - the main thing I want to ensure is that the call to `onNext` results in the `delivered` count being incremented.  I think I can do that with a sync block.

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseSubscription.java
andymc12: I think I had something similar coded up a while back - but that broke the reactive streams TCK.  IIUC, the subscriber can close or cancel the subscription while the subscription is still sending it requested items.

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseSubscription.java
reta: :+1: thanks!

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/sse/SseSubscription.java
reta: Aha I see, looks like an edge, good to know, thanks!

PR #662, opened 2020-04-25
CXF-8261: Exceptions being thrown in a ClientResponseFilter are ignored

reta:
Section 4.5.2 of the JAX-RS 2.1 specification:

> When a provider method throws an exception, the JAX-RS client runtime will map it to an instance of ProcessingException if thrown while processing a request, and to a ResponseProcessingException if thrown while processing a response.


reta:
Thanks @andymc12!

PR #666, opened 2020-04-29
DigestAuthSupplier: fix nc-value and update cnonce generation

amarkevich:
https://tools.ietf.org/html/rfc2617#section-3.2.2
1. The nc-value is the *hexadecimal* count of the number of requests
2. MUST NOT be specified if the server did not send a qop directive in the WWW-Authenticate header field


amarkevich:
I'll check

amarkevich:
@coheigea added tests against Spring Boot and Jetty server

PR #669, opened 2020-05-04
[CXF-8276] BOM POM File with only CXF modules

mederel:
CXF parent was changed in 3.2.5 to contain all the CXF modules in the dependency managemenet which was a great improvement for CXF users when integrating CXF as they would have to worry so much about dependency version. A project I am involved in though is also a framework project and provides CXF amongst a whole lot of other dependencies, and integrating the CXF thirdparty dependency tree becomes quite tricky.

Hence the need for a lighter BOM (build of material) POM file, containing only the modules of the CXF project. This would follow the example of netty for instance in https://github.com/netty/netty/blob/4.1/bom/pom.xml with their netty-bom module.


bom/pom.xml.bak
```diff
@@ -0,0 +1,617 @@
+<?xml version="1.0"?>
```
amarkevich, line 1: remove unwanted resources please

bom/pom.xml
```diff
@@ -0,0 +1,616 @@
...
+                <artifactId>cxf-rt-frontend-simple</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.cxf</groupId>
+                <artifactId>cxf-rt-javascript</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.cxf</groupId>
```
reta, line 158: We probably should remove the test dependencies (xxx-tests, xxx-itests, xxx-test, test-utils), I am not sure publishing them in BOM is useful, wdyt?

mederel: All modules that have a src/test but no src/test are now moved back to parent/pom.xml.

bom/pom.xml.bak
mederel: Indeed, stupid of me. File removed.

bom/pom.xml
```diff
@@ -0,0 +1,581 @@
...
+                <artifactId>cxf-spring-boot-starter-jaxrs</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.cxf</groupId>
+                <artifactId>cxf-spring-boot-starter-jaxws</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
```
amarkevich, line 382: one more test dependency

bom/pom.xml
```diff
@@ -0,0 +1,581 @@
...
+                <groupId>org.apache.cxf.karaf</groupId>
+                <artifactId>cxf-karaf-commands</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.cxf.services.sts</groupId>
+                <artifactId>cxf-services-sts-core</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
```
amarkevich, line 504: and this one

bom/pom.xml
mederel: Thing is I see here users outside of CXF scope. Test utility classes could be useful for somebody else no?

https://mvnrepository.com/artifact/org.apache.cxf/cxf-testutils/3.3.6

bom/pom.xml
mederel: Moved to parent with the other sts systests.

bom/pom.xml
reta: :+1: thank you

bom/pom.xml
reta: True, I see it being used by Camel fe (https://mvnrepository.com/artifact/org.apache.camel/camel-cxf/2.25.1), I think we could keep it there but with `<scope>test</scope>`, makes sense?

mederel: I don't think that it is a good idea, should be for each user to decide. Example: what if one'd want to develop additional test utility classes for their project using the CXF test utils?

Also there are multiple pom.xml in CXF that use cxf-testutils without scope test:
systests/rs-sse/rs-sse-base/pom.xml, systests/cdi/base/pom.xml, ...

reta: Got a very mixed set of opinions regarding test dependencies in BOMs, let's leave it like that, no explicit `<scope>`,  thank you.

PR #672, opened 2020-05-08
CXF-8206 Update to saaj-impl 1.5.2

amarkevich:


deki:
Seems to work but I haven't check osgi. I think we should target 3.4.0 for merge...

coheigea:
@amarkevich Can you rebase the branch off master?

amarkevich:
@coheigea done

coheigea:
@ffang  - Any objections?

ffang:
@amarkevich @coheigea 
Sorry for the late reply.

I tested OSGi related tests both with JDK8 and 11, all passed, so I think we are good to go!

Freeman

PR #675, opened 2020-05-29
CXF-8290 - Don't lock by default in org.apache.cxf.jaxrs.model.Abstra…

coheigea:
…ctResourceInfo#getProxyMap


coheigea:
@rmannibucau

PR #677, opened 2020-06-08
CXF-8296: Custom Boolean ParamConverter wrong behaviour

reta:
The general issue is with the way CXF handles `null` parameter values injection, it does not actually distinguish if the converter was applied and returned `null`, or the result is just `null` for any reason (no converter found fe).

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/InjectionUtils.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/InjectionUtils.java
@@ -45,6 +45,7 @@ import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.ResourceBundle;
 import java.util.Set;
 import java.util.SortedSet;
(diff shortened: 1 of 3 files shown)


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/InjectionUtils.java
```diff
@@ -539,21 +547,32 @@ private static RuntimeException createParamConversionException(ParameterType pTy
...
             ServerProviderFactory pf = ServerProviderFactory.getInstance(message);
             ParamConverter<T> pm = pf.createParameterHandler(pClass, genericType, anns, message);
             if (pm != null) {
-                result = pm.fromString(value);
+                return Optional.of(pm);
             }
         }
-        return result;
+        
+        return Optional.empty();
```
andymc12, line 562: Minor suggestion - you could replace the if-check and this line with `return Optional.ofNullable(pm);`

reta: Surely, less code, thanks @andymc12!

PR #678, opened 2020-06-08
[CXF-8299] Impl for QueryParamStyles in RestClient

andymc12:
This is a draft PR for now since the new API (new builder method and enum) is not yet available in Maven.  

When ready, this will resolve JIRA [CXF-8299](https://issues.apache.org/jira/browse/CXF-8299).

It allows users to configure the style of query parameters when multiple values are specified for the same parameter key.  This may be useful in scenarios where the client developer has little to no control over the RESTful endpoint - which might require PHP-style query parameters, or comma-separated values.


andymc12:
Superseded by #680

PR #680, opened 2020-06-15
[CXF-8299][CXF-8304] QueryParamStyles and CDI-managed providers for MP Rest Client 2.0

andymc12:
This PR fixes [CXF-8299](https://issues.apache.org/jira/browse/CXF-8299) (QueryParamStyles) and [CXF-8304](https://issues.apache.org/jira/browse/CXF-8304) (CDI-managed providers), and supersedes PR #678. 

For 8299, this allows the client to specify how collections should be handle when using query parameters - i.e. expanded, expanded with array syntax (required for PHP servers), or with a single key but comma-separated values.

For 8304, this changes the way providers are instantiated. If CDI is managing an instance of the provider class, then the client implementation should use that instance rather than creating a new instance.  This should only apply in cases where the interface registers the provider _class_, and not cases where it registers an _instance_ of the provider.

### DIFF
--- a/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientConfigurableImpl.java
+++ b/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientConfigurableImpl.java
@@ -31,6 +31,7 @@ import javax.ws.rs.ext.WriterInterceptor;
 import org.apache.cxf.common.util.PropertyUtils;
 import org.apache.cxf.jaxrs.impl.ConfigurableImpl;
 import org.apache.cxf.jaxrs.impl.ConfigurationImpl;
+import org.apache.cxf.microprofile.client.cdi.CDIFacade;
 import org.apache.cxf.microprofile.client.config.ConfigFacade;
 import org.eclipse.microprofile.rest.client.ext.ResponseExceptionMapper;
 
@@ -41,6 +42,8 @@ public class MicroProfileClientConfigurableImpl<C extends Configurable<C>>
         ClientResponseFilter.class, ReaderInterceptor.class, WriterInterceptor.class,
         MessageBodyWriter.class, MessageBodyReader.class, ResponseExceptionMapper.class};
     private static final String CONFIG_KEY_DISABLE_MAPPER = "microprofile.rest.client.disable.default.mapper";
+
+    private Instantiator instantiator = CDIFacade.getInstantiator().orElse(super.getInstantiator());
 
     public MicroProfileClientConfigurableImpl(C configurable) {
         this(configurable, null);
@@ -58,5 +61,10 @@ public class MicroProfileClientConfigurableImpl<C extends Configurable<C>>
         }
         return ConfigFacade.getOptionalValue(CONFIG_KEY_DISABLE_MAPPER,
                                              Boolean.class).orElse(false);
+    }
+
+    @Override
+    protected Instantiator getInstantiator() {
+        return instantiator;
     }
 }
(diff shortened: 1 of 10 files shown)


rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/CDIInstantiator.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Map;
+
+import org.apache.cxf.jaxrs.impl.ConfigurableImpl.Instantiator;
+
+
+public final class CDIInstantiator implements Instantiator {
+
+    static final CDIInstantiator INSTANCE = new CDIInstantiator();
+
+    private final Map<Object, Instance<?>> cdiInstances = new HashMap<>();
```
reta, line 32: Won't `IdentityHashMap` be a bit more appropriate here since we track references essentially?

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientConfigurableImpl.java
```diff
@@ -42,6 +43,8 @@
         MessageBodyWriter.class, MessageBodyReader.class, ResponseExceptionMapper.class};
     private static final String CONFIG_KEY_DISABLE_MAPPER = "microprofile.rest.client.disable.default.mapper";
 
+    private Instantiator instantiator = CDIFacade.getInstantiator().orElse(super.getInstantiator());
```
reta, line 46: Minor, probably should be `final`

reta, line 46: A bit more thoughts on that, the MicroProfileClientConfigurableImpl is created from `CxfTypeSafeClientBuilder`, which it turn is created from:
 - CDI: `RestClientBean`
 - Non-CDI: `CxfRestClientBuilderResolver`
May be we could pass the instantiator through `CxfTypeSafeClientBuilder`? Fe `RestClientBean` could rely on CDI one while CxfRestClientBuilderResolver could use default. 

WDYT?

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/CDIInstantiator.java
andymc12: Agreed - good catch.

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientConfigurableImpl.java
andymc12: Agreed.

andymc12: I think it is a good idea, but when I tried it out, I realized that it breaks one of the TCK test cases - it fails when a new client is built using the `RestClientBuilder` from within a CDI-managed bean - specifically,[ this test](https://github.com/eclipse/microprofile-rest-client/blob/24096f2df248938fc00185a587bc812d969b5845/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/cditests/CDIManagedProviderTest.java#L91).

reta: Gotcha, thanks !

PR #681, opened 2020-06-27
CXF-8099: mask sensitive logging elements

ashakirin:
Added an option to mask sensible XML and JSON elements in logging


rt/features/logging/pom.xml
```diff
@@ -38,6 +38,10 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
+        <dependency>
```
rmannibucau, line 41: maybe a bit overkill for the actual usage? IMHO it would be great to drop or reuse cxf utils.

asdf913:
transform() method not found in **org.apache.cxf.ext.logging.AbstractLoggingInterceptor**
image

rt/features/logging/pom.xml
ashakirin: Perhaps, I will check

ashakirin:
> transform() method not found in **org.apache.cxf.ext.logging.AbstractLoggingInterceptor** in **org.apache.cxf:cxf-rt-features-logging:3.3.7** dependency

transform() method was recently introduced and exists only in 3.4.0-SNAPSHOT version

asdf913:
Thanks you for your work!!

rt/features/logging/pom.xml
coheigea: Agreed, there is no need to import it just for StringUtils.containsIgnoreCase

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/MaskSensitiveHelper.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private String applyExpression(
+            final String originalLogString,
+            final String matchPatternTemplate,
+            final String replacementTemplate) {
+        String resultString = originalLogString;
+        for (final String elementName : sensitiveElementNames) {
+            final String matchPattern = matchPatternTemplate.replaceAll(ELEMENT_NAME_TEMPLATE, elementName);
+            final String replacement = replacementTemplate.replaceAll(ELEMENT_NAME_TEMPLATE, elementName);
```
coheigea, line 69: Could we find a way to do this matching in the constructor so it doesn't have to be done every time the expression is evaluated?

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/AbstractLoggingInterceptor.java
```diff
@@ -102,8 +110,12 @@ public void createExchangeId(Message message) {
         }
     }
 
-    protected String transform(final String originalLogString) {
+    protected String transform(final Message message, final String originalLogString) {
         return originalLogString;
     }
 
+    protected String maskSensitiveElements(final Message message, String originalLogString) {
+        return (new MaskSensitiveHelper(sensitiveElementNames))
```
coheigea, line 118: MaskSensitiveHelper is thread-safe, so it would be more performant only to create it once when sensitiveElementNames are set?

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/MaskSensitiveHelper.java
ashakirin: I made some measurements year ago - there are no different in modern java with or without pre-compiled patterns - it is automatically optimized. However I find code with replaceAll more readable as with pattern. Can change it, but win is minimal

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/AbstractLoggingInterceptor.java
ashakirin: yes, it is make sense

ashakirin: only the thing that can contradict: I would like to add sensitive http headers names as well, if new separate setter for headers names will be introduced, instantiation of helper will be weird

rt/features/logging/pom.xml
ashakirin: updated

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/AbstractLoggingInterceptor.java
ashakirin: But perhaps it doesn't make sense to introduce separate setters for elements and headers, user can put them in the same list

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/AbstractLoggingInterceptor.java
```diff
@@ -73,6 +78,14 @@ public long getInMemThreshold() {
         return threshold;
     }
 
+    public void addSensitiveElementNames(final List<String> sensitiveElementNames) {
+        maskSensitiveHelper.addSensitiveElementNames(sensitiveElementNames);
+    }
+
+    public void addSensitiveHeaders(final List<String> sensitiveHeaders) {
```
coheigea, line 85: Maybe rename to HTTPHeaders to make the meaning clear (as opposed to say SOAP headers)

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/AbstractLoggingInterceptor.java
```diff
@@ -102,8 +115,13 @@ public void createExchangeId(Message message) {
         }
     }
 
-    protected String transform(final String originalLogString) {
+    protected String transform(final Message message, final String originalLogString) {
         return originalLogString;
     }
 
+    protected String maskSensitiveElements(final Message message, String originalLogString) {
+        return Optional.of(maskSensitiveHelper)
```
coheigea, line 123: Why Optional here when maskSensitiveHelper is declared as final?

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/AbstractLoggingInterceptor.java
ashakirin: Ignore my last two comments - headers must be masked in different code and shouldn't be initialized in MaskSensitiveHelper at all

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/AbstractLoggingInterceptor.java
ashakirin: Optional makes no sense anymore, stays from time, when maskSensitiveHelper was instantiated only by setter and wasn't final. Removed.

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/AbstractLoggingInterceptor.java
ashakirin: ProtocolHeaders indeed make more sense. Other question if it makes sense to support masking sensitive SOAP headers as well?

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/MaskSensitiveHelper.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private String applyExpression(
+            final String originalLogString,
+            final String matchPatternTemplate,
+            final String replacementTemplate,
+            final List<String> sensitiveNames) {
+        String resultString = originalLogString;
+        for (final String sensitiveName : sensitiveNames) {
+            final String matchPattern = matchPatternTemplate.replaceAll(ELEMENT_NAME_TEMPLATE, sensitiveName);
+            final String replacement = replacementTemplate.replaceAll(ELEMENT_NAME_TEMPLATE, sensitiveName);
```
coheigea, line 74: Could we move this template replacement into the constructor? Maybe create a map of matchPattern -> replacement and iterate over this instead of sensitiveNames?

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/MaskSensitiveHelper.java
```diff
@@ -0,0 +1,79 @@
...
+    public void addSensitiveElementNames(final List<String> inSensitiveElementNames) {
+        this.sensitiveElementNames.addAll(inSensitiveElementNames);
+    }
+
+    public String maskSensitiveElements(
+            final Message message,
+            final String originalLogString) {
+        if (sensitiveElementNames.isEmpty() || message == null) {
+            return originalLogString;
+        }
```
coheigea, line 49: Could we also return here if originalLogString is null + remove the Optional.ofNullable in applyMasks below?

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/AbstractLoggingInterceptor.java
```diff
@@ -73,6 +82,15 @@ public long getInMemThreshold() {
...
     }
 
+    public void addSensitiveElementNames(final List<String> sensitiveElementNames) {
+        maskSensitiveHelper.addSensitiveElementNames(sensitiveElementNames);
+    }
+
+    public void addSensitiveProtocolHeaderNames(final List<String> sensitiveProtocolHeaderNames) {
+        sensitiveProtocolHeaderMap = sensitiveProtocolHeaderNames.stream()
+                .collect(Collectors.toMap(Function.identity(), name -> Boolean.TRUE));
+    }
```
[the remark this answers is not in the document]
ashakirin: Yes, set will be enough, you are right

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/event/DefaultLogEventMapper.java
```diff
@@ -95,6 +97,17 @@ public LogEvent map(Message message) {
...
     }
 
+    private Map<String, String> maskHeaders(
+            final Map<String, String> headerMap,
+            final Map<String, Boolean> sensitiveHeaderMap) {
+        final Map<String, String> maskedHeaderMap = new HashMap<>();
+        headerMap.keySet().forEach(key -> {
+            maskedHeaderMap.put(key, sensitiveHeaderMap.containsKey(key)
+                    ? MASKED_HEADER_VALUE : headerMap.get(key));
+        });
```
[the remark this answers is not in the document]
ashakirin: Have replaced List on Set for all arguments, makes more sense and it is performant

PR #683, opened 2020-07-15
CXF-8303: MP: Context propagation impossible using AsyncInvocationInterceptorFactory

reta:
1. Aligning MP Client implementation to use only one executor service, configured through `MicroProfileClientFactoryBean` and propagated via contextual property `EXECUTOR_SERVICE_PROPERTY`
2. Always initialize executor service in the `MicroProfileClientFactoryBean`, using `ForkJoin` pool as default if not specified
3. Removed async delegation from `MPRestClientCallback::createFuture` since it is always executed in the context of executor service (configured through `MicroProfileClientFactoryBean`)

### DIFF
--- a/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientFactoryBean.java
+++ b/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientFactoryBean.java
@@ -19,15 +19,12 @@
 package org.apache.cxf.microprofile.client;
 
 import java.net.URI;
-import java.security.AccessController;
-import java.security.PrivilegedAction;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.ForkJoinPool;
 
 import javax.ws.rs.client.ClientRequestFilter;
 import javax.ws.rs.client.ClientResponseFilter;
@@ -66,7 +63,7 @@ public class MicroProfileClientFactoryBean extends JAXRSClientFactoryBean {
         super(new MicroProfileServiceFactoryBean());
         this.configuration = configuration.getConfiguration();
         this.comparator = MicroProfileClientProviderFactory.createComparator(this);
-        this.executorService = (executorService == null) ? defaultExecutorService() : executorService; 
+        this.executorService = (executorService == null) ? Utils.defaultExecutorService() : executorService; 
         this.secConfig = secConfig;
         super.setAddress(baseUri);
         super.setServiceClass(aClass);
@@ -156,11 +153,4 @@ public class MicroProfileClientFactoryBean extends JAXRSClientFactoryBean {
         }
         return providers;
     }
-    
-    private static ExecutorService defaultExecutorService() {
-        return AccessController.doPrivileged((PrivilegedAction<ExecutorService>)() -> {
-            return ForkJoinPool.commonPool();
-        });
-    }
-
 }
--- a/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientFactoryBean.java
+++ b/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientFactoryBean.java
@@ -19,12 +19,15 @@
 package org.apache.cxf.microprofile.client;
 
 import java.net.URI;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientFactoryBean.java
```diff
@@ -63,7 +66,7 @@ public MicroProfileClientFactoryBean(MicroProfileClientConfigurableImpl<RestClie
         super(new MicroProfileServiceFactoryBean());
         this.configuration = configuration.getConfiguration();
         this.comparator = MicroProfileClientProviderFactory.createComparator(this);
-        this.executorService = executorService;
+        this.executorService = (executorService == null) ? defaultExecutorService() : executorService;
```
rmannibucau, line 69: maybe make it lazy to not create it (forkjoin common pool) if the app does not use it and the client for its lifecycle?

systests/microprofile/client/async/pom.xml
```diff
@@ -76,6 +76,12 @@
                 <version>${cxf.johnzon.version}</version>
                 <scope>test</scope>
             </dependency>
+            <dependency>
```
rmannibucau, line 79: think you can use JsonbJaxrsProvider from johnzon-jsonb instead of importing this one which does not use jsonb, would be more consistent with MP probably?

reta: Sure, done, thanks!

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientFactoryBean.java
reta: Thanks for the review! This one is very tricky to make lazy. Essentially, this `executorService` sets (or unsets) the CXF's `EXECUTOR_SERVICE_PROPERTY` which basically controls the how the async client invocation should be handled (it is used by MP and other clients). If we don't provide `executorService` upfront, the property won't be set and async callback (`MPRestClientCallback`) is going to be called in the context of CXF's work queue pool. The common `ForkJoinPool` is present 99.9% of the time.

rmannibucau: es = new LazyExecutor();

lazyexecutor 100% delegates to common one, it just enables to not create threads until it is done so the pool is async, not the facing instance ;)

reta: Sorry, this part I didn't get. With respect to threads (aka workers), `ForkJoinPool` is 100% lazy: it does not preallocate threads upfront if there is no work to be done, the `ForkJoinPool`'s threads are created only when work is scheduled.

reta:
Thanks a lot, @andymc12 !

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/MicroProfileClientFactoryBean.java
rmannibucau: Yep, was not a good phrasing choice. Fact is getting triggers the common (var) instantiation which brings a lot of work - for nothing in some cases and that's what I'd like to avoid.
Having a facade enables to just load a class and do a new without dependencies (static init) and speed up the startup.
We faced it in openwebbeans and this is why we have that wrapper https://github.com/apache/openwebbeans/blob/9b04c5e906294bcad7c61f30b5c5fcc8956b2f17/webbeans-impl/src/main/java/org/apache/webbeans/event/NotificationManager.java#L954.

reta: :+1: thanks for bearing with me, the concern is clear now, fixed!

rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/Utils.java
```diff
@@ -21,34 +21,109 @@
...
+
+        @Override
+        public <T> T invokeAny(Collection<? extends Callable<T>> tasks, long timeout, 
+                TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
+            return getCommonPool().invokeAny(tasks, timeout, unit);
+        }
+    }
+    
+    private static ExecutorService getCommonPool() {
+        return AccessController.doPrivileged((PrivilegedAction<ExecutorService>) () -> {
```
rmannibucau, line 125: Guess you want to bypass the dopriviledge when there is no security manager
Otherwise looks good to me now

reta: :+1: done, thanks!

PR #684, opened 2020-07-20
[CXF-8317] Make netty transport HttpObjectAggregator.maxContentLength configurable

vladikM:


rt/transports/http-netty/netty-client/src/main/java/org/apache/cxf/transport/http/netty/client/NettyHttpClientPipelineFactory.java
```diff
@@ -42,18 +43,22 @@
...
+    private final int maxContentLength;
+
     public NettyHttpClientPipelineFactory(TLSClientParameters clientParameters) {
         this(clientParameters, 0);
     }
 
     public NettyHttpClientPipelineFactory(TLSClientParameters clientParameters, int readTimeout) {
         this.tlsClientParameters = clientParameters;
         this.readTimeout = readTimeout;
+        this.maxContentLength = SystemPropertyAction.getInteger(MAX_CONTENT_LENGTH, 1048576);
```
coheigea, line 61: I think it would be better to configure this in the same way as for readTimeout etc., by passing an int value through to this class, and in NettyHttpConduit checking the CXF Message for the value, e.g. see: https://github.com/apache/cxf/blob/b8e24192378aa2b9bd8110186673d8904f6237d5/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java#L617

rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
```diff
@@ -652,6 +652,25 @@ protected static int determineConnectionTimeout(Message message,
...
+            try {
+                maxResponseContentLength = Integer.parseInt(obj.toString());
+            } catch (NumberFormatException e) {
+                LOG.log(Level.WARNING, "INVALID_TIMEOUT_FORMAT", new Object[] {
+                    Message.MAX_RESPONSE_CONTENT_LENGTH, obj.toString()
+                });
+            }
+        }
+        if (maxResponseContentLength == null) {
+            maxResponseContentLength = Integer.MAX_VALUE;
```
coheigea, line 669: We should probably default instead to 1048576, which was the previous default. Maybe extract this number to a constant.

core/src/main/java/org/apache/cxf/message/Message.java
```diff
@@ -196,6 +196,12 @@
      */
     String THREAD_SAFE_STAX_FACTORIES = Message.class.getName() + ".THREAD_SAFE_STAX_FACTORIES";
 
+
+    /**
+     * Integer property to specify how much bytes client must be able to handle.
+     */
+    String MAX_RESPONSE_CONTENT_LENGTH = Message.class.getName() + ".MAX_RESPONSE_CONTENT_LENGTH";
```
coheigea, line 203: I think it's probably better to restrict this configuration variable to Netty instead of the CXF Message object, as it only applies to Netty.

rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
```diff
@@ -652,6 +652,25 @@ protected static int determineConnectionTimeout(Message message,
         return (int)ctimeout;
     }
 
+    protected static int determineMaxResponseContentLength(Message message,
```
coheigea, line 655: Move into Netty subclass.

rt/transports/http/src/main/resources/schemas/wsdl/http-conf.xsd
```diff
@@ -390,6 +390,13 @@
                 </xs:documentation>
             </xs:annotation>      
         </xs:attribute>
+        <xs:attribute name="MaxResponseContentLength" type="ptp:ParameterizedInt" use="optional" >
```
coheigea, line 393: I don't think this is required.

vladikM: removed

rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
vladikM: change default to 1048576, extracted constant

core/src/main/java/org/apache/cxf/message/Message.java
vladikM: removed from Message

rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
vladikM: moved

rt/transports/http-netty/netty-client/src/main/java/org/apache/cxf/transport/http/netty/client/NettyHttpClientPipelineFactory.java
```diff
@@ -44,16 +44,24 @@
...
     private final int readTimeout;
-    
+    private final int maxContentLength;
+
     public NettyHttpClientPipelineFactory(TLSClientParameters clientParameters) {
         this(clientParameters, 0);
     }
 
     public NettyHttpClientPipelineFactory(TLSClientParameters clientParameters, int readTimeout) {
+        this(clientParameters, readTimeout, 1048576);
```
coheigea, line 57: Could we just refer to a single constant value (DEFAULT_MAX_RESPONSE_CONTENT_LENGTH)?

rt/transports/http-netty/netty-client/src/main/java/org/apache/cxf/transport/http/netty/client/NettyHttpConduit.java
```diff
@@ -651,5 +660,24 @@ public void postShutdown() {
...
+            } catch (NumberFormatException e) {
+                LOG.log(Level.WARNING, "INVALID_TIMEOUT_FORMAT", new Object[] {
+                    MAX_RESPONSE_CONTENT_LENGTH, obj.toString()
+                });
+            }
+        }
+        if (maxResponseContentLength == null) {
+            maxResponseContentLength = SystemPropertyAction.getInteger(MAX_RESPONSE_CONTENT_LENGTH,
+                DEFAULT_MAX_RESPONSE_CONTENT_LENGTH);
+        }
```
coheigea, line 678: I don't think the system property is necessary, as determineConnection/ReceiveTimeout don't check it, only the message.

vladikM: removed

rt/transports/http-netty/netty-client/src/main/java/org/apache/cxf/transport/http/netty/client/NettyHttpClientPipelineFactory.java
vladikM: replaced with constant

PR #685, opened 2020-07-23
Update to activemq 5.16.0

amarkevich:


coheigea:
There is a problem with OSGi - https://issues.apache.org/jira/browse/AMQ-8009

coheigea:
I've submitted a PR to AMQ to fix the issue with Spring - https://issues.apache.org/jira/browse/AMQ-8009

osgi/itests/src/test/java/org/apache/cxf/osgi/itests/soap/JmsServiceTest.java
```diff
@@ -78,8 +78,10 @@ private static ActiveMQConnectionFactory createConnectionFactory() {
         return OptionUtils.combine(
             cxfBaseConfig(),
             features(cxfUrl, "cxf-jaxws", "cxf-transports-jms"),
+            features(maven().groupId("org.apache.karaf.features").artifactId("spring-legacy").versionAsInProject()
```
amarkevich, line 81: spring-legacy feature descriptor added to avoid activemq-client feature installation error

coheigea: Could you add in a link to the AMQ issue here so that we know to remove it when 5.16.1 is released? Same for the spring-legacy dependency in the pom.

amarkevich: added

PR #687, opened 2020-07-30
Update to hsqldb 2.5.1

amarkevich:


rt/rs/security/sso/oidc/src/main/java/org/apache/cxf/rs/security/oidc/idp/OidcUserSubject.java
```diff
@@ -54,6 +55,7 @@ public OidcUserSubject(UserSubject sub) {
 
     @Lob
     @Basic(fetch = FetchType.EAGER)
+    @Column(length = 1024)
```
coheigea, line 58: Where does 1024 come from?

amarkevich: Looks like recent Hibernate cause [1]: 

> creates column as CLOB(255) 

So I set some higher value.

1. https://stackoverflow.com/a/26805926/3476615

coheigea: I'm wondering if a JWT token can be longer than 1024 characters....say if it has a lot of claims?

amarkevich: updated @Column annotation: explicit 'blob' column type without length

PR #688, opened 2020-08-10
CXF-8324: Allow wider range of values for attachment properties

theobisproject:
Provide better error if the given value is not allowed
Document allowed values for properties

I also had to do some special handling if someone provides a BigInteger which has a larger value than `Long.MAX_VALUE`.

### DIFF
--- a/core/src/test/java/org/apache/cxf/attachment/AttachmentUtilTest.java
+++ b/core/src/test/java/org/apache/cxf/attachment/AttachmentUtilTest.java
@@ -18,8 +18,19 @@
  */
 package org.apache.cxf.attachment;
 
+import java.io.File;
+import java.io.IOException;
+import java.math.BigInteger;
+
+import org.apache.cxf.io.CachedOutputStream;
+import org.apache.cxf.message.Message;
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/apache/cxf/attachment/AttachmentUtilTest.java
```diff
@@ -122,4 +133,204 @@ public void testCreateContentID() throws Exception {
...
+                cos);
+        replay(cos);
+        cos.setOutputDir(attachmentDirectory);
+        cos.setThreshold(102400L);
+        verify(cos);
+    }
+
+    @Test
+    public void stringAsAttachmentDirectory() throws IOException {
+        String attachmentDirectory = "/dev/null";
```
coheigea, line 319: Maybe for /dev/null we could instead just use something like Files.createTempDirectory instead - I'm not sure if the former would work properly on windows?

theobisproject: As long as nobody want's to access the file this works on Windows. A File instance does not validate the path when creating it.

PR #690, opened 2020-08-26
InputStream should not be closed prematurely

faucct:


coheigea:
What's the context behind this pull request? Can you create a JIRA describing the problem?

faucct:
I don't think I remember the whole context, but the problem was that in case of `MediaType.MULTIPART_FORM_DATA_TYPE` the message body was also a stream and was being consumed twice. When debugging this issue I have found this hack and decided to plug in there.

andymc12 (approved):
I think this change makes sense.

I think we put in the first isCompatible/copyAndGetEntityStream code in place to allow other JAX-RS interceptors to read the form data so that we could populate `@FormParam` parameters, but still leave the entity stream in a state that could be processed by user-provided entity providers and interceptors.  HTH

gnodet:
Closing this PR — the underlying issue was addressed by CXF-8356 (PR #719, merged Nov 2020), which fixed the `null` media type handling in `processRequestBodyParameter()`. Before that fix, when `mt` was `null`, the code fell into the `copyAndGetEntityStream` path for all content types. With the fix, `mt` now defaults to `contentTypeMt`, and multipart/form-data correctly takes the raw stream path — which is the correct behavior since multipart body readers handle the stream lifecycle properly. Thank you for the contribution!

PR #691, opened 2020-08-27
Update Netty to 4.1.51.Final

mkarg:
This PR proposes to update Netty to its latest bugfix 4.1.51.Final.


coheigea:
Thanks, but we have since updated to 4.1.52.Final.

PR #692, opened 2020-09-09
CXF-8227: Failure to comply with JAX-RS spec with ContainerRequestContext and WriterInterceptorContext

reta:
For some reasons, the implementation branched off into request / message property holders. If the request was available in the message context, all properties were mapped to request attributes, otherwise to message properties container. In order to preserve the compatibility, the fix still supports request attributes however at the same time populates message properties container.


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/PropertyHolderFactory.java
```diff
@@ -41,6 +46,60 @@ public static PropertyHolder getPropertyHolder(Message m) {
         void setProperty(String name, Object value);
         Collection<String> getPropertyNames();
     }
+    
+    private static class ServletRequestPropertyHolder extends MessagePropertyHolder {
+        private static final String ENDPOINT_ADDRESS_PROPERTY = "org.apache.cxf.transport.endpoint.address";
+        private HttpServletRequest request;
```
andymc12, line 52:
```suggestion
        private final HttpServletRequest request;
```

it looks like this is not intended to be changed, so we could explicitly mark it as final.

systests/jaxrs/src/test/java/org/apache/cxf/systest/jaxrs/BookApplication.java
```diff
@@ -106,6 +106,12 @@ public void setDefaultId(List<String> ids) {
         public void aroundWriteTo(WriterInterceptorContext context) throws IOException,
             WebApplicationException {
             context.getHeaders().putSingle("BookWriter", "TheBook");
+            
+            final Object property = context.getProperty("x-book");
+            if (property != null) {
+                context.getHeaders().putSingle("X-Book-Header", property);
```
andymc12, line 112:
```suggestion
                context.getHeaders().putSingle("X-Book-Response-Header", property);
```

Just a suggestion, but IIUC, the intent is to show that property set from the request filter is propagated to the writer interceptor (in the response flow).  By using a different header in the response, it makes it clear that the test is not just propagating headers from the request back to the response.

systests/jaxrs/src/test/java/org/apache/cxf/systest/jaxrs/JAXRSClientServerNonSpringBookTest.java
```diff
@@ -186,17 +186,33 @@ public void testGetBook123Application11PerRequest() throws Exception {
             doTestPerRequest("http://localhost:" + PORT + "/application11/thebooks/bookstore2/bookheaders");
         assertEquals("TheBook", r.getHeaderString("BookWriter"));
     }
+    
+    @Test
+    public void testGetBook123Application11PerRequestWithHeader() throws Exception {
+        Response r = 
+            doTestPerRequest("http://localhost:" + PORT + "/application11/thebooks/bookstore2/bookheaders", 
+                "PropValue");
+        assertEquals("PropValue", r.getHeaderString("X-Book-Header"));
```
andymc12, line 195:
```suggestion
        assertEquals("PropValue", r.getHeaderString("X-Book-Response-Header"));
```

see previous comment

andymc12 (approved):
Only minor comments - and totally optional.  LGTM - thanks @reta!

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/PropertyHolderFactory.java
reta: :+1: correct, thank you!

systests/jaxrs/src/test/java/org/apache/cxf/systest/jaxrs/BookApplication.java
reta: Completely agree, rewrote the test to reflect its purpose, thanks a lot!

reta:
Thanks a lot for the review, @andymc12 !

lapo-luchini:
This has fixed setting the attribute in `ContainerRequestFilter.filter` and reading it from `WriterInterceptor.aroundWriteTo` but still seems to lack attributes defined in a `HttpServletRequest`.

The work-around seems to be to inject the servlet request itself and reading from there directly:

```java
@Context
private HttpServletRequest servletRequest;
```

… but since the servlet request exists (after all, it was injected correctly) I think also `WriterInterceptorContext.getParameter` should access to it as stated in the JavaDoc: _In a Servlet container, on the server side, the properties are backed by the ServletRequest and contain all the attributes available in the ServletRequest._.

reta:
> This has fixed setting the attribute in `ContainerRequestFilter.filter` and reading it from `WriterInterceptor.aroundWriteTo` but still seems to lack attributes defined in a `HttpServletRequest`.

Correct, the fix was focused on JAR-RS part, not the `HttpServletRequest`, but `WriterInterceptorContext` may need a special treatment probably, thank you @lapo-luchini

reta:
@lapo-luchini I was looking into the way `WriterInterceptorContext.getProperty` works and it should provider access to servlet attributes (since it uses the same `PropertyHolderFactory` as `ContainerRequestContext` does ). Could you please provider a reproducer for the issue? Thank you.

PR #693, opened 2020-09-15
[CXF-8338] Add missing getters to JweJsonConsumer

jbeckers:
I'm using cxf-rt-rs-security-jose's JweJsonConsumer to parse a Json JWE (that I just created and encrypted using JweJsonProducer).

I'm doing this because I need to pass its components (headers, kid, iv, aad, ciphertext, auth tag, ...) to another API individually (the API does not accept the full JsonJWE).

Currently however, only getAad(), getAadText() and some getters for the headers are available. There's no getIv(), getCipherText(), etc

This pull request add the missing getters


coheigea (approved):

jbeckers:
Thanks!

PR #695, opened 2020-09-18
Enable to pass a bval Validator to BeanValidationProvider (it is more…

rmannibucau:
… than enough and faster than creating validators again and again) + enable to cache the fact no validation is needed

### DIFF
--- a/core/src/main/java/org/apache/cxf/validation/BeanValidationProvider.java
+++ b/core/src/main/java/org/apache/cxf/validation/BeanValidationProvider.java
@@ -216,7 +216,7 @@ public class BeanValidationProvider implements AutoCloseable {
         }
 
         public boolean shouldValidateBean(final Validator validator, final Class<?> clazz) {
-            return types.computeIfAbsent(clazz, it -> validator.getConstraintsForClass(it).hasConstraints());
+            return types.computeIfAbsent(clazz, it -> validator.getConstraintsForClass(it).isBeanConstrained());
         }
     }
 }
--- a/core/src/main/java/org/apache/cxf/validation/BeanValidationProvider.java
+++ b/core/src/main/java/org/apache/cxf/validation/BeanValidationProvider.java
@@ -21,6 +21,9 @@ package org.apache.cxf.validation;
 import java.lang.reflect.Method;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.function.Supplier;
 import java.util.logging.Logger;
 
 import javax.validation.Configuration;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/cxf/validation/BeanValidationProvider.java
```diff
@@ -111,46 +142,81 @@ private static void initFactoryConfig(Configuration<?> factoryCfg, ValidationCon
...
+                throw new ResponseConstraintViolationException(violations);
+            }
         }
     }
 
     public< T > void validateReturnValue(final T bean) {
-        final Set<ConstraintViolation< T > > violations = doValidateBean(bean);
+        Validator validator = factory.get();
+        if (runtimeCache != null && bean != null
+                && !runtimeCache.shouldValidateBean(validator, bean.getClass())) {
```
reta, line 171: Shouldn't it be `!runtimeCache.shouldValidateReturnedValue`?

core/src/main/java/org/apache/cxf/validation/BeanValidationProvider.java
```diff
@@ -111,46 +142,81 @@ private static void initFactoryConfig(Configuration<?> factoryCfg, ValidationCon
...
+        final Set<ConstraintViolation< T > > violations = doValidateBean(factory.get(), bean);
         if (!violations.isEmpty()) {
             throw new ConstraintViolationException(violations);
         }
     }
 
-    private< T > Set<ConstraintViolation< T > > doValidateBean(final T bean) {
-        return factory.getValidator().validate(bean);
+    private< T > Set<ConstraintViolation< T > > doValidateBean(final Validator validator, final T bean) {
+        if (validator.getConstraintsForClass(bean.getClass()).isBeanConstrained()) {
```
reta, line 188: Probably call to `runtime.shouldValidateBean` is needed

core/src/main/java/org/apache/cxf/validation/BeanValidationProvider.java
reta: Oh nevermind, there is no method in this call

core/src/main/java/org/apache/cxf/validation/BeanValidationProvider.java
```diff
@@ -111,46 +142,81 @@ private static void initFactoryConfig(Configuration<?> factoryCfg, ValidationCon
...
 
-        return factory.getValidator().forExecutables();
+        public boolean shouldValidateReturnedValue(final Validator validator, final Method method) {
+            return returnedValues.computeIfAbsent(method, m -> validator.getConstraintsForClass(m.getDeclaringClass())
+                    .getConstraintsForMethod(m.getName(), method.getParameterTypes())
+                    .hasConstrainedReturnValue());
+        }
+
+        public boolean shouldValidateBean(final Validator validator, final Class<?> clazz) {
+            return types.computeIfAbsent(clazz, it -> validator.getConstraintsForClass(it).hasConstraints());
```
reta, line 219: Should you also check for `isBeanConstrained` and cache that?

core/src/main/java/org/apache/cxf/validation/BeanValidationProvider.java
rmannibucau: think this call is always protected by a public call before which already have the exit condition - at least was the intent

rmannibucau:
@reta thanks for the review, fixed the last comment, think others are ok (already "protected" by the caller)

reta (approved):

reta:
@rmannibucau anything outstanding left for this pr? or good to merge?

rmannibucau:
@reta good to merge for me

PR #697, opened 2020-09-22
[CXF-8346] TCK Changes for URI and hasEntity checks

andymc12:
Fixes various "hasEntity" TCK tests where the entityStream is non-null, but empty.  Also fixes an issue with `ContainerRequestContext.setRequestUri(...)` methods.

Resolves [CXF-8346](https://issues.apache.org/jira/browse/CXF-8346).

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
@@ -20,7 +20,6 @@
 package org.apache.cxf.jaxrs.impl;
 
 import java.io.ByteArrayInputStream;
-import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.PushbackInputStream;
@@ -191,9 +190,8 @@ public final class ResponseImpl extends Response {
             } catch (IOException ex) {
                 throw new ProcessingException(ex);
             }
-        } else {
-            return true;
         }
+        return true;
     }
 
     @Override
@@ -478,7 +476,7 @@ public final class ResponseImpl extends Response {
                                                                   responseMessage);
                 // close the entity after readEntity is called.
                 T tCastLastEntity = castLastEntity();
-                shouldClose = shouldClose && !(tCastLastEntity instanceof Closeable)
+                shouldClose = shouldClose && !(tCastLastEntity instanceof AutoCloseable)
                     && !(tCastLastEntity instanceof Source);
                 if (closeAfterRead && shouldClose) {
                     close();
@@ -629,39 +627,8 @@ public final class ResponseImpl extends Response {
         };
     }
 
-    private enum PrimitiveTypes {
-        BYTE(Byte.class, byte.class) { },
-        SHORT(Short.class, short.class) { },
-        INTEGER(Integer.class, int.class) { },
-        LONG(Long.class, long.class) { },
-        FLOAT(Float.class, float.class) { },
-        DOUBLE(Double.class, double.class) { },
-        BOOLEAN(Boolean.class, boolean.class) { },
-        CHAR(Character.class, char.class) { };
-
-        private final Class<?> wrapper;
-        private final Class<?> primitive;
-
-        PrimitiveTypes(Class<?> wrapper, Class<?> primitive) {
-            this.wrapper = wrapper;
-            this.primitive = primitive;
-        }
-
-        public static PrimitiveTypes forType(Class<?> type) {
-            for (PrimitiveTypes primitive : PrimitiveTypes.values()) {
-                if (primitive.supports(type)) {
-                    return primitive;
-                }
-            }
-            return null;
-        }
-
-        public boolean supports(Class<?> type) {
-            return type == wrapper || type == primitive;
-        }
-    }
-
     private static boolean isBasicType(Class<?> type) {
-        return PrimitiveTypes.forType(type) != null || Number.class.isAssignableFrom(type);
+        return type.isPrimitive() || Number.class.isAssignableFrom(type) || Boolean.class.isAssignableFrom(type)
+            || Character.class.isAssignableFrom(type);
     }
 }
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
@@ -20,8 +20,10 @@
 package org.apache.cxf.jaxrs.impl;
 
 import java.io.ByteArrayInputStream;
+import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.PushbackInputStream;
 import java.io.Reader;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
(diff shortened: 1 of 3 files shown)


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
```diff
@@ -137,22 +143,70 @@ public Object getActualEntity() {
...
+                    return i != -1;
+                } else {
+                    try {
+                        if (is.available() > 0) {
+                            return true;
+                        }
+                    } catch (IOException ioe) {
+                        //Do nothing
+                    }
+                    int b = is.read();
```
reta, line 173: This would modify the state of the stream, correct?

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
```diff
@@ -137,22 +143,70 @@ public Object getActualEntity() {
...
+                    if (b == -1) {
+                        return false;
+                    }
+                    PushbackInputStream pbis;
+                    if (is instanceof PushbackInputStream) {
+                        pbis = (PushbackInputStream) is;
+                    } else {
+                        pbis = new PushbackInputStream(is, 1);
+                        if (lastEntity != null) {
```
reta, line 182: Unfamiliar with this part, do you have insights about `lastEntity` vs `entity`?

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
```diff
@@ -218,42 +274,46 @@ private Date doGetDate(String dateHeader) {
...
     }
 
+    @Override
     public URI getLocation() {
         Object header = metadata.getFirst(HttpHeaders.LOCATION);
-        if (header == null && outMessage != null) {
-            header = outMessage.get(Message.REQUEST_URI);
-        }
-        return header == null || header instanceof URI ? (URI) header
+        return header == null || header instanceof URI ? (URI)header
```
reta, line 305: Sorry, why the `outMessage.get(Message.REQUEST_URI)` got removed?

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
```diff
@@ -400,12 +472,30 @@ private Link makeAbsoluteLink(Link link) {
...
-                                                                       responseMessage);
-                autoClose(cls, false);
-                return castLastEntity();
+                                                                  anns,
+                                                                  entityStream,
+                                                                  mediaType,
+                                                                  responseMessage);
+                // close the entity after readEntity is called.
+                T tCastLastEntity = castLastEntity();
+                shouldClose = shouldClose && !(tCastLastEntity instanceof Closeable)
```
reta, line 481: Shouldn't `AutoCloseable` also be checked?

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
```diff
@@ -527,10 +621,47 @@ public String getReasonPhrase() {
...
+            return null;
+        }
+
+        public boolean supports(Class<?> type) {
+            return type == wrapper || type == primitive;
+        }
+    }
+
+    private static boolean isBasicType(Class<?> type) {
+        return PrimitiveTypes.forType(type) != null || Number.class.isAssignableFrom(type);
```
reta, line 665: You probably could reduce it to `return type.isPrimitive() || Number.class.isAssignableFrom(type) ||
Boolean.class.isAssignableFrom(type) || Char.class.isAssignableFrom(type));`  (PrimitiveTypes is not really needed in this case).

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
andymc12: `Closeable` extends `AutoCloseable`, so we could get both with by checking `AutoCloseable` - I'll update this.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
andymc12: I believe that was a TCK test.  IIRC, the TCK tests that `getLocation` should return null unless explicitly set as a `Location: <someURI>` header (via the `ResponseBuilder.location(URI)` method).  

I'm not positive, but I think it addresses this TCK test (currently listed in the failing tests):
com/sun/ts/tests/jaxrs/ee/rs/core/response/JAXRSClient.java#getLocationNotPresentTest_from_standalone: JAXRSClient_getLocationNotPresentTest_from_standalone

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
andymc12: Good call - I'm not sure why the PrimitiveTypes inner class was created... my first guess was performance, but it actually looks like it would perform _worse_.  I'll use your suggestion.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
andymc12: This does read one byte of the stream.  If it is -1, then it simply returns false (no entity).  If there is data on the stream, then we construct the PushbackInputStream to unread the byte we checked before.  I think this approach is optimal for the case where there is no entity, but not harmful in the case where there is.  If the stream supports marking (fair uncommon imo) or the available method (much more common), then we'll never get this far.  

Unless you object, I'd like to leave it this way, but I can add a unit test to ensure the different expectations and complete coverage (input stream supporting marking, available, etc.).

reta: No objections, the reasoning was to keep PushbackInputStream  manipulations close by but indeed, you have a good argument in not doing so, thank you.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
andymc12: I believe that `lastEntity` is used for optimizing the case where the entity is an InputStream and a user calls `readEntity(SomeObject.class)` - lastEntity will then be set to the instance of `SomeObject` so that subsequent readEntity calls can just return that object rather then re-reading it from the stream.  But since it is also possible to read an entity as an InputStream, this code covers cases like:
```
response.readEntity(InputStream.class);
if (response.hasEntity()) {
    response.readEntity(InputStream.class); // or response.readEntity(SomeObject.class), etc
```

reta: 👍 thank you

reta (approved):
Thanks @andymc12 !

coheigea:
Hi @andymc12 ,

This change is causing quite a few system test failures, e.g. https://ci-builds.apache.org/job/CXF/job/CXF-JDK8/85/#showFailuresLink
Could you take a look? Maybe the tests themselves need to be updated.

andymc12:
@coheigea Sorry about that.  I have fixes for the test cases proposed in #703.  Can you take a look at them to verify that you and/or @reta are ok with the changes?  Thanks

PR #699, opened 2020-09-24
CXF-8337: wsdl2java generates exceptions fields which do not follow naming convention

reta:
wsdl2java generates wrapper exception class for fault bean. The generated exception has method getFaultInfo(). This method name is correct according to JAX-WS specification. However this method returns fault bean stored in a field called "name of fault bean". The name of the field and getter do not follow the naming convention. 

In general, it does not cause issue unless such fault beans becomes serialization / deserialization targets (the absence of the setter method `setFaultInfo`  makes the matter difficult).

@coheigea @dkulp may I ask you please to take a look? thanks!


reta:
@coheigea if you don't mind to take a look, pretty trivial change :) thanks!

coheigea:
@reta No objections from me, but the build is currently broken - so I'd prefer to make sure all tests are green before merging.

PR #700, opened 2020-09-25
[CXF-7925] Dynamic WSDL Client creation fails on JDK 11 because it cannot compile generated classes

elguardian:
Allow to specify the classpath for app server where class loaders are not URL based.


elguardian:
this is under testing for integrating with wildfly

PR #702, opened 2020-10-01
Remove processing of older servlet versions

andymc12:
This PR is intended to get rid of code paths that are specific to Servlet versions below 3.1 - if we still want to support those older servlet versions, then please reject this PR, but they're really old and would probably break with other parts of CXF that depend on newer versions of Java/Jakarta EE specs.

Thanks!


gnodet:
Closing this stale PR — the code targets the old `javax.servlet` namespace. However, the cleanup is still valid: CXF now requires Jakarta Servlet 6.x, so the pre-3.1 backward-compatibility code is dead. A fresh PR will be submitted to remove this dead code against the current codebase. Thank you for the contribution!

PR #707, opened 2020-10-06
[CXF-8353] MediaType validation

andymc12:
Performs validation on the type and subtype of the MediaType string passed-in according to https://tools.ietf.org/html/rfc2045#section-5.1

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/MediaTypeHeaderProvider.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/MediaTypeHeaderProvider.java
@@ -87,7 +87,7 @@ public class MediaTypeHeaderProvider implements HeaderDelegate<MediaType> {
 
         String type = mType.substring(0, i);
         String subtype = mType.substring(i + 1, end);
-        if (subtype.indexOf('/') != -1) {
+        if (!isValid(type) || !isValid(subtype)) {
             throw new IllegalArgumentException("Invalid media type string: " + mType);
         }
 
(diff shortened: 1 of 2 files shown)


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/MediaTypeHeaderProvider.java
```diff
@@ -190,4 +190,35 @@ private static MediaType handleMediaTypeWithoutSubtype(String mType) {
         }
         throw new IllegalArgumentException("Media type separator is missing");
     }
+
+    // Determines whether the type or subtype contains any of the tspecials characters defined at:
+    // https://tools.ietf.org/html/rfc2045#section-5.1
+    private static boolean isValid(String str) {
+        final int len = str.length();
```
reta, line 197: Minor, may be check `str.trim().length()`, it seems weird media types like `application/   ` are still passing through.

reta (approved):

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/MediaTypeHeaderProvider.java
andymc12: @reta - sounds good.  Since we end up trimming both type and subType at lines 114 and 115, I moved those trims to before the `isValid` method is called (to lines 88 and 89).  Let me know if you'd prefer the trim to be in isValid instead.  Otherwise, I'll plan to merge this later today.  Thanks for the review!

reta: Looks good, thanks @andymc12 !

PR #713, opened 2020-10-25
CXF-8359: fixed mask for XML elements containing attributes

ashakirin:
Fixed mask for XML elements containing attributes

### DIFF
--- a/rt/features/logging/src/main/java/org/apache/cxf/ext/logging/MaskSensitiveHelper.java
+++ b/rt/features/logging/src/main/java/org/apache/cxf/ext/logging/MaskSensitiveHelper.java
@@ -20,6 +20,7 @@ package org.apache.cxf.ext.logging;
 
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


rt/features/logging/src/main/java/org/apache/cxf/ext/logging/MaskSensitiveHelper.java
```diff
@@ -70,15 +71,19 @@ public String maskSensitiveElements(
             final Message message,
             final String originalLogString) {
         if (replacementsXML.isEmpty() && replacementsJSON.isEmpty()
-                || message == null
```
reta, line 73: Why this condition was removed? It is exactly what will be returned on line #80 but with unnecessary use of `Optional::ofNullable(message)`?

ashakirin: The statement "contentType = Optional.ofNullable(message).map(m -> (String) m.get(Message.CONTENT_TYPE))" handles null message out of the box. Null check on top is redundant

reta: That is correct, so the question becomes: why do you need `Optional` at all? With the previous check in place, just:

```
final String contentType = (String) m.get(Message.CONTENT_TYPE));
if (contentType  == null) {
      return originalLogString;
}
```

It looks as straightforward as it gets (no need to create unnecessary instances).

ashakirin: Matter of taste, for me Optional alternative is a bit compacter

PR #716, opened 2020-10-29
CXF-8252: Add Micrometer metric support for JAX-RS (Spring Boot)

reta:
@shark300 fyi, there are a few changes but they should not impact the JAX-WS integration anyhow (expect `uri` tag which is using the `REQUEST_URI` instead of `BASE_PATH`)


shark300:
I'm not familiar with JAX-RS, so thank you for your support!

PR #717, opened 2020-11-02
Upgrade boot and spring

davidkarlsen:
This change is [Reviewable](https://reviewable.io/reviews/apache/cxf/717)


amarkevich:
Spring version update reverted: https://github.com/apache/cxf/commit/70fef6f5a78bb69548628a4380ac4abb98404dc9

reta:
@davidkarlsen that is correct, thanks @amarkevich, we have reported the regression (https://github.com/spring-projects/spring-framework/issues/25987) which should be fixed in next releases, it does not impact CXF directly but we have a number of tests to verify inter-frameworks integrations and those started to fail.

coheigea:
@davidkarlsen I already merged this update to master yesterday

PR #723, opened 2020-11-13
Update to Spring Boot 2.4.0 / Spring Framework 5.3.1 / Spring Security 5.4.1 / Reactor 3.4.0 / Micrometer 1.6.1

reta:
Update to Spring Boot 2.4.0 
Update to Spring Framework 5.3.1
Update to Spring Security 5.4.1
Update to Project Reactor 3.4.0 
Update to Micrometer 1.6.1

### DIFF
--- a/systests/transports/pom.xml
+++ b/systests/transports/pom.xml
@@ -331,6 +331,11 @@
             <version>${cxf.bcprov.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
 </project>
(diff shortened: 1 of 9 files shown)


reta:
@deki @coheigea @amarkevich the update looks good now, one thing I intended to incorporate is switch to Maven Surefire 3.0.0-M5 (so we could run JUnit 4 & JUnit 5 test cases side by side) but I run into an issue with tooling tests `org.apache.cxf.tools.common.ProcessorTestBase::getClassPath()` which seems to be relying on some Surefire intrinsic details, probably better to tackle as separate effort. If no objections, I would like to merge it into master.

amarkevich (approved):

systests/transports/pom.xml
```diff
@@ -331,6 +331,11 @@
             <version>${cxf.bcprov.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
```
coheigea, line 338: Why the need for the vintage-engine dependency here?

reta: The `transports` uses a mix of JUnit 4 and JUnit 5 tests, we need the Vintage engine to run JUnit 4 tests using JUnit 5 platform, otherwise they are going to be skipped.

coheigea (approved):

PR #724, opened 2020-11-16
cxf-tools-common: remove ProcessorTestBase.getClassPath

amarkevich:
more details at #723


reta:
Thanks a lot, @amarkevich !

PR #725, opened 2020-11-17
CXF-8372: Update Maven Surefire to 3.0.0-M5

reta:
@coheigea @amarkevich switching to Maven Surefire to 3.0.0-M5, thanks for help guys!


amarkevich (approved):

amarkevich:
both failures caused by unstable tests

reta:
@amarkevich there is something weird going on, one of the `transport-udp` test cases fails pretty consistently and I don't understand the reasons yet, looking into it ...

reta:
@coheigea @amarkevich seems like we are set, all `11,579 tests` are there, it turned out to be more difficult but the current configuration seems to be good and stable

PR #728, opened 2020-11-20
[CXF-8378]:Fix NoClassDefFoundError: org/apache/cxf/common/util/Refle…

jimma:
…ctionUtil

### DIFF
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/ReferencingAuthenticator.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/ReferencingAuthenticator.java
@@ -25,12 +25,13 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.net.Authenticator;
 import java.net.PasswordAuthentication;
-
-import org.apache.cxf.common.util.ReflectionUtil;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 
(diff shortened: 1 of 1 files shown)


rt/transports/http/src/main/java/org/apache/cxf/transport/http/ReferencingAuthenticator.java
```diff
@@ -122,15 +125,41 @@ PasswordAuthentication tryWith(Authenticator a) throws Exception {
         if (a == null) {
             return null;
         }
-        for (final Field f : ReflectionUtil.getDeclaredFields(Authenticator.class)) {
+        Field[] fileds = null;
```
andrei-ivanov, line 128: typo: `fileds`

jimma: Fixed. Thanks @andrei-ivanov .

andymc12 (approved):
LGTM - my only comment would be that you could check for the security manager in a static boolean and save repeated calls to `System.getSecurityManager() == null`.  Thanks!

jimma:
> LGTM - my only comment would be that you could check for the security manager in a static boolean and save repeated calls to `System.getSecurityManager() == null`. Thanks!

Thanks for review @andymc12 .  Changed this as you suggested.

PR #732, opened 2020-11-25
CXF-8357: RxJava2/RxJava3: Add support of Maybe type

reta:
We do support `Flowable` / `Observale` / `Single` but not `Maybe`


systests/jaxrs/src/test/java/org/apache/cxf/systest/jaxrs/reactor/MonoReactorTest.java
```diff
@@ -112,4 +113,21 @@ public void testMonoEmpty() throws Exception {
             .expectComplete()
             .verify();
     }
+    
+    @Test
+    public void testGetError() throws Exception {
```
reta, line 118: Adding test case for `Mono.error()` as well

PR #735, opened 2020-12-10
Update to Jetty 9.4.35.v20201120

reta:
The change in PropertyFileLoginModule  caused tests to fail (dependency on thread-local static `JAASLoginService.INSTANCE` has been introduced), see please https://github.com/eclipse/jetty.project/issues/5486


reta:
@coheigea @amarkevich could you guys please take a look, the changes are in tests only but the impact may be larger.

coheigea (approved):

PR #738, opened 2021-01-02
Cxf 3891 regex for add sensitive element names is replace only the last element

CarlosPanarello:
Fix the match pattern xml template for multiple elements in xml


ashakirin (approved):

reta:
@ashakirin should we merge this one? thanks!

PR #740, opened 2021-01-20
Fix typos "doesnt" -> "doesn't".

garydgregory:


tools/corba/src/main/java/org/apache/cxf/tools/corba/processors/wsdl/WSDLToCorbaHelper.java
```diff
@@ -1431,7 +1431,7 @@ private CorbaType processOMGUnion(XmlSchemaComplexType complex, QName defaultNam
                 } else if (fields.size() == 1) {
                     caselist.add("TRUE");
                 } else {
-                    String msg = "Discriminator Type doesnt match number of Choices in Union:" + name;
+                    String msg = "Discriminator Type doesn't match number of Choices in Union:" + name;
```
andymc12, line 1434:
```suggestion
                    String msg = "Discriminator Type does not match number of Choices in Union:" + name;
```

For log messages, it might be easier to translate "does not" than "doesn't".

andymc12 (commented):
I'd suggest "does not" at least for the logged warning.  That said, I know in other open source products, they recommend not changing warning and error messages so that user scripts that might be scanning for those particular log messages won't be broken.  I'm not sure what the log message migration policy is for CXF, so I'll defer to someone who has more experience with the product.  But otherwise, the changes look good to me.  Thanks for improving the readability of the code!

coheigea (approved):

PR #745, opened 2021-02-14
CXF-8422: Unclosed input streams after using org.apache.cxf.tools.wsdlto.WSDLToJava

reta:
Multiple input stream leaks where `URIResolver` is used. Marked `URIResolver`  as `AutoCloseable` and refactored its usage to `try-with-resources` in most places.

### DIFF
--- a/core/src/main/java/org/apache/cxf/catalog/OASISCatalogManager.java
+++ b/core/src/main/java/org/apache/cxf/catalog/OASISCatalogManager.java
@@ -99,11 +99,9 @@ public class OASISCatalogManager {
                 public String getResolvedEntity(String publicId, String systemId) {
                     String s = super.getResolvedEntity(publicId, systemId);
                     if (s != null && s.startsWith("classpath:")) {
-                        try {
-                            try (URIResolver r = new URIResolver(s)) {
-                                if (r.isResolved()) {
-                                    return r.getURL().toExternalForm();
-                                }
+                        try (URIResolver r = new URIResolver(s)) {
+                            if (r.isResolved()) {
+                                return r.getURL().toExternalForm();
                             }
                         } catch (IOException e) {
                             //ignore
--- a/core/src/main/java/org/apache/cxf/catalog/OASISCatalogManager.java
+++ b/core/src/main/java/org/apache/cxf/catalog/OASISCatalogManager.java
@@ -100,10 +100,10 @@ public class OASISCatalogManager {
                     String s = super.getResolvedEntity(publicId, systemId);
                     if (s != null && s.startsWith("classpath:")) {
                         try {
-                            URIResolver r = new URIResolver(s);
-                            if (r.isResolved()) {
-                                r.getInputStream().close();
-                                return r.getURL().toExternalForm();
+                            try (URIResolver r = new URIResolver(s)) {
+                                if (r.isResolved()) {
+                                    return r.getURL().toExternalForm();
+                                }
                             }
                         } catch (IOException e) {
                             //ignore
--- a/rt/frontend/simple/src/main/java/org/apache/cxf/endpoint/dynamic/DynamicClientFactory.java
+++ b/rt/frontend/simple/src/main/java/org/apache/cxf/endpoint/dynamic/DynamicClientFactory.java
@@ -923,11 +923,9 @@ public class DynamicClientFactory {
         }
 
 
-        try {
-            try (URIResolver resolver = new URIResolver(base, target)) {
-                if (resolver.isResolved()) {
-                    target = resolver.getURI().toString();
-                }
+        try (URIResolver resolver = new URIResolver(base, target)) {
+            if (resolver.isResolved()) {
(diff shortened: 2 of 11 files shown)


reta:
@coheigea would you mind to take a look please, should not take long, thanks!

tools/wadlto/jaxrs/src/main/java/org/apache/cxf/tools/wadlto/jaxb/CustomizationParser.java
```diff
@@ -118,8 +118,9 @@ private void addBinding(String bindingFile) throws XMLStreamException {
 
         Element root = null;
         try {
-            URIResolver resolver = new URIResolver(bindingFile);
-            root = StaxUtils.read(resolver.getInputStream()).getDocumentElement();
+            try (URIResolver resolver = new URIResolver(bindingFile)) {
```
amarkevich, line 121: redundant inner try

reta: :+1, fixed, thanks!

amarkevich: there are same changes in another places...

tools/wsdlto/databinding/jaxb/src/main/java/org/apache/cxf/tools/wsdlto/databinding/jaxb/JAXBDataBinding.java
```diff
@@ -1305,9 +1305,10 @@ private static String mapSchemaLocation(String target, String base, OASISCatalog
 
 
         try {
-            URIResolver resolver = new URIResolver(base, target);
-            if (resolver.isResolved()) {
-                target = resolver.getURI().toString();
+            try (URIResolver resolver = new URIResolver(base, target)) {
```
coheigea, line 1308: Do we really need to have a nested try statement here?

core/src/main/java/org/apache/cxf/catalog/OASISCatalogManager.java
```diff
@@ -100,10 +100,10 @@ public String getResolvedEntity(String publicId, String systemId) {
                     String s = super.getResolvedEntity(publicId, systemId);
                     if (s != null && s.startsWith("classpath:")) {
                         try {
-                            URIResolver r = new URIResolver(s);
-                            if (r.isResolved()) {
-                                r.getInputStream().close();
-                                return r.getURL().toExternalForm();
+                            try (URIResolver r = new URIResolver(s)) {
```
coheigea, line 103: Do we really need to have a nested try statement here?

coheigea (changes requested):

core/src/main/java/org/apache/cxf/catalog/OASISCatalogManager.java
reta: No, not really in this case, simplified, thanks!

tools/wsdlto/databinding/jaxb/src/main/java/org/apache/cxf/tools/wsdlto/databinding/jaxb/JAXBDataBinding.java
reta: Nope, outer is enough, thanks!

tools/wadlto/jaxrs/src/main/java/org/apache/cxf/tools/wadlto/jaxb/CustomizationParser.java
reta: Tried to address all of them, there are some places where I had to keep nested`try` structures since the exiting behavior (throw or return) would change.

rt/frontend/simple/src/main/java/org/apache/cxf/endpoint/dynamic/DynamicClientFactory.java
```diff
@@ -729,10 +729,10 @@ static void setupClasspath(StringBuilder classPath, ClassLoader classLoader)
 
     private URL composeUrl(String s) {
         try {
-            URIResolver resolver = new URIResolver(null, s, getClass());
-
-            if (resolver.isResolved()) {
-                return resolver.getURI().toURL();
+            try (URIResolver resolver = new URIResolver(null, s, getClass())) {
```
coheigea, line 732: Can we remove the nested try statements here as well?

reta: This one is a bit tricky since the `ServiceConstructionException` has to be thrown if URI is not resolved and not throwing any exceptions. I refactored it, please let me know what do you think.

coheigea (approved):

PR #746, opened 2021-02-16
[CXF-8361] Finalized support for Jakarta NS

ropalka:
https://issues.apache.org/jira/browse/CXF-8361
 * introduced new org.apache.cxf.jaxws.handler.jakarta_types package for JAXB de/serialization
 * added new methods in HandlerChainBuilder
 * enhanced handlers de/serialization tests

### DIFF
--- a/rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/handler/AnnotationHandlerChainBuilder.java
+++ b/rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/handler/AnnotationHandlerChainBuilder.java
@@ -25,43 +25,32 @@ import java.security.PrivilegedAction;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.ResourceBundle;
-import java.util.StringTokenizer;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.regex.Pattern;
 
 import javax.jws.HandlerChain;
 import javax.jws.WebService;
-import javax.xml.bind.JAXBContext;
-import javax.xml.bind.JAXBException;
 import javax.xml.namespace.QName;
 import javax.xml.ws.WebServiceException;
 import javax.xml.ws.handler.Handler;
 
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
-import org.w3c.dom.Node;
 
 import org.apache.cxf.Bus;
 import org.apache.cxf.common.classloader.ClassLoaderUtils;
 import org.apache.cxf.common.i18n.BundleUtils;
 import org.apache.cxf.common.i18n.Message;
-import org.apache.cxf.common.jaxb.JAXBUtils;
 import org.apache.cxf.common.logging.LogUtils;
 import org.apache.cxf.common.util.StringUtils;
+import org.apache.cxf.jaxws.handler.types.PortComponentHandlerType;
 import org.apache.cxf.staxutils.StaxUtils;
+
 
 @SuppressWarnings("rawtypes")
 public class AnnotationHandlerChainBuilder extends HandlerChainBuilder {
-
-    private static final String HANDLER_CHAINS_E = "handler-chains";
-    private static final String HANDLER_CHAIN_E = "handler-chain";
-    private static final String JAKARTAEE_NS = "https://jakarta.ee/xml/ns/jakartaee";
-    private static final String JAVAEE_NS = "http://java.sun.com/xml/ns/javaee";
     private static final Logger LOG = LogUtils.getL7dLogger(AnnotationHandlerChainBuilder.class);
     private static final ResourceBundle BUNDLE = LOG.getResourceBundle();
-    private static JAXBContext context;
-
     private ClassLoader classLoader;
 
     public AnnotationHandlerChainBuilder() {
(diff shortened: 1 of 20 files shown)


rsearls (approved):

reta:
Thanks for the PR, @ropalka, I think we got to the point when it is probably make sense to completely separate Jakartaee and Javaee name handler chains (I think you would agree). I suggest to:
1. Introduce `org.apache.cxf.jaxws.handler.jakartaee` and `org.apache.cxf.jaxws.handler.javaee`
2. Introduce `JakartaeeHandlerChainBuilder` and `JavaeeHandlerChainBuilder`
3. Basically, I think it safe to assume that either JakartaEE or JavaEE namespace is going to be used, so in this case `AnnotationHandlerChainBuilder` could inspect the namespace, instantiate  `JakartaeeHandlerChainBuilder` or`JavaeeHandlerChainBuilder`, and delegate to one of those. In this case, the conditionals should be gone and implementation becomes much cleaner. 

What do you think?

ropalka:
Hi @reta . It makes sense to me. What I am worry about is backward compatibility.
Isn't jaxws frontend considered to be a public API? In other words if I would refactor package
from org.apache.cxf.jaxws.handler.types to org.apache.cxf.jaxws.handler.javaee
I would break compatibility.

reta:
Thanks for the reply @ropalka. We won't be changing the public API, which is outlined by `AnnotationHandlerChainBuilder`. The `JakartaeeHandlerChainBuilder`  and `JavaeeHandlerChainBuilder` would encapsulate the implementation specifics to each namespace, and are not supposed to be used outside `AnnotationHandlerChainBuilder`. All the existing classes, under `org.apache.cxf.jaxws.handler.types` fe, could stay were they are. What do you think?

ropalka:
Hi @reta . The problem with your proposal is AnnotationHandlerChainBuilder is extending HandlerChainBuilder class.
If I would introduce both JakartaeeHandlerChainBuilder and JavaeeHandlerChain builders I would need to drop
"AnnotationHandlerChainBuilder extends HandlerChainBuilder" relationship which is again not public API backward compatible change. The proposal PR I sent is 100% backward compatible.

ropalka:
I enhanced this PR to be as much backward compatible as possible @reta.
I introduced new _org.apache.cxf.jaxws.handler.jakartaee_ package as you proposed.
I left _org.apache.cxf.jaxws.handler.types_ package untouched because of BC.
I introduced JAXB deserialization adaptors to be able to eliminate HandlerChainBuilder
'jakartaee' methods I indroduced in previous pull request proposal. WDYT about it now?

reta:
> 
> 
> I enhanced this PR to be as much backward compatible as possible @reta.
> I introduced new _org.apache.cxf.jaxws.handler.jakartaee_ package as you proposed.
> I left _org.apache.cxf.jaxws.handler.types_ package untouched because of BC.
> I introduced JAXB deserialization adaptors to be able to eliminate HandlerChainBuilder
> 'jakartaee' methods I indroduced in previous pull request proposal. WDYT about it now?

@ropalka thank you for the effort, it definitely looks better but the main problem is still unsolved: the processing of the JavaEE and JakartaEE namespaces is tangled. May I ask you please to take at look at https://github.com/apache/cxf/pull/756, it is based of `master` (without your changes) but illustrates the idea of how the processing could be separated. AFAIK it does not change or alter any public APIs, only the implementation details have been moved around. Would appreciate to hear any concerns, thank you.

jimma:
It's good to not change AnnotationHandlerChainBuilder's public api and hierarchy.   +1 for this change.

ropalka:
I updated this PR as you suggested @reta, please review.

rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/handler/AnnotationHandlerChainBuilder.java
```diff
@@ -177,149 +129,8 @@ public ClassLoader run() {
...
-        }
-    }
-
-    private static synchronized JAXBContext getContextForPortComponentHandlerType()
-        throws JAXBException {
-        if (context == null) {
-            context = JAXBContext.newInstance(PortComponentHandlerType.class);
-        }
-        return context;
+    List<Handler> buildHandlerChain(PortComponentHandlerType ht) {
```
reta, line 132: I would prefer to keep this implementation details (either `private` or even remove the function altogether), but I am perfrectly fine to make the change after merge.

rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/handler/BaseHandlerChainBuilder.java
```diff
@@ -0,0 +1,221 @@
...
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+
+import org.apache.cxf.common.i18n.BundleUtils;
+import org.apache.cxf.staxutils.StaxUtils;
+
+abstract class BaseHandlerChainBuilder {
+    private static final String HANDLER_CHAINS_E = "handler-chains";
+    private static final String HANDLER_CHAIN_E = "handler-chain";
+    final AnnotationHandlerChainBuilder delegate;
```
reta, line 42: Minor, I would make remove it from `BaseHandlerChainBuilder`, it is needed only in concrete implementation, will take care of it after merge if you prefer.

reta:
@ropalka thank you, LGTM, I have a few minor comments which I could address myself if you prefer, otherwise - I think we are good to go

reta (approved):

rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/handler/BaseHandlerChainBuilder.java
ropalka: I have no objections @reta - cleaning it up after merge sounds good to me.

rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/handler/AnnotationHandlerChainBuilder.java
ropalka: It cannot be private as it is used by both JakartaeeHandlerChainBuilder & JavaeeHandlerChainBuilder.

reta:
Thanks again for your work, @ropalka

lulseged:
I have tried 3.5.0-SNAPSHOT and miss a lot of jakarta.* packages. When do you plan to move to jakarta namespaces?
The change will not be backward compatible. What is the use of making it backward compatible?
I would create 4.0.0-SNAPSHOT and start jakarta namespace from there and keep 3.x.x javax namespace.

reta:
@lulseged please follow https://github.com/apache/cxf/pull/737

PR #747, opened 2021-02-17
Unclosed input streams needs to be closed across different classes using Java language feature 

alanmehio:
It would be good to use  java language feature "try with resources"  to close the input streams across different java classes.
The main purpose for doing so is to prevent possible memory leak. This pull request is triggered from the fix for 
 [core/src/main/java/org/apache/cxf/resource/URIResolver.java ](https://github.com/apache/cxf/pull/745/files#diff-4b89a654670abdd7e7144d7a91184997e9a0347798dc9e9bedc2c2e9aa9b2d13)


alanmehio:
Will close this. I will first start to send the improvement in a newly created JIRA as a starting point.

PR #748, opened 2021-02-20
CXF-8422: Unclosed input streams after using org.apache.cxf.tools.wsdlto.WSDLToJava (targeting StAX's XMLStreamReader)

reta:
```
    /**
     *<p>
     * Note: as per StAX 1.0 specs, this method does NOT close the underlying
     * input reader. That is, unless the new StAX2 property
     * {@link org.codehaus.stax2.XMLInputFactory2#P_AUTO_CLOSE_INPUT} is
     * set to true.
     */
    @Override
    public void close() throws XMLStreamException 
```

Introduced new system property:
 - `org.apache.cxf.staxutils.autoCloseInputSource`, by default is `false` (preserving the current behavior)
 - when set, the stream behind `InputSource` is going to be closed along with `XMLStreamReader`

### DIFF
--- a/core/src/main/java/org/apache/cxf/staxutils/StaxUtils.java
+++ b/core/src/main/java/org/apache/cxf/staxutils/StaxUtils.java
@@ -121,6 +121,8 @@ public final class StaxUtils {
         "org.apache.cxf.staxutils.innerElementCountThreshold";
     private static final String INNER_ELEMENT_LEVEL_SYSTEM_PROP =
         "org.apache.cxf.staxutils.innerElementLevelThreshold";
+    private static final String AUTO_CLOSE_INPUT_SOURCE_PROP =
+        "org.apache.cxf.staxutils.autoCloseInputSource";
 
     private static final Logger LOG = LogUtils.getL7dLogger(StaxUtils.class);
 
@@ -151,6 +153,7 @@ public final class StaxUtils {
     private static final int PARSER_POOL_SIZE_VAL =
             getInteger("org.apache.cxf.staxutils.pool-size", 20);
     private static final boolean ALLOW_INSECURE_PARSER_VAL;
+    private static final boolean AUTO_CLOSE_INPUT_SOURCE;
 
     // Here we check old names first and then new names for the threshold properties
     private static final int MAX_ELEMENT_DEPTH_VAL =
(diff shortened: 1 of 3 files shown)


reta:
@coheigea @amarkevich mind taking a look guys? one more resource leak in certain circumstances in case of `XMLStreamReader` usage

core/src/main/java/org/apache/cxf/staxutils/StaxUtils.java
```diff
@@ -1660,22 +1667,38 @@ public static XMLStreamReader createXMLStreamReader(InputSource src) {
...
+
             if (src.getEncoding() == null) {
-                StreamSource ss = new StreamSource(src.getByteStream(), sysId);
+                final StreamSource ss = new StreamSource(is, sysId);
                 ss.setPublicId(pubId);
-                return createXMLStreamReader(ss);
+                
+                final XMLStreamReader xmlStreamReader = createXMLStreamReader(ss);
+                if (AUTO_CLOSE_INPUT_SOURCE) {
+                    return new AutoCloseableXMLStreamReader(xmlStreamReader, is);
```
amarkevich, line 1678: AutoCloseable wont close source stream because returning XMLStreamReader cant be defined inside try-with-resources block

reta: That is correct (unless typecasted) but the value here is in how `AutoCloseableXMLStreamReader` defines `close` method: closing the reader and input stream. In my initial attempt, the `AutoCloseableXMLStreamReader` was promoted to return value (covering try-with-resources) but the consequences of closing the `InputSource` streams all the time are not clear to me. It could break things easily.

coheigea (commented):
I have the following failing tests, but maybe they're caused by the Jetty upgrade?

[ERROR]   JAXRSClientServerBookTest.testGetBookByURL:1390->getAndCompareAsStrings:2934->getAndCompare:2956 expected:<200> but was:<400>
[ERROR]   JAXRSClientServerBookTest.testOptions:1427

amarkevich:
@coheigea yes; probably due to https://github.com/eclipse/jetty.project/issues/4275

reta:
@coheigea @amarkevich too often Jetty minor update is a surprise, I will revert it for now, sorry for that guys

reta:
@amarkevich @coheigea thanks for heads up guys, tests should be back to green

coheigea (approved):

PR #751, opened 2021-02-24
CXF-8429: Illegal reflective access in XercesSchemaValidationUtils

reta:
Using `Xerces` as the substitution of the JDK's implementation (sealed even more in JDK16). Sadly, not all features are supported by `Xerces`, fe `http://javax.xml.XMLConstants/property/accessExternalSchema` is not recognizable. 

@ffang could you please take a look? thank you!

### DIFF
--- a/rt/wsdl/pom.xml
+++ b/rt/wsdl/pom.xml
@@ -41,7 +41,6 @@
         <dependency>
             <groupId>org.apache.cxf</groupId>
             <artifactId>cxf-core</artifactId>
-            <version>${project.version}</version>
         </dependency>
         <dependency>
             <groupId>wsdl4j</groupId>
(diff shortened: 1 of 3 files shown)


rt/wsdl/pom.xml
```diff
@@ -119,4 +117,19 @@
...
     </build>
+    
+    <profiles>
+        <profile>
+            <id>xerces</id>
+            <activation>
+                <jdk>[9,)</jdk>
+            </activation>
+            <dependencies>
+                <dependency>
```
reta, line 128: @ffang Another option I've been considering is opt-in model, remove profile and have `xercesImpl` as optional dependency:

```
               <dependency>
                    <groupId>xerces</groupId>
                    <artifactId>xercesImpl</artifactId>
                    <optional>true</optional>
                </dependency>
```

ffang:
@reta I think a jdk9+ profile should be good for now.

Hopefully later xerces-impl release could have more advanced features support to keep up with the one from JDK for external usage, since the com.sun.org.apache.xerces.internal* package really means internal usage for later JDKs

Thanks!
Freeman

reta:
Thanks @ffang , so we are good to merge?

ffang:
@reta yes, I think so

PR #762, opened 2021-03-21
very minor  SonarLint

alanmehio:
SonarLint very minor issue in the core module only.

### DIFF
--- a/core/src/main/java/org/apache/cxf/attachment/AttachmentDeserializer.java
+++ b/core/src/main/java/org/apache/cxf/attachment/AttachmentDeserializer.java
@@ -23,6 +23,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.PushbackInputStream;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
@@ -97,8 +98,6 @@ public class AttachmentDeserializer {
 
     private byte[] boundary;
 
-    private String contentType;
-
     private LazyAttachmentCollection attachments;
 
     private Message message;
@@ -141,7 +140,7 @@ public class AttachmentDeserializer {
     }
 
     protected void initializeRootMessage() throws IOException {
-        contentType = (String) message.get(Message.CONTENT_TYPE);
+        String contentType = (String) message.get(Message.CONTENT_TYPE);
 
         if (contentType == null) {
             throw new IllegalStateException("Content-Type can not be empty!");
@@ -160,7 +159,7 @@ public class AttachmentDeserializer {
             if (null == boundaryString) {
                 throw new IOException("Couldn't determine the boundary from the message!");
             }
-            boundary = boundaryString.getBytes("utf-8");
+            boundary = boundaryString.getBytes(StandardCharsets.UTF_8);
 
             stream = new PushbackInputStream(message.getContent(InputStream.class), PUSHBACK_AMOUNT);
             if (!readTillFirstBoundary(stream, boundary)) {
@@ -189,7 +188,7 @@ public class AttachmentDeserializer {
         }
     }
 
-    private String findBoundaryFromContentType(String ct) throws IOException {
+    private String findBoundaryFromContentType(String ct) {
         // Use regex to get the boundary and return null if it's not found
         Matcher m = CONTENT_TYPE_BOUNDARY_PATTERN.matcher(ct);
         return m.find() ? "--" + m.group(1) : null;
@@ -456,11 +455,8 @@ public class AttachmentDeserializer {
             }
             value = line.substring(separator);
         }
-        List<String> v = heads.get(name);
-        if (v == null) {
-            v = new ArrayList<>(1);
-            heads.put(name, v);
-        }
+        List<String> v = heads.computeIfAbsent(name, k -> new ArrayList<>(1));
+        heads.put(name, v);
         v.add(value);
     }
 
--- a/core/src/main/java/org/apache/cxf/attachment/AttachmentDeserializer.java
+++ b/core/src/main/java/org/apache/cxf/attachment/AttachmentDeserializer.java
@@ -456,7 +456,6 @@ public class AttachmentDeserializer {
             value = line.substring(separator);
         }
         List<String> v = heads.computeIfAbsent(name, k -> new ArrayList<>(1));
-        heads.put(name, v);
         v.add(value);
     }
 
--- a/core/src/main/java/org/apache/cxf/attachment/AttachmentUtil.java
+++ b/core/src/main/java/org/apache/cxf/attachment/AttachmentUtil.java
@@ -75,6 +75,9 @@ public final class AttachmentUtil {
     private static final Random BOUND_RANDOM = new Random();
     private static final CommandMap DEFAULT_COMMAND_MAP = CommandMap.getDefaultCommandMap();
     private static final MailcapCommandMap COMMAND_MAP = new EnhancedMailcapCommandMap();
+    
+    private static final String ERROR_MSG = "The value set as ";
+    private static final String BINARY = "binary";
 
     static final class EnhancedMailcapCommandMap extends MailcapCommandMap {
         @Override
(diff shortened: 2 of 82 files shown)


core/src/main/java/org/apache/cxf/attachment/AttachmentDeserializer.java
```diff
@@ -456,11 +455,8 @@ private void addHeaderLine(Map<String, List<String>> heads, StringBuilder line)
             }
             value = line.substring(separator);
         }
-        List<String> v = heads.get(name);
-        if (v == null) {
-            v = new ArrayList<>(1);
-            heads.put(name, v);
-        }
+        List<String> v = heads.computeIfAbsent(name, k -> new ArrayList<>(1));
+        heads.put(name, v);
```
amarkevich, line 459: additional put is unnecessary

alanmehio: @amarkevich  .  many thanks for the feedback. I actually missed the javadoc for Map#computeIfAbsent  
which states:
 The mostcommon usage is to construct a new object serving as an initialmapped value or memoized result, as in: 
 map.computeIfAbsent(key, k -> new Value(f(k)));

additional put has been removed now. with thanks

core/src/main/java/org/apache/cxf/attachment/AttachmentUtil.java
```diff
@@ -75,6 +75,9 @@
     private static final Random BOUND_RANDOM = new Random();
     private static final CommandMap DEFAULT_COMMAND_MAP = CommandMap.getDefaultCommandMap();
     private static final MailcapCommandMap COMMAND_MAP = new EnhancedMailcapCommandMap();
+    
+    private static final String ERROR_MSG = "The value set as ";
```
amarkevich, line 79: this only a prefix of message. Is it possible to keep it as is? Which check report this?

alanmehio: @amarkevich  . I will revert the changes . It is from SonarLint stating the below
**String literals should not be duplicated (java:S1192)**
Duplicated string literals make the process of refactoring error-prone, since you must be sure to update all occurrences.
On the other hand, constants can be referenced from many places, but only need to be updated in a single place.

I find it before was much better to read the code. Once we add constants String it interrupt the sequence of code reading .

core/src/main/java/org/apache/cxf/attachment/AttachmentUtil.java
```diff
@@ -75,6 +75,8 @@
     private static final Random BOUND_RANDOM = new Random();
     private static final CommandMap DEFAULT_COMMAND_MAP = CommandMap.getDefaultCommandMap();
     private static final MailcapCommandMap COMMAND_MAP = new EnhancedMailcapCommandMap();
+    
+    private static final String BINARY = "binary";
```
amarkevich, line 79: make it package visible and reuse in AttachmentSerializer like BODY_ATTACHMENT_ID

alanmehio: @amarkevich  . This is done with thanks

core/src/main/java/org/apache/cxf/bus/spring/BusEntityResolver.java
```diff
@@ -82,7 +82,7 @@ public InputSource resolveEntity(String publicId, String systemId) throws SAXExc
         String resourceLocation = schemaMappings.get(systemId);
         if (resourceLocation != null && publicId == null) {
             Resource resource = new ClassPathResource(resourceLocation, classLoader);
-            if (resource.exists()) {
+            if (resource.exists() && source != null) {
```
amarkevich, line 85: its better to move null check from L78 to L82

alanmehio: @amarkevich  in this case my check for null will not be needed any more . Done with thanks

core/src/main/java/org/apache/cxf/common/logging/AbstractDelegatingLogger.java
```diff
@@ -119,6 +128,11 @@ public void logp(Level level, String sourceClass, String sourceMethod, String ms
         }
     }
 
+    /**
```
amarkevich, line 131: avoid such unuseful comments

alanmehio: @amarkevich  . Agree . Done with thanks.

alanmehio:
@coheigea  You changes recently into the master branch has broken my merge commit . Please see [this ](https://github.com/apache/cxf/commit/4720f1f2e0da307c729fa6c97e68d49866fe0083) 
Also, I have the latest from upstream master into my branch and I can build locally with success  with the below command
 mvn process-classes compile -Pnochecks -DskipTests  on latest Java 8 from Oracle. Any idea

amarkevich:
@alanmehio fixed: https://github.com/apache/cxf/commit/ca5d02bf9581a28971c8b8d021732972dce5377e

alanmehio:
@amarkevich  There will be no changes to this pull request. I want to move to the next SonarLint level (medium) by opening another JIRA with more specific task to do. Please accept the PR

core/src/main/java/org/apache/cxf/common/util/SpringClasspathScanner.java
```diff
@@ -170,7 +172,7 @@ private static boolean isSpringInOsgi() {
 
     private ResourcePatternResolver getResolver(ClassLoader loader) {
         ResourcePatternResolver resolver = null;
-        if (IN_OSGI) {
+        if (IN_OSGI != null && IN_OSGI.booleanValue()) {
```
amarkevich, line 175: use primitive boolean

core/src/main/java/org/apache/cxf/io/ReaderInputStream.java
```diff
@@ -120,9 +119,9 @@ public ReaderInputStream(Reader reader, CharsetEncoder encoder, int bufferSize)
         this.reader = reader;
         this.encoder = encoder;
         this.encoderIn = CharBuffer.allocate(bufferSize);
-        ((Buffer)this.encoderIn).flip();
+        (this.encoderIn).flip();
```
amarkevich, line 122: flip() do nothing for new allocated buffer

alanmehio: @amarkevich Done with thanks . Good observation

core/src/main/java/org/apache/cxf/common/util/SpringClasspathScanner.java
alanmehio: @amarkevich . Done with thanks.  Great  idea from you.

core/src/main/java/org/apache/cxf/common/util/SpringClasspathScanner.java
```diff
@@ -170,7 +170,8 @@ private static boolean isSpringInOsgi() {
 
     private ResourcePatternResolver getResolver(ClassLoader loader) {
         ResourcePatternResolver resolver = null;
-        if (IN_OSGI) {
+        inOSGI = isSpringInOsgi();
```
amarkevich, line 173: move this call as initial initialization to avoid 'isSpringInOsgi' call for each 'getResolver' invoke

core/src/main/java/org/apache/cxf/io/ReaderInputStream.java
amarkevich: it affect encoderIn buffer too

alanmehio: @amarkevich  . Apology I missed  it Good you have spotted it . Done with thanks

core/src/main/java/org/apache/cxf/common/util/SpringClasspathScanner.java
alanmehio: @amarkevich  The best place is to revert and place a static initializer  since  it is not good idea to put a static method call into the constructor

amarkevich: expected impl:
`private static final boolean IN_OSGI = isSpringInOsgi();`

alanmehio: @amarkevich  done with thanks.

amarkevich (approved):
@alanmehio thanks!

PR #766, opened 2021-04-02
TestUtil: set port for fullName and simpleName only

amarkevich:


gnodet:
Closing this stale PR. The TestUtil port allocation approach here is narrow in scope and hasn't been updated in several years. Thank you for the contribution\!

PR #768, opened 2021-04-02
Cxf 8445

alanmehio:
Sonar rule to reduce complexity of a method will entail more unit tests and code refactoring. 
This is considered a medium level  if there are unit tests ready for the affected method; however,  if unit tests does not exist for the whole class or the methods inside the class which are affected, this will make the complex level to over medium.
The refactor is focused on method bases and is incremental. This means some classes contains more then one methods to be have reduction in cognitive complexity, one method will be taken and the rest will be accomplished in different pull request. 
This pull request will not try to refactor all the complex cognitive methods but rather will do some and once the pull is accepted, it will proceed to next pull and so forth until the core methods has no issue in relation to cognitive complexity

 Sonar rule: java:S3776 reduce Cognitive Complexity of a method; [see CognitiveComplexity](https://www.sonarsource.com/docs/CognitiveComplexity.pdf)

No more changes will be committed for this PR . Since we want each PR  has very min. changes in this area (CognitiveComplexity)


core/src/main/java/org/apache/cxf/attachment/AttachmentUtil.java
```diff
@@ -170,18 +170,31 @@ public static boolean isMtomEnabled(Message message) {
...
+    
+    private static void setOutputDirectory(Object directory, CachedOutputStream bos) throws IOException {
         if (directory != null) {
             if (directory instanceof File) {
-                bos.setOutputDir((File) directory);
+                bos.setOutputDir((File)directory);
             } else if (directory instanceof String) {
-                bos.setOutputDir(new File((String) directory));
+                String directoryStr  = (String)directory;
+                directoryStr = directoryStr.replace("..", "");
```
coheigea, line 188: How is this change related to the overall PR?

alanmehio: @coheigea  you are right. This has nothing to do with PR. The codeQL was failing [Java path Injection rule](https://codeql.github.com/codeql-query-help/java/java-path-injection/)
Can somebody looks at the codeQL it is disturbing and causing the build to fail .  I will revert the changes and again this has nothing to do with PR since the build was failing due to CodeQL check on java path Injection.  If the CodeQL can be minimized to something realistic which does not affect code , it would be great

alanmehio:
The check is failed now due to CodeQL  rule which force us to check the directory name not to contain ".."   Any help would be appreciated to relax the codeQL rule.  I are doing refactoring and never touch the business logic.

PR #770, opened 2021-04-07
[CXF-8450] Creates a CachedJAXBDataBinding

WoodenAndrew:
Creates a reordered JAXBDataBinding implementation that permits expanding internal class list


WoodenAndrew:
Here we go. It got picked up by Jira.

PR #771, opened 2021-04-07
CXF-8442 Java 16 & CachedOutputStream Fix IOException closed after multiple close() & postClose() calls 

Sevyls:
see https://issues.apache.org/jira/browse/CXF-8442

### DIFF
--- a/core/src/main/java/org/apache/cxf/io/CacheAndWriteOutputStream.java
+++ b/core/src/main/java/org/apache/cxf/io/CacheAndWriteOutputStream.java
@@ -29,7 +29,7 @@ import java.io.OutputStream;
  *
  */
 public class CacheAndWriteOutputStream extends CachedOutputStream {
-    private boolean isClosed = true;
+    private boolean isClosed = false;
     OutputStream flowThroughStream;
     long count;
     long limit = Long.MAX_VALUE;
--- a/core/src/main/java/org/apache/cxf/io/CacheAndWriteOutputStream.java
+++ b/core/src/main/java/org/apache/cxf/io/CacheAndWriteOutputStream.java
@@ -29,7 +29,7 @@ import java.io.OutputStream;
  *
  */
 public class CacheAndWriteOutputStream extends CachedOutputStream {
-
+    private boolean isClosed = true;
     OutputStream flowThroughStream;
     long count;
     long limit = Long.MAX_VALUE;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/cxf/io/CachedOutputStream.java
```diff
@@ -225,7 +225,11 @@ public void close() throws IOException {
             ciphers.clean();
         }
         if (!maybeDeleteTempFile(currentStream)) {
-            postClose();
+            try {
```
reta, line 228: @Sevyls thank you for the investigation and the consequent suggested fix. I think we should attack the problem from the different angle (as you also mentioned in the ticket):
 - `postClose()` could raise `IOException` so does `close()` and this is legitimate, no need to catch it
 - however, as you found out, `CacheAndWriteOutputStream` may attempt to close the underlying stream many times
I think the proper solution would be:
 - `CacheAndWriteOutputStream`  should keep track if the `flowThroughStream` has been closed already 
 - `closeFlowthroughStream()` and `postClose()` should set or/and check if the `flowThroughStream` has been closed already and behave appropriately (flush/close or do nothing)

What do you think?

Sevyls:
Hi @reta,

Thanks for your answer and suggestions. I reverted my changes in `CacheAndWriteOutputStream`.

I added a variable isClosed to `CacheAndWriteOutputStream` and as `postClose()` and `closeFlowthroughStream()` had the same implementation I reduced it to a call to `postClose()`. 

Please review my pull request again, see the test case `CacheAndWriteOutputStreamTest#testCloseMultipleTimes()`.

core/src/main/java/org/apache/cxf/io/CacheAndWriteOutputStream.java
```diff
@@ -47,19 +47,32 @@ public void setCacheLimit(long l) {
...
     }
 
     public OutputStream getFlowThroughStream() {
         return flowThroughStream;
     }
 
+    @Override
+    protected void doClose() throws IOException {
+        super.doClose();
+        isClosed = true;
```
reta, line 68: Just for the record, probably not necessary (since `close()` would call `postClose()`) but won't hurt

core/src/main/java/org/apache/cxf/io/CacheAndWriteOutputStream.java
```diff
@@ -29,7 +29,7 @@
  *
  */
 public class CacheAndWriteOutputStream extends CachedOutputStream {
-
+    private boolean isClosed = false;
```
reta, line 32: The stream should not be shared between different threads, so we may not need `volatile` or atomic

reta (approved):

reta:
@Sevyls Thank you.

core/src/main/java/org/apache/cxf/io/CacheAndWriteOutputStream.java
reta: @Sevyls I had to remove `close()` and `doClose()`, the complex relationships between these methods led to the issues that `isClosed` flag being set **before** closing the streams. As the results, streams stayed open.

PR #775, opened 2021-04-11
CXF-8344: org.apache.cxf.jaxrs.utils.AnnotationUtils#getNameBindings does not use ClassUnwrapper

reta:
I also added a couple of tests for `@NameBinding` and custom CDI scopes (which would be using `PerRequestResourceProvider` invocation), it turned out we had none.

@rmannibucau mind taking a look please?  thanks a lot!

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/JAXRSServerFactoryBean.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/JAXRSServerFactoryBean.java
@@ -100,11 +100,10 @@ public class JAXRSServerFactoryBean extends AbstractJAXRSFactoryBean {
 
     public void setApplicationInfo(ApplicationInfo provider) {
         appProvider = provider;
-        Set<String> appNameBindings = AnnotationUtils.getNameBindings(provider.getProvider()
-                                                                      .getClass().getAnnotations());
+        Set<String> appNameBindings = AnnotationUtils.getNameBindings(bus, provider.getProvider().getClass());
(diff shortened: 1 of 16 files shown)


rmannibucau:
@reta not sure I get the custom scope impl and need (@ApplicationScoped or @RequestScoped should do he same and properly impl get(bean)) but otherwise looks good

reta:
> ... @ApplicationScoped or @RequestScoped should do he same and properly impl get(bean)) but otherwise looks good

Aha, so thought but here are some findings, in `JAXRSCdiResourceExtension`:
- we create `PerRequestResourceProvider` when `isCxfSingleton(beanManager, bean)` is `false`
-  in turn, `isCxfSingleton` checks the scope
    ```
    boolean isCxfSingleton(final BeanManager beanManager, final Bean<?> bean) {
        return beanManager.isNormalScope(bean.getScope()) || isConsideredSingleton(bean.getScope());
    }
    ```
- and `beanManager.isNormalScope(bean.getScope())` is `true` for all pre-predefined `@XxxScoped` scopes  because each of those has `@NormalScope` annotation as well

So I introduced the custom scope to trigger the `PerRequestResourceProvider`  flow. What do you think, see any issues with `isCxfSingleton` implementation?

rmannibucau:
@reta hmm, both case are interesting but my point was that normal scoped beans are required to be proxied whereas others not. That said it also depends how CDI beans are proxies so maybe saner to test with a custom fake proxy not copying annotation since I suspect several CDI impl will copy them to not break frameworks, wdyt?

reta:
@rmannibucau gotcha, yeah sure, I will add yet another test case to make sure proxies are involved, thanks!

PR #778, opened 2021-04-14
[CXF-7996] Fix `@Encode` TCK test

andymc12:
This should fix the following Jakarta RESTful WS TCK failures:
> com/sun/ts/tests/jaxrs/ee/rs/beanparam/form/plain/JAXRSClient.java#formFieldParamEntityWithEncodedTest_from_standalone: JAXRSClient_formFieldParamEntityWithEncodedTest_from_standalone

> com/sun/ts/tests/jaxrs/ee/rs/beanparam/form/plain/JAXRSClient.java#formParamEntityWithEncodedTest_from_standalone: JAXRSClient_formParamEntityWithEncodedTest_from_standalone

The issue is that (prior to this fix) form parameters are read into a map as either decoded or encoded based on whether the first parameter has the `@Encoded` annotation or not.  This can result in the `@Encoded` annotation on second or subsequent parameters getting ignored.

The fix is to read the parameters into the map without performing any decoding until it is determined which parameters should be decoded.

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/JAXRSUtils.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/JAXRSUtils.java
@@ -1029,15 +1029,15 @@ public final class JAXRSUtils {
         MultivaluedMap<String, String> params =
             (MultivaluedMap<String, String>)m.get(FormUtils.FORM_PARAM_MAP);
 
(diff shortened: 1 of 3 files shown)


reta (commented):
Got the problem now, thanks a lot for the fix @andymc12 !

reta (approved):

andymc12:
Should resolve https://issues.apache.org/jira/browse/CXF-8461 and https://issues.apache.org/jira/browse/CXF-8462

dblevins:
Thank you @andymc12 !

PR #791, opened 2021-05-04
CXF-8536 Conditional log for NamespaceMapper not found

bergerst:
Only logs the message "Could not create a NamespaceMapper ..." if it really is the case


reta (approved):

reta:
Thank you @bergerst

PR #793, opened 2021-05-06
Remove String.replaceAll JAXRSUtils performance fix

dettmoney:
It's much better to use substring to remove the last '/' character than replaceAll from a performance perspective. Certain applications can see up to 2% throughput gains with this change.


andymc12 (approved):
LGTM - Thanks @dettmoney!

ffang (approved):

PR #801, opened 2021-05-21
[CXF-8540] Fix logging event type of CORS preflight requests

borcsokj:


rt/features/logging/src/main/java/org/apache/cxf/ext/logging/event/DefaultLogEventMapper.java
```diff
@@ -320,7 +322,7 @@ public EventType getEventType(Message message) {
      */
     private boolean isRESTFault(Message message) {
         Object opName = message.getExchange().get("org.apache.cxf.resource.operation.name");
-        if (opName == null) {
+        if (opName == null && !PREFLIGHT_PASSED.equals(message.getExchange().get(CORS_FILTER))) {
```
reta, line 325: Thanks for the PR @borcsokj , I think your solution relies heavily on internals of `CrossOriginResourceSharingFilter` but should not. By and large, anyone could come up with own CORS filter implementation and  `DefaultLogEventMapper` has to work as expected.

I would suggest to rethink the approach and combine the checks of `opName` and `responseCode` (which should be available). In this case, even if operation is `null`, the `responseCode` would help to distinguish fault from non-fault, what do you think? Thank you.

borcsokj: My first approach was checking responseCode (if available) even if operation is empty but I found "For REST we also consider a response to be a fault if the operation is not found" in JavaDoc of the method. I think logging should be separated for internals of other components (i.e. CORS) as you wrote but I don't know the reason why REST message missing opName is marked as fault.

borcsokj: I pushed an update removing CORS filter internals from `DefaultLogEventMapper`, what do you think about it @reta ?

reta:
Thank you @borcsokj , LGTM, could you please target `master` branch so we could run all the checks and backport it to 3.4.x / 3.3.x, thank you.

borcsokj:
Thank you @reta , I updated target to `master`

reta:
Thank you @borcsokj , probably re-targeting needs rebase, I see a large number of conflicts reported

PR #805, opened 2021-05-28
Make CDI and MP Config dependencies optional for MicroProfile REST client

brevilo:
* The Rest Client for MicroProfile specification defines CDI and MP Config as optional
* Analogous upstream change: https://github.com/eclipse/microprofile-rest-client/pull/309

Cheers

### DIFF
--- a/rt/rs/microprofile-client/pom.xml
+++ b/rt/rs/microprofile-client/pom.xml
@@ -134,15 +134,20 @@
         <dependency>
             <groupId>org.eclipse.microprofile.config</groupId>
             <artifactId>microprofile-config-api</artifactId>
+            <scope>provided</scope>
+            <optional>true</optional>
         </dependency>
         <dependency>
             <groupId>jakarta.annotation</groupId>
             <artifactId>jakarta.annotation-api</artifactId>
+            <scope>provided</scope>
+            <optional>true</optional>
         </dependency>
         <dependency>
             <groupId>jakarta.enterprise</groupId>
             <artifactId>jakarta.enterprise.cdi-api</artifactId>
             <scope>provided</scope>
+            <optional>true</optional>
         </dependency>
         <dependency>
             <groupId>cglib</groupId>


rmannibucau:
Hi @brevilo I'd be to make it really optional, including in https://github.com/apache/cxf/blob/master/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/config/ConfigFacade.java. See https://github.com/apache/geronimo-openapi/blob/master/geronimo-openapi-impl/src/main/java/org/apache/geronimo/microprofile/openapi/config/GeronimoOpenAPIConfig.java for an example

brevilo:
Hi, can you please rephrase "I'd be to make it really optional" again? It seems to be missing a word, I think.

Thanks

brevilo:
That said, if there's a better way to achieve the same, I'd be happy with it as well. What I care about is that adding the REST client to project doesn't make CDI or MP Config a hard dependency.

brevilo:
Thanks for your patience ;) Not sure I follow, though. Doesn't [this](https://github.com/apache/cxf/blob/master/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/config/ConfigFacade.java) show that MP Config is already optional, at least code-wise? BTW, the same appears to be the case for the client's [CDI support](https://github.com/apache/cxf/blob/master/rt/rs/microprofile-client/src/main/java/org/apache/cxf/microprofile/client/cdi/CDIFacade.java).

That's presumably the reason I could successfully compile `cxf-rt-rs-mp-client` after making said POM dependencies optional, isn't it? If so, it seems that the mandatory POM dependencies were in fact just a remnant, although the code already requires them only optionally. What am I missing?

rmannibucau:
@brevilo the facade relies on Config type so I don't think so, this is why I think a test is needed too.

andymc12:
@rmannibucau fwiw, the ConfigFacade works in Open Liberty with and without MP Config.

That said, it should be possible to create a new test project that runs a small portion of the TCK - and that project would not include CDI or MP Config dependencies.  I think that would demonstrate that CDI and Config are truly optional.

rmannibucau:
@andymc12 without an additional class indirection it depends the deployment - even in liberty - so is unsafe. SecurityManager, javaagents, loader instrumentations are as much sources of issues without one more class in between. So agree it works sometimes but think the point remains.

brevilo:
Sorry, I don't have enough context knowledge to provide the additional tests needed. The overall goal here is to make CDI and MP Config optional for the MP REST client, as the spec mandates ([confirmed](https://github.com/eclipse/microprofile-rest-client/issues/308#issuecomment-806057762) by @andymc12). How do we best proceed to achieve this?

rmannibucau:
@brevilo guess adding tests - security manager, javaagent visiting the classes and loading visited types - to harness it is the best way to ensure it works.

side note: optional=true is not sufficient in the pom, scope=provided is what makes it optional for consumers.

brevilo:
~~Would it be better if I close this PR and turn it into an issue?~~ Nope, there are no public issues.

rmannibucau:
There are at https://issues.apache.org/jira/projects/CXF/issues

brevilo:
Thanks for the pointer. Might be the better/quicker way to get the underlying issue resolved. Do you agree?

andymc12:
Hi @brevilo - I'm on holiday today and tomorrow, but I can look at adding a
test case when I get back.  Thanks!

brevilo:
Teamwork! Sounds great 👍

andymc12:
@brevilo @rmannibucau I just merged #808 which tests that the CXF MP Rest Client works even when no CDI or MP Config is present on the classpath.  That shows that it should work at runtime.  Given those tests, I think it is ok to merge these changes as well. Wdyt?

andymc12 (approved):

rmannibucau:
@andymc12 I spoke of surefire excludes cause it harnesses "in force mode" the exclusion. Your test without mp-config uses mp-config due to the grand-parent pom which imports geronimo-config-impl. The "check it is not there" pattern is wrongly written:

        try {
            Class.forName("org.eclipse.microprofile.config.ConfigProvider");
            fail("ConfigProvider API class is on the classpath - results from this test project are not valid");
        } catch (Throwable t) {
            // expected
        }

If it fails the AssertionError is caught by the catch and silently ignored. Moving from Throwable to ClassNotFoundException would be sufficient to start seeing the test failing.

andymc12:
@rmannibucau Thanks for catching that.  I have a fix in PR #810 - would you mind reviewing?

brevilo:
Hi guys. Since #810 got merged, how do we progress with this PR?

rmannibucau:
Guess scope should be adjusted to provided and voilà?
It is a breaking change for consumers so we should pby ask ourselves if a bom is needed or not (no strong opinion on that).

brevilo:
Hm, not sure. `jakarta.enterprise.cdi-api` already had a scope of `provided` and I don't think that worked, hence the `optional` property. Doesn't `provided` imply that the runtime will provide the dependency? That's exactly the opposite of what I'm trying to achieve here. IOW, the runtime container does not and should not provide the purely optional dependency. Any dependency resolver/validator (like `bnd`) must not fail if those optional dependencies are "missing".

rmannibucau:
@brevilo provided is the way to say it is optional with maven, optional=true means it can be there or not and if i'm not mistaken was just introduced to please some tool but its usage is not unified whereas provided is clearly there to mean "the project is not the one responsible to handle the presence of the dependency" which is exactly what you are trying to achieve. To avoid surprises I would use provided and fix the maven bundle instructions in the plugin/related properties.

brevilo:
Ok, I just tested it again with `provided` instead of `optional`. It doesn't work:
```
! could not resolve the bundles: [org.apache.cxf.cxf-rt-rs-mp-client-3.5.0.SNAPSHOT org.osgi.framework.BundleException: Could not resolve module: org.apache.cxf.cxf-rt-rs-mp-client [170]
  Unresolved requirement: Import-Package: javax.enterprise.context; version="[2.0.0,3.0.0)"
```

rmannibucau:
@brevilo yes, this is why I said you must fix the OSGi stuff. If it is easier for you just set provided and optional. Point is that this issue is about making it provided (ie the user ensures it is there at runtime) by default and not "optional".

brevilo:
> If it is easier for you just set provided and optional

Sure, if that satisfies all requirements :) [Apparently](https://github.com/apache/cxf/blob/master/rt/rs/microprofile-client/pom.xml#L56), others went that route before us as well! Encouraging!

I checked it and it works for me...

PR updated 👍

rmannibucau (approved):

brevilo:
Thanks 🥳 What's needed to see this merged?

brevilo:
@reta, do you want to chime in on this (final) one as well?

rt/rs/microprofile-client/pom.xml
```diff
@@ -134,15 +134,20 @@
         <dependency>
             <groupId>org.eclipse.microprofile.config</groupId>
             <artifactId>microprofile-config-api</artifactId>
+            <scope>provided</scope>
```
reta, line 137: @brevilo @rmannibucau Guys, I think `<scope>provided</scope>` is redundant here,  `<optional>true</optional>` is suffice by all means. Why:
 - `<scope>provided</scope>` means that dependency is **required** and has to be provided at runtime [1]
 - `<optional>true</optional>` means dependency is **not required** but for some features only [2]
 
[1] https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html
[2] https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html

rmannibucau: True until you use either a plugin which uses only one or a scraper/lib/tool which reads the pom and only uses one.
Strictly speaking (maven), provided means "will be there at compile time but not transitively", nothing more (rest is wording) and it was originally intended for specs (like here funnily) but optional was added (for OSGi I guess) to do the same so as of today, if the pom is consumed (and it is for CXF) I guess the best is to have both (minimal being provided which is the most supported one I guess?).

reta: Fair enough, let's go with both than :)

reta (approved):

brevilo:
Hooray! I take this could be released in 3.3 and/or 3.4 in September?

brevilo:
> I take this could be released in 3.3 and/or 3.4 in September?

Background: just wondering since @reta added #806 to the 3.3 and 3.4 branches right away 😇

reta:
@brevilo backporting right now :-)

brevilo:
Woohoo 🥳

(Didn't mean to be impatient)

PR #806, opened 2021-05-28
MicroProfile REST client: enable SPI/ServiceLoader under OSGi 

brevilo:
* Using OSGi's [Service Loader Mediator](https://blog.osgi.org/2013/02/javautilserviceloader-in-osgi.html) (i.e. Apache Aries [SPI Fly](https://aries.apache.org/documentation/modules/spi-fly.html))
* Made all requirements optional (minimally invasive change)
* ~~Removing `Bundle-ActivationPolicy` header _locally_ as it prevents the OSGi ServiceLoader Mediator (SPI Fly) to work when set to [`Lazy`](https://github.com/apache/cxf/blob/master/parent/pom.xml#L653) (in the parent POM). I committed this separately should there be a better solution.~~

Based on the following discussion: https://github.com/eclipse/microprofile-rest-client/issues/308#issuecomment-813638983

Cheers

### DIFF
--- a/rt/rs/microprofile-client/pom.xml
+++ b/rt/rs/microprofile-client/pom.xml
@@ -185,6 +185,25 @@
           <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <configuration>
+                    <instructions>
+                        <Require-Capability>osgi.extender; filter:="(osgi.extender=osgi.serviceloader.registrar)"; resolution:=optional</Require-Capability>
+                        <Provide-Capability>
+                            osgi.serviceloader; osgi.serviceloader=org.eclipse.microprofile.rest.client.spi.RestClientBuilderResolver;
+                            register:=org.apache.cxf.microprofile.client.spi.CxfRestClientBuilderResolver
+                        </Provide-Capability>
+                    </instructions>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
     <profiles>
         <profile>
             <id>setup.eclipse</id>


andymc12 (approved):
This looks good to me, but I'd like to leave the PR open at least over the long (US) weekend in case others want to review it.  Thanks @brevilo!

rt/rs/microprofile-client/pom.xml
```diff
@@ -185,6 +185,26 @@
...
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <configuration>
+                    <instructions>
+                        <_removeheaders>Bundle-ActivationPolicy</_removeheaders>
```
reta, line 196: It should be possible to use `cxf.bundle.activator` in order to not populate the manifest:
```
<properties
    <cxf.bundle.activator />
</properties>
```

brevilo: Hm, where would these properties belong? I'd like to test if that alternative works as you suggest.

Thanks

reta: Thanks @brevilo, you could find these properties in https://github.com/apache/cxf/blob/master/parent/pom.xml#L62

brevilo: Sorry, I don't follow yet. Should I remove the properties from the main POM you referenced? My original approach was meant to constrain the change's impact to the POM in question only.

reta: My bad @brevilo , apologies, should explain it in more details. Apache CXF uses `cxf.bundle.activator` property (which is declared in `parent/pom.xml` and (re)defined in `microprofile-client/pom.xml` (https://github.com/apache/cxf/blob/master/rt/rs/microprofile-client/pom.xml#L44). If you change it to empty value (or just remove it from `pom.xml` altogether), the `Bundle-ActivationPolicy` manifest entry won't be included and you don't need to do that:
```
 <_removeheaders>Bundle-ActivationPolicy</_removeheaders>
```

Since the activator would be gone, it is very likely the exported services should need to be changed as well (may be not, unsure about it but I could look), so this where the second property, `cxf.export.service` (https://github.com/apache/cxf/blob/master/rt/rs/microprofile-client/pom.xml#L41) comes handy.

Hope it helps, please let me know if you still have questions, thanks.

brevilo: Thanks @reta, I totally missed "(re)defined in `microprofile-client/pom.xml`" before, hence my confusion.

Anyhow, your approach doesn't work. If I remove only `cxf.bundle.activator` I get an exception that still involves `org.apache.cxf.bus.osgi.CXFActivator`. If I also remove `cxf.export.service` the exception is gone but `MANIFEST.MF` still contains `Bundle-ActivationPolicy: lazy` which is the root cause of the problem I'm trying to solve here: lazy activation just doesn't work for SPI/ServiceLoader.

Since Aries Blueprint is used I had a look at how it might deal with the bundle activation policy. Unfortunately [that part](https://aries.apache.org/documentation/modules/blueprint.html#_lazy_and_eager_activiation) is missing from the documentation :-/ Any idea?

reta: Oh right, for some reasons I messed up the `Bundle-ActivationPolicy` and `Bundle-Activator`. Indeed the policy is hard-coded and these properties won't help. Yeah, the Aries documentation is not too helpful in these regards, we could introduce the new property to manage activation policy but I found this (quite old) issue in the tracker https://issues.apache.org/jira/browse/CXF-6347 which clarifies why `Bundle-ActivationPolicy` was set to `lazy`. Mind if I take some time to look more into the problem with `Bundle-ActivationPolicy`? You are deploying in Apache Karaf, correct? Thank you.

brevilo: > You are deploying in Apache Karaf, correct?
Yes, eventually. Right now I'm testing within Eclipse using bndtools.

reta: Hey @brevilo , sorry for the delay, but I managed to look at the issue with activation policy. There are a few interesting findings with respect to bundle activation policy and Aries SPI Fly. I will try to summarize them here, my apologies if I repeat the things you have already mentioned / knew about:

> Lazy Activation Policy: A lazy activation policy indicates that the bundle, once started, must not be activated until it receives the first request to load a class. (https://docs.osgi.org/specification/osgi.core/8.0.0/framework.lifecycle.html#i3270439)

Basically, the important thing in our context is `... once started ... ` because Aries SPI Fly will be tracking all bundles in STARTING or ACTIVE (started) state: https://github.com/apache/aries/blob/trunk/spi-fly/spi-fly-core/src/main/java/org/apache/aries/spifly/BaseActivator.java#L97. From all the tests I have done (I have used Apache Karaf 4.3.2), the consumer was able to properly find `org.eclipse.microprofile.rest.client.spi.RestClientBuilderResolver` service(s) once the `cxf-rt-rs-mp-client` is started / starting. 

May be you could share the sample project to look at with respect to lazy / eager activation policy? In any case, I have added `cxf.bundle.activation.policy` to allow activation policy overrides: https://github.com/apache/cxf/pull/827.

Thank you.

brevilo: > From all the tests I have done (I have used Apache Karaf 4.3.2), the consumer was able to properly find `org.eclipse.microprofile.rest.client.spi.RestClientBuilderResolver` service(s) once the `cxf-rt-rs-mp-client` is started / starting.

Just to avoid misunderstandings, did your tests involve any changes to CXF (like using the properties added in #827) or did it work for you out of the box? IOW, could you reproduce my original problem at all?

Providing a sample will take some time as my current project is embedded in a more complex environment. Yet I'll test using the property introduced in #827 soon. Looks like setting that to `eager` should do the trick. If it does, do you want me to amend my PR and use that property instead of my header removal? Would be great to get rid of the CXF fork I have to use right now :)

Thanks!

reta: > Just to avoid misunderstandings, did your tests involve any changes to CXF (like using the property added in #827) or did it work for you out of the box? IOW, could you reproduce my original problem at all?

I pulled your PR as-is, created a sample bundle to consume `RestClientBuilderResolver` and played w/o `lazy` `cxf-rt-rs-mp-client` bundle activation policy. I was not able to reproduce your original problem, once the bundle is started, the service loader mechanism kicks in (with `lazy` and `eager`). 

> Looks like setting that to eager should do the trick. 

Just leave it empty, like `<cxf.bundle.activation.policy />`, it will be equivalent to eager activation. 

>  If it does, do you want me to amend my PR and use that property instead of my header removal? 

In general, yes, but it would be great to have reproducer for the problem, to understand why it happens. If it is easier for you, we could get capabilities merged (it is good addition) and troubleshoot the bundle activation later on. 

> Would be great to get rid of the CXF fork I have to use right now :)
Certainly eager to help you here. 

Thanks!

brevilo: > I pulled your PR as-is, created a sample bundle to consume `RestClientBuilderResolver` and played w/o `lazy` `cxf-rt-rs-mp-client` bundle activation policy.

Sorry for being pedantic here but we need to be clear on this: did you pull my PR and *then* tested the loader mechanism? It's of course expected to work that way. Instead of pulling the whole PR you'd want to cherry-pick 2161767 only.

> Just leave it empty, like `<cxf.bundle.activation.policy />`, it will be equivalent to eager activation.

Ok, tested the new property and it works as expected (prevent the header from being added in the first place). I'd now go ahead and amend 3b1df04 accordingly.

> In general, yes, but it would be great to have reproducer for the problem, to understand why it happens.

See above. `lazy` activation simply means that the bundle is never STARTING or ACTIVE, so SPI Fly won't find it. As soon as you use `eager` (which is the default anyway) the bundle is started and SPI Fly can do its job.

> If it is easier for you, we could get capabilities merged (it is good addition) and troubleshoot the bundle activation later on.

Doesn't really help since I need both 🤷‍♂️ 

Cheers

brevilo: Darn! I found the culprit! I had a typo in my start level override for `cxf-rt-rs-mp-client`, so it had the same as `org.apache.aries.spifly.dynamic.bundle`. While the former always has a higher ID, I suspect it was started after the latter, so this shouldn't matter. Since that isn't assured I had increased the start level - but had a typo in there. Now that I fixed that blunder the whole activation policy stuff isn't needed anymore. The provider is found just fine. Duh!

I removed the activation header commit from this PR. Sorry for your time I wasted. 🤦‍♂️

reta: Awesome, thank you @brevilo!

reta (approved):

brevilo:
Yeah, finally 🥳 Thanks everyone!

PR #808, opened 2021-06-01
Run Async systests with no CDI or MP Config

andymc12:
This PR re-runs the MP Rest Client async tests but without using CDI or MP Config.  This is intended to verify that all of the dependencies on CDI and MP Config truly are optional, and it should address some of the concern raised in PR #805.

### DIFF
--- a/systests/microprofile/client/nocdi/pom.xml
+++ b/systests/microprofile/client/nocdi/pom.xml
@@ -1,0 +1,136 @@
+<?xml version="1.0"?>
+<!--
+  ~  Licensed to the Apache Software Foundation (ASF) under one
+  ~  or more contributor license agreements.  See the NOTICE file
+  ~  distributed with this work for additional information
+  ~  regarding copyright ownership.  The ASF licenses this file
+  ~  to you under the Apache License, Version 2.0 (the
+  ~  "License"); you may not use this file except in compliance
+  ~  with the License.  You may obtain a copy of the License at
+  ~
+  ~  http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~  Unless required by applicable law or agreed to in writing,
+  ~  software distributed under the License is distributed on an
+  ~  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  ~  KIND, either express or implied.  See the License for the
+  ~  specific language governing permissions and limitations
+  ~  under the License.
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <parent>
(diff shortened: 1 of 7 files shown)


systests/microprofile/client/nocdi/src/test/java/org/apache/cxf/systest/microprofile/rest/client/AsyncMethodTest.java
```diff
@@ -0,0 +1,239 @@
...
+import org.eclipse.microprofile.rest.client.tck.providers.TestMessageBodyReader;
+import org.eclipse.microprofile.rest.client.tck.providers.TestMessageBodyWriter;
+import org.eclipse.microprofile.rest.client.tck.providers.TestParamConverterProvider;
+import org.eclipse.microprofile.rest.client.tck.providers.TestReaderInterceptor;
+import org.eclipse.microprofile.rest.client.tck.providers.TestWriterInterceptor;
+
+import org.junit.Rule;
+import org.junit.Test;
+
+//CHECKSTYLE:OFF
```
reta, line 54: Minor, `//CHECKSTYLE:OFF` probably is not needed

systests/microprofile/client/nocdi/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <parent>
+        <artifactId>cxf-microprofile-tck</artifactId>
+        <groupId>org.apache.cxf.systests</groupId>
+        <version>3.5.0-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.apache.cxf.systests</groupId>
+    <artifactId>cxf-systests-microprofile-async</artifactId>
```
reta, line 29: Please `cxf-systests-microprofile-nocdi` or `cxf-systests-microprofile-async-nocdi` , we already have `cxf-systests-microprofile-async`

andymc12: Good catch - I copied/pasted from the async tests and missed that line. Should be fixed now.

systests/microprofile/client/nocdi/src/test/java/org/apache/cxf/systest/microprofile/rest/client/AsyncMethodTest.java
andymc12: It didn't like the wildcard import, but there were only 4-5 methods to import, so I added them explicitly and removed the checkstyle comments. Thanks.

reta: Gotcha, thanks!

reta (approved):

PR #809, opened 2021-06-02
CXF-8387: Allow special characters to be used for a queryParam name.

reta:
Allow special characters to be used for a queryParam name.


andymc12 (approved):
LGTM - thanks @reta!

PR #810, opened 2021-06-03
Ensure test runs without CDI/MP Config

andymc12:
From https://github.com/apache/cxf/pull/805#issuecomment-853608507 :
> @andymc12 I spoke of surefire excludes cause it harnesses "in force mode" the exclusion. Your test without mp-config uses mp-config due to the grand-parent pom which imports geronimo-config-impl. The "check it is not there" pattern is wrongly written:
>
>    try {
>        Class.forName("org.eclipse.microprofile.config.ConfigProvider");
>        fail("ConfigProvider API class is on the classpath - results from this test project are not valid");
>    } catch (Throwable t) {
>        // expected
>    }
>If it fails the AssertionError is caught by the catch and silently ignored. Moving from Throwable to ClassNotFoundException would be sufficient to start seeing the test failing.

@rmannibucau you are absolutely correct.  Good catch!  

In this PR, I'm adding the surefire exclude and changing the code to catch `ClassNotFoundException` instead of `Throwable`.

Thanks!


rmannibucau:
Hi @andymc12 ,

Seems it works thanks type erasure (Optional<Config> -> Optional).

Also tested with a patched sirona version to ensure it was compliant with a javaagent rewriting classes. If of any interest here is what I tested:

. JVM args:

    -javaagent:/opt/rmannibucau/dev/sirona/agent/javaagent/target/sirona-javaagent-0.8-SNAPSHOT-shaded.jar=environment-debug=true|debug=true

. sirona.properties:
    com.github.rmannibucau.sirona.javaagent.listener.CounterListener.includes=prefix:org.apache.cxf.microprofile.client.config.ConfigFacade
    com.github.rmannibucau.sirona.javaagent.listener.CounterListener.excludes=container:jvm
    com.github.rmannibucau.sirona.javaagent.path.tracking.activate=false
    com.github.rmannibucau.sirona.logging.counter.clearOnCollect=false
    
So it looks like it works but I think I still prefer a mp-config indirection option - ConfigFacade impl could be bound in the bus as extension - which will avoid to go through an exception each time a config is read org.apache.cxf.microprofile.client.config.ConfigFacade#config and rely on type erasure.

Style side note - I saw it reviewing your PR so thought I could share it:

   Optional<Config> c = config();
    return c.isPresent() ? c.get().getOptionalValue(propertyName, clazz) : Optional.empty();

should likely be (in most config accessors)

    return config().flatMap(c -> c.getOptionalValue(propertyName, clazz));

PR #812, opened 2021-06-08
[CXF-8544] JAX-RS server-side SSE using POST

timothyjward:
Create a failing test to show that SSE doesn't work for POST resource methods

### DIFF
--- a/systests/rs-sse/rs-sse-base/src/main/java/org/apache/cxf/systest/jaxrs/sse/AbstractSseTest.java
+++ b/systests/rs-sse/rs-sse-base/src/main/java/org/apache/cxf/systest/jaxrs/sse/AbstractSseTest.java
@@ -18,6 +18,12 @@
  */
 package org.apache.cxf.systest.jaxrs.sse;
 
+import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -26,16 +32,19 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 
+import javax.ws.rs.client.Entity;
 import javax.ws.rs.client.WebTarget;
 import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
+import javax.ws.rs.ext.MessageBodyReader;
 import javax.ws.rs.sse.InboundSseEvent;
 import javax.ws.rs.sse.SseEventSource;
 import javax.ws.rs.sse.SseEventSource.Builder;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.jaxrs.json.JacksonJsonProvider;
 
 import org.junit.Before;
 import org.junit.Test;
@@ -45,7 +54,10 @@ import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.hasItems;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
 
 public abstract class AbstractSseTest extends AbstractSseBaseTest {
     @Before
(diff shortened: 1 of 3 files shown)


timothyjward:
Are there any comments about this PR? It would be really helpful to see the problem fixed.

rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/interceptor/SseInterceptor.java
```diff
@@ -73,16 +73,6 @@ public void handleMessage(Message message) {
                     servletResponse = (HttpServletResponse)response;
                     builder = Response.status(servletResponse.getStatus());
                     
-                    @SuppressWarnings("unchecked")
```
reta, line 76: The headers are included into the response for a reason, for example, at last these two would be lost.

```
< Host: localhost:8686
< User-Agent: curl/7.71.1
```

In general, CXF uses `PROTOCOL_HEADERS` to store additional headers, which could be mutated and returned in the response. It would be good to know what exactly is the issue with adding headers to response? Any particular headers are problematic?

systests/rs-sse/rs-sse-base/src/main/java/org/apache/cxf/systest/jaxrs/sse/AbstractSseBaseTest.java
```diff
@@ -58,7 +58,7 @@ protected WebClient createWebClient(final String url) {
     protected WebTarget createWebTarget(final String url) {
         return ClientBuilder
             .newClient()
-            .property("http.receive.timeout", 8000)
+            .property("http.receive.timeout", 8000000)
```
reta, line 61: Please revert

systests/rs-sse/rs-sse-base/src/main/java/org/apache/cxf/systest/jaxrs/sse/AbstractSseTest.java
```diff
@@ -103,6 +115,65 @@ public void testBooksStreamIsReturnedFromInboundSseEvents() throws InterruptedEx
         );
     }
 
+    @SuppressWarnings("unchecked")
+    @Test
+    public void testBooksStreamIsReturnedFromInboundSseEventsWithPOST() throws InterruptedException, IOException {
```
reta, line 120: Thanks for the test, it will help to reproduce the problem.

rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/interceptor/SseInterceptor.java
timothyjward: > It would be good to know what exactly is the issue with adding headers to response?

At the point the headers are copied the content of `PROTOCOL_HEADERS` is the set of HTTP headers from the incoming request. There are numerous issues with copying these to the response, with several capable of causing potentially serious bugs. A non-exhaustive exclusion list would contain:

* `Content-Length`- This contains the length in bytes of the incoming request entity. If set in the response then it is saying that the length of the response is that number of bytes, which is almost certainly wrong. This can cause responses to be cut short, or clients to hang waiting for data that never comes.
* `Content-Type` - This contains the content type of the incoming request entity. If set in the response then it may override the correct content type of `text/event-stream` with something like `text/plain` or `application/json`. This will cause clients to behave incorrectly.
* `Content-Encoding`- This contains the encoding of the incoming request entity. If set in the response then it may cause clients to incorrectly assume the response is compressed with `gzip`, or some other encoding. It could also mask cases where the response *is* compressed but the request wasn't.
* `Transfer-Encoding` - This will contain any encoding performed on the request by the client, or by intermediate proxies. If used then it may override the response which should be using a value of `chunked` for an HTTP 1.1 event stream
* `Upgrade` - The client may request an HTTP upgrade (e.g. to HTTP 2.0, or to WebSocket), with a fallback to HTTP 1.1 SSE if the upgrade isn't possible on the server side. Copying the Upgrade request header back to the client makes no sense according to the upgrade protocol and may cause strange errors.

There are then plenty of headers which only make sense on a request, and should not be used in a response. The [`Host`](https://datatracker.ietf.org/doc/html/rfc2616#section-14.23) and [`User-Agent`](https://datatracker.ietf.org/doc/html/rfc2616#section-14.43) headers that you reference are examples of this sort of header, others would include [`Cookie`](https://datatracker.ietf.org/doc/html/rfc6265#section-5.4), [`Authorization](https://datatracker.ietf.org/doc/html/rfc7235#section-4.2) and [`Accept`](https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.2). Adding the request headers to an HTTP response is confusing as they aren't supposed to be there and don't fit the expected semantics. Depending on the strictness of the HTTP client the presence of any of these headers may also cause issues.

> The headers are included into the response for a reason
I suppose my question would be, what is the reason? It doesn't seem to be following the normal HTTP header rules, but I may be missing something.

systests/rs-sse/rs-sse-base/src/main/java/org/apache/cxf/systest/jaxrs/sse/AbstractSseBaseTest.java
timothyjward: Apologies - I changed this to allow debugging. Clearly a timeout of over two hours isn't advisable for automated integration tests!

rt/rs/sse/src/main/java/org/apache/cxf/jaxrs/sse/interceptor/SseInterceptor.java
reta: Thanks a lot for explanation @timothyjward, I suspect `Content-Length` could be an issue in this case, I will take a look shortly as well to confirm that. Have you seen the examples of `Upgrade` headers to be sent as well in response since you mentioned it? (would appreciate the pointers).  

> There are then plenty of headers which only make sense on a request, and should not be used in a response. The Host and User-Agent headers that you reference are examples of this sort of header, others would include Cookie, Authorization and Accept. Adding the request headers to an HTTP response is confusing as they aren't supposed to be there and don't fit the expected semantics. 

I would argue that there is no harm of sending `User-Agent` or `Host` back in the response, moreover there is absolutely nothing preventing users from sending any headers in the response (obviously, CXF does not send Cookie & co back).  But I think this is subject of another discussion. Back to your question:

> I suppose my question would be, what is the reason? 

Right now PROTOCOL_HEADERS are used in many places (primarily, interceptors) to propagate headers down to the response level, they are flushed by last interceptor in the chain (like `SseInterceptor` in this case). You could look in the codebase for example, I would be happy to provide direct pointers if you are curious.

reta: @timothyjward the problem turned out to be a bit more complicated, it was not about PROTOCOL_HEADERS per se but where they were coming from, it should be fixed in https://github.com/apache/cxf/pull/821

timothyjward: That absolutely looks like the root cause. Thanks for the investigation. 

I’m happy to redo this PR to just include the POST system test if you would like to have it included in the test bucket.

reta: That would be great, thank you !

timothyjward:
I have updated this PR by removing the commit which changed the SseInterceptor, and squashing the timeout fix into the test commit. I have also rebased the PR on top of the latest master. This should now give a single commit containing the test for SSE/POST which will apply cleanly onto `master`. I can confirm that the test passes as a result of the fix in #821 .

reta (approved):
Thanks @timothyjward !

PR #813, opened 2021-06-08
CXF-8545 - support POST requests from SseEventSource

timothyjward:
Add support for POST requests from the SseEventSource

Builds on top of PR #812 (for testing)


reta:
Thanks for the PR, @timothyjward , I believe we should take a step back and think about what we would like to implement first. So we want to provide the option to use POST for SSE events but the `SseEventSource` is GET-oriented and does not have the means to provide the request payload: this is by design and I believe we should not alter that anyhow. This is what the consumer of this API would expect. Probably, the conclusion is we have to define the APIs which would explicitly model SSE-over-POST, it should not be squeezed into `SseEventSource`. May be something along these lines:

```
public interface SsePostEventSource<?> {
    void open(Entity<?> payload);
    void close();
    boolean isOpen();
    void register(Consumer<InboundSseEvent> onEvent);
    ...
}
```

Now, the important question is: what should happen on reconnect? Should the same request payload be sent to the endpoint over and over again? Or should it be up to consumer to decide? I inclined to think that we should use second option since it is very likely the payload could be outdated already. So the API should support that, again just rough idea:

```
    ...
    void register(Provider<Entity<?>> onReconnect);
    ...
```

But it may not be sufficient though, the LAST_EVENT_ID_HEADER could come very handy for example. Anyway, there are few things to think about, I hope it makes sense, thank you.

gnodet:
Closing — the design feedback from @reta was never addressed. The approach of squeezing POST support into the GET-oriented `SseEventSource` API was rejected in favor of a dedicated `SsePostEventSource` interface with proper reconnect semantics.

JIRA: [CXF-8545](https://issues.apache.org/jira/browse/CXF-8545)

If SSE-over-POST is still desired, a fresh PR with the API design reta outlined would be welcome.

PR #816, opened 2021-06-16
fix ModelEncryptionSupport serializer/deserializer

hlavki:
DefaultEncryptingOAuthDataProvider is practically unusable because serialization errors in ModelEncryptionSupport:

1. `new Client("clientId", null, false)` was serialized as with space as secret. So when the client was deserialized, then secretId was `' '` instead of null.
2. All collections were serialized with space e.g. `"openid, email"` then deserialized values with spaces: `'openid', ' email'`

This patch provides fixes for this behavior.


gnodet:
Superseded by #2953 with a clean implementation from main. The core fixes are the same (tokenizeString returns empty instead of space, collections use proper serialization instead of toString()), with additional improvements:
- Empty check in parseSimpleMap to avoid NPE
- Applied getStringPart to client id/secret in recreateClient
- Kept CXF's original code formatting style

Thank you @hlavki for identifying these bugs and the original implementation.

gnodet:
Closing: superseded by PR #2953 which provides a clean fix for the ModelEncryptionSupport serializer/deserializer issue.

PR #818, opened 2021-06-22
CXF-8556: Null AbstractEndpointFactory Properties Interferes with Class-Level PathParam When Creating Proxies with Defined ClientState

reta:
CXF-8556: Null AbstractEndpointFactory Properties Interferes with Class-Level PathParam When Creating Proxies with Defined ClientState


nrickles3:
The changes in this pull request have correctly fixed the issue described here https://issues.apache.org/jira/browse/CXF-8556.

reta:
@nrickles3 Thank you very much for confirming

PR #819, opened 2021-06-23
CXF-8555 - JsonMapObjectReaderWriter escapes now any back slash

coheigea:
The logic is changed so that it now only escapes " and \ if they are not already escaped. I did not choose to escape /, because this broke a ton of tests in different modules.

In addition, I followed the existing logic for /, so that after parsing it strips out the backslashes that were previously escaped.


reta (approved):
LGTM, thanks @coheigea !

PR #822, opened 2021-07-01
CXF-8557: Incorrect Proxy Path Segmenting when @Path Annotation RegexExpression Contains '/'

reta:
A service in question needs to be reachable via multiple paths. The @Path parameter allows for regex matching on a class-level like the following: @Path("/{a: regex}"). A class-level `@PathParam` is created by the name of a. On the server side, the following annotations both correctly receive requests at `/foo/bar`:

```
@Path("/{a : foo/bar}")
@Path("/{a : foo\\/bar}")
```

However, on the client side, the following URI(s)
```
final URI uri = UriBuilder
            .fromUri("/")
            .path("{a : foo/bar}")
            .buildFromEncoded("foo/bar");
```
are not properly constructed due to the fact that `/` is used as path segment separator without taking into account the URI template expression.

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/JAXRSUtils.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/JAXRSUtils.java
@@ -184,8 +184,9 @@ public final class JAXRSUtils {
     /**
      * Parses path segments taking into account the URI templates and template regexes. Per RFC-3986, 
      * "A path consists of a sequence of path segments separated by a slash ("/") character.", however
-     * it is possible to include slash ("/") inside template regex, for example {a:b/c}. In this case,
-     * the whole template definition is extracted as a path segment, without breaking it.
+     * it is possible to include slash ("/") inside template regex, for example "/my/path/{a:b/c}", see 
+     * please {@link URITemplate}. In this case, the whole template definition is extracted as a path 
+     * segment, without breaking it.
      * @param thePath path
      * @param decode should the path segments be decoded or not
      * @param ignoreLastSlash should the last slash be ignored or not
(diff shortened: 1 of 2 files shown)


reta:
@andymc12 if you have a moment please to take a look, much appreciate it

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/JAXRSUtils.java
```diff
@@ -181,20 +181,51 @@ public static boolean isStreamingOutType(Class<?> type) {
...
+                    final String segment = thePath.substring(start, i);
+                    segments.add(new PathSegmentImpl(segment, decode));
+                }
+                
+                // advance the positions, empty path segments
+                start = i + 1;
+            } else if (thePath.charAt(i) == '{') {
+                ++depth;
+            } else if (thePath.charAt(i) == '}') {
+                --depth; // could go negative, since the template could be unbalanced
```
andymc12, line 215: if it goes negative, should we throw an IllegalArgumentException?

andymc12 (approved):
@reta - I think this looks good.  The one thing that I'm sure about is what the expected behavior is when the unclosed curly braces are used. I lean toward throwing an `IllegalArgumentException` unless they are escaped, but I'm not really sure that that is the correct behavior either, so handling them as part of the path seems reasonable too.

Thanks for finding and fixing this!

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/JAXRSUtils.java
reta: Thanks a lot for looking, @andymc12 . I was also not sure what the correct behavior should be. I think throwing the exception in this case would be surprising to the users (we may break existing code), however I think we have to address this issue in any case. I would suggest to fallback to straightforward `/` separator (current implementation) if we detect in unparsable template definitions (like unbalanced curly braces). What do you think?

PR #825, opened 2021-07-08
CXF-8561: SseEventSource sometimes sends partial events when being closed

reta:
According to the specification [1]:

> Once the end of the file is reached, any pending data must be discarded. (If the file ends in the middle of an event, before the final empty line, the incomplete event is not dispatched.)

[1] https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream


andymc12 (approved):
LGTM - thanks!

reta:
Thanks a lot, @andymc12 !

PR #826, opened 2021-07-13
Delete temporary queue when it is used

valentin-matignon:
Temporary queues are never deleted when the are used.
This patch provides fixes for this behavior.

### DIFF
--- a/rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSConduit.java
+++ b/rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSConduit.java
@@ -510,8 +510,8 @@ public class JMSConduit extends AbstractConduit implements JMSExchangeSender, Me
         }
     }
     public synchronized void close() {
-        jmsConfig.resetCachedReplyDestination();
         shutdownListeners();
+        jmsConfig.resetCachedReplyDestination();
         ResourceCloser.close(connection);
         connection = null;
         LOG.log(Level.FINE, "JMSConduit closed ");
--- a/rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSConduit.java
+++ b/rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSConduit.java
@@ -36,7 +36,6 @@ import javax.jms.ExceptionListener;
 import javax.jms.JMSException;
 import javax.jms.MessageListener;
 import javax.jms.Session;
-import javax.jms.TemporaryQueue;
 
 import org.apache.cxf.Bus;
 import org.apache.cxf.buslifecycle.BusLifeCycleListener;
@@ -128,10 +127,14 @@ public class JMSConduit extends AbstractConduit implements JMSExchangeSender, Me
     private void trySetExListener(Connection conn) {
         try {
             conn.setExceptionListener(new ExceptionListener() {
-                
+
                 @Override
                 public void onException(JMSException exception) {
-                    jmsConfig.resetCachedReplyDestination();
+                    try {
+                        jmsConfig.resetCachedReplyDestination();
+                    } catch (JMSException e) {
+                        // setException is not supported on all providers
+                    }
                     staticReplyDestination = null;
                 }
             });
@@ -189,7 +192,11 @@ public class JMSConduit extends AbstractConduit implements JMSExchangeSender, Me
                 }
                 ResourceCloser.close(connection);
                 this.connection = null;
-                jmsConfig.resetCachedReplyDestination();
+                try {
+                    jmsConfig.resetCachedReplyDestination();
+                } catch (JMSException jmsException) {
+                    // Ignore
+                }
             }
             this.staticReplyDestination = null;
             try {
@@ -283,9 +290,6 @@ public class JMSConduit extends AbstractConduit implements JMSExchangeSender, Me
                                                                      jmsConfig.getReceiveTimeout(),
                                                                      jmsConfig.isPubSubNoLocal(),
                                                                      exchange);
-                    if (replyDestination instanceof TemporaryQueue) {
-                        ((TemporaryQueue) replyDestination).delete();
-                    }
                     processReplyMessage(exchange, replyMessage);
                 } else {
                     try {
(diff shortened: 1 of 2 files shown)


rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSConduit.java
```diff
@@ -282,6 +283,9 @@ private void sendAndReceiveMessage(final Exchange exchange, final Object request
                                                                      jmsConfig.getReceiveTimeout(),
                                                                      jmsConfig.isPubSubNoLocal(),
                                                                      exchange);
+                    if (replyDestination instanceof TemporaryQueue) {
```
reta, line 286: Thanks for the patch @valentin-matignon , I believe the this is not the right place to delete the temporary queue. 
 - `JMSConfiguration` is the class that create temporary queue (and caches it)
 - `JMSConfiguration` has the method `resetCachedReplyDestination` which looks like the right place to delete the temporary queue (it caches the destination in `replyToDestinationDest`)
 - the `JMSConduit` should be likely calling  `JMSConfiguration::resetCachedReplyDestination()` on shutdown
 - it would be great to have at least one test case for the issue

What do you think? Thank you.

valentin-matignon: Hi @reta, thank you for your answer and sorry to respond so late !

I agree with you, `JMSConfiguration` should delete temporary queues.
As you advised, in my last commit i delete temporary queues in the `resetCachedReplyDestination` method of the `JMSConfiguration` class.

However, `resetCachedReplyDestination` is called in two places places in the `JMSConduit` class :

- in a `ExceptionListener` in the `trySetExListener` method
- in the `sendExchange` method when a `JMSException` occurs

So i had to add the `resetCachedReplyDestination` in the `close` method of `JMSConduit` class to be sure that temporary queues are deleted every time they are used.

I didn't find a proper way to add one test case in unit tests, if you have some suggestions i could add it in my PR.

rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSConduit.java
```diff
@@ -510,6 +510,7 @@ private synchronized void shutdownListeners() {
         }
     }
     public synchronized void close() {
+        jmsConfig.resetCachedReplyDestination();
```
reta, line 513: I believe we should call it after the `shutdownListeners()` (because we still have listeners on the queue being deleted) but before `ResourceCloser.close(connection)`.

reta:
@valentin-matignon I was testing the changes and it does not seem like this PR is really needed. The JMS spec [1] says
> TemporaryQueue createTemporaryQueue() throws JMSException
>
> Creates a TemporaryQueue object. Its lifetime will be that of the Connection unless it is deleted earlier.

And indeed their lifetime is bound to the connection (tests confirm that): connection closes, queue is gone. So my question to you is: which message broker are you using that is keeping the temporary queues upon connection closing?

[1] https://docs.oracle.com/javaee/7/api/javax/jms/Session.html#createTemporaryQueue--

rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSConduit.java
valentin-matignon: I moved it.

valentin-matignon:
@reta, i'm using a TIBCO broker and the **_com.tibco:tibjms:8.3.0_**, **_javax.jms:javax.jms-api:2.0.1_** dependencies in my program.

Which tests are you talking about ?

- I tested `testRequestQueueResponseTempQueue` and `testRequestTopicResponseTempQueue` tests of `RequestResponseTest` class and the `delete` method of the associated temporary queue (`ActiveMQTempQueue` in this case) **is never called** because it is not called by the `close` method of the associated connection (`ActiveMQConnection` in this case)
- I tested `testTempQueueIssue` test of `PooledConnectionTempQueueTest` class and the `delete` method of the associated temporary queue (`ActiveMQTempQueue` in this case) **is called** because it is called by the `close` method of the associated connection (`PooledConnection` in this case)

Have you noticed the same or am I wrong?

reta:
@valentin-matignon you are right, at least in case of ActiveMQ, the pooled connection do cleanup temporary queue, but non-pooled - not. My apologies for confusion, I was checking ActiveMQ's `DestinationSource` and it was properly reporting the removal of the temporary queues (but it seems like - not physical removal).

reta:
@valentin-matignon do you mind if I push one test case your pull request? thank you.

valentin-matignon:
@reta no problem, you can do it.

rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSConfiguration.java
```diff
@@ -19,22 +19,29 @@
...
 import org.apache.cxf.common.injection.NoJSR250Annotations;
+import org.apache.cxf.common.logging.LogUtils;
 import org.apache.cxf.transport.jms.util.DestinationResolver;
 import org.apache.cxf.transport.jms.util.JMSDestinationResolver;
 import org.apache.cxf.transport.jms.util.JndiHelper;
 
 @NoJSR250Annotations
 public class JMSConfiguration {
+
+    private static final Logger LOG = LogUtils.getL7dLogger(JMSConfiguration.class);
```
reta, line 43: Could you please move `LOG` after `DEFAULT_VALUE`,  (the `Checkstyle` will complain)

rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSConduit.java
```diff
@@ -511,6 +511,7 @@ private synchronized void shutdownListeners() {
     }
     public synchronized void close() {
         shutdownListeners();
+        jmsConfig.resetCachedReplyDestination();
```
reta, line 514: One similar modification is need in `sendExchange` method, right now the connection is closed before the `jmsConfig.resetCachedReplyDestination();`, so there is no chance to delete temporary query. The correct sequence should be:

```
                jmsConfig.resetCachedReplyDestination();
                ResourceCloser.close(connection);
                this.connection = null;
```

reta:
Issue: https://issues.apache.org/jira/browse/CXF-8591

rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/JMSConduit.java
valentin-matignon: You're right, done.

reta (approved):
Thank you @valentin-matignon !

valentin-matignon:
> Thank you @valentin-matignon !

Thanks to you for the code review @reta 😄 !

PR #831, opened 2021-08-06
CXF-8574: MP Rest call on jax-rs interface with long @Path and 3 path params causes error since

reta:
The usage of varargs made unwanted parameters treated as variables


reta:
@andymc12 if you have a minute, mind please taking a look? thank you!

andymc12 (approved):
@reta this looks good to me. One alternative might be to change `MicroProfileClientFactoryBean` to not send the `TLSConfiguration` to `MicroProfileClientProxyImpl` - afaict, it is not used by the proxyimpl - it looks like the factory bean uses it in the `initClient` method instead. Wdyt? I'm fine either way.

reta:
Thanks @andymc12 , I had exactly the same dilemma 😄 , decided to keep it since we adeady passing it, could be useful for debug / introspection purposes, thanks for the review!

PR #833, opened 2021-08-11
CXF-8579: Switch JDK compiler to use JDK 11

mmoayyed:
Avoid errors when building on JDK 11:

```
"[ERROR] /home/jitpack/build/rt/security-saml/src/main/java/org/apache/cxf/rt/security/saml/utils/SAMLUtils.java:[37,36] cannot access org.apache.wss4j.common.saml.SamlAssertionWrapper
  bad class file: /home/jitpack/.m2/repository/org/apache/wss4j/wss4j-ws-security-common/2.5.0-SNAPSHOT/wss4j-ws-security-common-2.5.0-SNAPSHOT.jar(org/apache/wss4j/common/saml/SamlAssertionWrapper.class)
    class file has wrong version 55.0, should be 52.0
    Please remove or make sure it appears in the correct subdirectory of the classpath."
```

This PR is related to: https://issues.apache.org/jira/browse/CXF-8579


reta:
@coheigea we could close this one, right? we are on JDK-11 already

coheigea:
Yes, let's close it.

PR #836, opened 2021-08-25
[CXF-8578] Bind bridge methods to operations

octachoron:
For each real method bound to an operation, add the corresponding bridge
methods of the class as secondary bindings. This gives them a
method-to-operation mapping each, enabling them to be used in client
proxies. On the other hand, only the real method gets an
operation-to-method mapping, which avoids re-introducing CXF-7670.

### DIFF
--- a/distribution/src/main/appended-resources/META-INF/NOTICE
+++ b/distribution/src/main/appended-resources/META-INF/NOTICE
@@ -58,3 +58,5 @@ This product includes code from the Nimbus JOSE + JWT project, under the Apache
 license 2.0 (https://github.com/felx/nimbus-jose-jwt). Copyright 2012-2017,
 Connect2id Ltd.
 
+Portions of this software were developed at
+Cloudera, Inc (https://www.cloudera.com/).
--- a/core/src/main/java/org/apache/cxf/common/util/ReflectionUtil.java
+++ b/core/src/main/java/org/apache/cxf/common/util/ReflectionUtil.java
@@ -32,7 +32,11 @@ import java.security.PrivilegedAction;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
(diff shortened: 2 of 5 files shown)


distribution/src/main/appended-resources/META-INF/NOTICE
```diff
@@ -58,3 +58,5 @@ This product includes code from the Nimbus JOSE + JWT project, under the Apache
 license 2.0 (https://github.com/felx/nimbus-jose-jwt). Copyright 2012-2017,
 Connect2id Ltd.
 
+Portions of this software were developed at
```
ffang, line 61: First of all, thanks so much for the contribution!

But I'm totally not sure if we should/can add this into NOTICE. Let's wait other comments from the community.

distribution/src/main/appended-resources/META-INF/NOTICE
```diff
@@ -58,3 +58,5 @@ This product includes code from the Nimbus JOSE + JWT project, under the Apache
...
+Cloudera, Inc (https://www.cloudera.com/).
```
ffang, line 62: First of all, thanks so much for the contribution!

But I'm totally not sure if we should/can add this into NOTICE. Let's wait other comments from the community.

octachoron: I'm unsure (and curious) about this part as well. Thank you very much for pointing this out!

reta:
@octachoron thanks a lot for the issue and the pull request. First of all, I would like to agree with you that handling of the bridge / synthetic methods could be improved (especially, when using client proxies) but I would like to suggest a bit different fix here: https://github.com/apache/cxf/pull/842. The main idea is to complement the operation with additional binding details while evaluating resource class. Certain optimizations (like skipping the methods returning `Object`, ...) would also probably be beneficial. 

Your feedback would be appreciated, the edge cases you have demonstrated in the ticket are passing now, thank you.

octachoron:
@reta #842 does sound like a better and more straightforward way to do this, both in terms of obtaining and binding the bridge methods, and testing through an actual client proxy. I also tested the approach against the code where we originally saw the symptoms, and it runs without issues.

I will go ahead and close this PR in favor of yours. Thank you very much for the review and suggestion.

PR #846, opened 2021-09-08
CXF-8593: Error when installing CXF OpenAPI v3 under Karaf 4.2.11

reta:
Thanks to @ffang for identifying the root cause: the root cause of this error is that in the container two `javax.servlet` versions are required.


ffang (approved):

ffang:
Thanks @reta for the quick fix!

PR #847, opened 2021-09-08
[CXF-8594]:Fix regular expression parameter with non-capturing group …

jimma:
…doesn't work

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/model/URITemplate.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/model/URITemplate.java
@@ -83,7 +83,7 @@ public final class URITemplate {
                     // however do not add them if they already exist since that will cause the Matcher
                     // to create extraneous values.  Parens identify a group so multiple parens would
                     // indicate multiple groups.
-                    if (pattern.startsWith("(") && pattern.endsWith(")")) {
+                    if (pattern.startsWith("(") && pattern.endsWith(")") && !pattern.startsWith("(?:")) {
                         patternBuilder.append(pattern);
                     } else {
                         patternBuilder.append('(');
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/model/URITemplate.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/model/URITemplate.java
@@ -83,7 +83,7 @@ public final class URITemplate {
                     // however do not add them if they already exist since that will cause the Matcher
                     // to create extraneous values.  Parens identify a group so multiple parens would
                     // indicate multiple groups.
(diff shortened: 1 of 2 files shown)


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/model/URITemplate.java
```diff
@@ -83,7 +83,7 @@ public URITemplate(String theTemplate) {
                     // however do not add them if they already exist since that will cause the Matcher
                     // to create extraneous values.  Parens identify a group so multiple parens would
                     // indicate multiple groups.
-                    if (pattern.startsWith("(") && pattern.endsWith(")")) {
+                    if (pattern.startsWith("(") && pattern.endsWith(")") && !pattern.startsWith("(?:")) {
```
reta, line 86: Thanks for the fix @jimma , I am wondering if we should cover other special constructs [1] with more broader`!pattern.startsWith("(?")` check, for example `(?>X)`, an independent, non-capturing group, etc

```
(?<name>X) 	                X, as a named-capturing group
(?:X) 	                        X, as a non-capturing group
(?idmsuxU-idmsuxU)      Nothing, but turns match flags i d m s u x U on - off
(?idmsux-idmsux:X)        X, as a non-capturing group with the given flags i d m s u x on - off
(?=X) 	                        X, via zero-width positive lookahead
(?!X) 	                        X, via zero-width negative lookahead
(?<=X) 	                        X, via zero-width positive lookbehind
(?<!X) 	                        X, via zero-width negative lookbehind
(?>X) 	                        X, as an independent, non-capturing group
```

[1] https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html

jimma: @reta  You are right ! I don't know there is another non-capture group like (?>X) and so many lookahead/lookbehind expressions.  Regex is powerful!  I already improved the fix as you suggested. Please review it again. Thanks !

reta (approved):

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/model/URITemplate.java
reta: 👍 thanks @jimma !

reta:
@jimma this one could be integrated, right @jimma ?

jimma:
@reta  what do you mean by integrated ?

reta:
@jimma oh sorry, I meant - merged

jimma:
@reta  Sure. All tests are passed, and it's ready to merge.

reta:
Thanks @jimma , I think you could merge it, right? Or you prefer me to merge and backport it? Thanks!

jimma:
@reta  If you get time today to merge and backport , please give it go. Otherwise I'll back to look tomorrow morning.

PR #849, opened 2021-09-12
CXF-7543: JAX-RS Features not used in proxies or WebClients

reta:
The `JAXRSClientFactoryBean` does not support JAX-RS features, as such if the feature registers additional providers, then those providers are never considered for use on the proxy or `WebClient` instance created.

### DIFF
--- a/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/JAXRSClientFactoryBean.java
+++ b/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/JAXRSClientFactoryBean.java
@@ -24,12 +24,19 @@ import java.lang.reflect.Type;
 import java.net.URI;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.logging.Logger;
 
 import javax.ws.rs.CookieParam;
 import javax.ws.rs.HeaderParam;
+import javax.ws.rs.RuntimeType;
+import javax.ws.rs.core.Configurable;
+import javax.ws.rs.core.Configuration;
+import javax.ws.rs.core.Feature;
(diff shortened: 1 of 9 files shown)


rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/JAXRSClientFactoryBean.java
```diff
@@ -403,13 +412,45 @@ protected void initClient(AbstractClient client, Endpoint ep, boolean addHeaders
             });
         }
     }
+    
+    protected <C extends Configurable<C>> Configurable<?> getConfigurableFor(C context) {
```
andymc12, line 416: Just thinking that this method might be better off named `newConfigurableFor` to avoid confusion.  My first thought was that this method might be used to cache the configurable based on the passed-in context - like a `Map<C, Configurable<C>>`.  No worries if you'd prefer to keep it as-is though.

andymc12 (approved):
LGTM - only one comment, and you can ignore it if you like.  Thanks for putting this together!

rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/JAXRSClientFactoryBean.java
reta: Thanks a lot for looking, @andymc12 ! I was also initially though of "create" naming but in case of MP, it turned out to be not exactly the case: the existing instance of the configurable is used for any context. In this regards, `getConfigurableFor` is a bit more honest name with respect to semantics, but `getOrCreateConfigurableFor` would also make sense. What do you think?

andymc12: Oh - I totally missed that.  I think either way (`getConfigurableFor` or `getOrCreateConfigurableFor`) is fine.  Thanks for the clarification.

andymc12 (approved):

PR #850, opened 2021-09-13
[CXF-8596]:Fix infinite loop in FaultOutInterceptor

jimma:


core/src/main/java/org/apache/cxf/interceptor/FaultOutInterceptor.java
```diff
@@ -81,6 +82,15 @@ public void handleMessage(Message message) {
 
             MessagePartInfo part = fi.getFirstMessagePart();
             DataBinding db = service.getDataBinding();
+            boolean faultDetail = cause instanceof SOAPFaultException
+                    && ((SOAPFaultException)cause).getFault().hasDetail();
+            if (faultDetail) {
+                //the fault detail is already set and we don't need to write
+                //Writes the SoapFaultException to detail will go into an infinite loop because
+                //SoapFaultException's cause field is itself
```
reta, line 90: I am wondering if we should specifically check for the loop in `SoapFaultException`, following the `getCause()` chain? It seems like otherwise we are cutting off `if (isDOMSupported(db)) {` branch altogether in some cases (not sure how important it is to be fair).

jimma: This FaultOutInterceptor invocation actually goes from WebFaultOutInterceptor: https://github.com/apache/cxf/blob/master/rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/interceptors/WebFaultOutInterceptor.java#L192. There is check for SoapFaultException in WebFaultOutInterceptor

coheigea:
LGTM but @dkulp can you take a look?

jimma:
@dkulp  Can you help please review this ?  @ffang Any comments ?

dkulp:
My gut feeling is that this is "wrong".   This adds some additional direct "jaxws" dependencies on cxf-core which is something we've tried to reduce.   My thought would be that the WebFaultOutInterceptor should have an additional "if" near the bottom before calling the super.handleMessage(message).   Something like:

```
} else if (f instanceof SoapFault && f.getCause() instanceof SOAPFaultException 
            && ((SOAPFaultException)f.getCause()).getFault().hasDetail()) {
            return;
}
```
That makes the new test pass, but I'm not sure if it fully solves the original problem so it would be great if you could test that and resubmit the PR with that change if it does.

jimma:
> My gut feeling is that this is "wrong". This adds some additional direct "jaxws" dependencies on cxf-core which is something we've tried to reduce. My thought would be that the WebFaultOutInterceptor should have an additional "if" near the bottom before calling the super.handleMessage(message). Something like:

>  } else if (f instanceof SoapFault && f.getCause() instanceof SOAPFaultException 
            && ((SOAPFaultException)f.getCause()).getFault().hasDetail()) {
            return;
> }

@dkulp  Thanks for the comments. I thought about this solution, but I am not sure if there will be some failure goes into FaultOutInterceptorChain when wsa is enabled.  When wsa is enabled, will it must go into WebFaultOutInterceptor first ?

jimma:
Checked this with WSAFaultToClientServerTest and it shows the FaultTo a different destination goes into WebFaultOutInteceptor first. So adding this check in WebFaultOut Interceptor looks better. I changed as @dkulp suggested.

PR #851, opened 2021-09-17
CXF-8597: CXF JAXRS client not closing HTTP connections

reta:
Relates to https://github.com/apache/cxf/pull/697, the entity input stream is indeed not closed after consumption. But according to the specification for `public abstract <T> T readEntity(Class<T> entityType, Annotation[] annotations)`:

> A message instance returned from this method will be cached for subsequent retrievals via getEntity(). Unless the supplied entitytype is an input stream, this method automatically closes the an unconsumed original response entity data stream if open.

- [X] Add Test Case(s) for WebClient
- [X] Add Test Case(s) for JAXRSClientFactoryBean

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
@@ -55,7 +55,6 @@ import javax.ws.rs.core.Response.Status.Family;
 import javax.ws.rs.ext.ReaderInterceptor;
 import javax.ws.rs.ext.RuntimeDelegate.HeaderDelegate;
 import javax.xml.stream.XMLStreamReader;
-import javax.xml.transform.Source;
 
 import org.apache.cxf.helpers.IOUtils;
 import org.apache.cxf.io.ReaderInputStream;
@@ -405,7 +404,7 @@ public final class ResponseImpl extends Response {
 
     @Override
     public <T> T readEntity(Class<T> cls, Annotation[] anns) throws ProcessingException, IllegalStateException {
-        return doReadEntity(cls, cls, anns, true);
+        return doReadEntity(cls, cls, anns);
     }
 
     @Override
@@ -413,20 +412,13 @@ public final class ResponseImpl extends Response {
     public <T> T readEntity(GenericType<T> genType, Annotation[] anns)
         throws ProcessingException, IllegalStateException {
         return doReadEntity((Class<T>) genType.getRawType(),
-                            genType.getType(), anns, true);
+                            genType.getType(), anns);
     }
 
     public <T> T doReadEntity(Class<T> cls, Type t, Annotation[] anns)
         throws ProcessingException, IllegalStateException {
-        return doReadEntity(cls, t, anns, false);
-    }
-
-    public <T> T doReadEntity(Class<T> cls, Type t, Annotation[] anns, boolean closeAfterRead)
-        throws ProcessingException, IllegalStateException {
 
         checkEntityIsClosed();
-        //according to javadoc, should close when is not buffered.
-        boolean shouldClose = !this.entityBufferred;
 
         if (lastEntity != null && cls.isAssignableFrom(lastEntity.getClass())
             && !(lastEntity instanceof InputStream)) {
@@ -476,11 +468,7 @@ public final class ResponseImpl extends Response {
                                                                   responseMessage);
                 // close the entity after readEntity is called.
                 T tCastLastEntity = castLastEntity();
-                shouldClose = shouldClose && !(tCastLastEntity instanceof AutoCloseable)
-                    && !(tCastLastEntity instanceof Source);
-                if (closeAfterRead && shouldClose) {
-                    close();
-                }
+                autoClose(cls, false);
                 return tCastLastEntity;
             } catch (NoContentException ex) {
                 //when content is empty, return null instead of throw exception to pass TCK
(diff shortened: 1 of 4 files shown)


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
```diff
@@ -418,7 +418,7 @@ private Link makeAbsoluteLink(Link link) {
 
     public <T> T doReadEntity(Class<T> cls, Type t, Annotation[] anns)
         throws ProcessingException, IllegalStateException {
-        return doReadEntity(cls, t, anns, false);
+        return doReadEntity(cls, t, anns, true);
```
reta, line 421: @andymc12 may you please take a look? I suspect we may break some TCK tests but it seems like the method must close the entity (if input stream) after consumption. Thank you!

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
```diff
@@ -418,7 +418,7 @@ private Link makeAbsoluteLink(Link link) {
 
     public <T> T doReadEntity(Class<T> cls, Type t, Annotation[] anns)
         throws ProcessingException, IllegalStateException {
-        return doReadEntity(cls, t, anns, false);
+        return doReadEntity(cls, t, anns, !InputStream.class.isAssignableFrom(cls));
```
reta, line 421: @andymc12 may you please take a look? I suspect we may break some TCK tests but it seems like the method must close the entity (if input stream) after consumption. Thank you!

andymc12: Hmm... My take on this is that the user should be closing their `Response` objects (which are `AutoCloseable` making it easier to do so).  Despite my comment to the contrary, the `readEntity` methods are not supposed to close the `Response` - they are supposed to close the underlying entity stream.  Maybe we could fix this by always closing the entity stream in the finally block?

FWIW, RESTEasy uses a `finalize()` method on their `Response` implementation class that will close the response once garbage collected as a safety net.  That might also work here. Wdyt?

lkoe: the bug author here :-)
@andymc12 the use case in question is a simplistic JAXRS Proxy Client where user code really has no access to the response object. 
Something like this:
```
    public interface TestResource {

        @POST
        @Path("/test")
        @Consumes(MediaType.APPLICATION_JSON)
        @Produces(MediaType.APPLICATION_JSON)
        TestModel test(TestModel body);
    }
```

reta: @andymc12 thanks a lot for your thoughts, indeed closing the response is the right thing to do but in case of client proxies, it is not explicit. Using finalizers could be an option but they are deprecated since JDK-9 [1] and are discouraged. I will look more closely into, it seems like we could isolate the client proxies, thank you 

[1] https://bugs.openjdk.java.net/browse/JDK-8165641

andymc12: In the case of proxy clients, I think CXF should be closing the `Response`.  I think we may already do that in some cases where the `Response` object is not visible to the user.  Maybe we should modify the `WebClient.castResponse` method to also conditionally close the response - something like this:
```
@SuppressWarnings("unchecked")
private <T> T castResponse(Response r, Class<T> responseClass) {
    if (responseClass == Response.class)
        return r;
    try (Response r2 = r) {
        return (T) r.getEntity();
    }
}
```

This might need some testing, but I cannot think of a case where we would need to keep the `Response` open if the user cannot access it. An alternative would be to close the `Response` in the various `invoke` methods, but that might be more code than necessary.  Then maybe something similar in `ClientProxyImpl.handleResponse`.  Wdyt?

reta: Thanks @andymc12 , yes, I was thinking about `ClientProxyImpl.handleResponse` first to basically use try-with-resources on response, I will explore this path shortly.

reta: @andymc12 I think I nailed it down. In refactoring, we have lost the semantics of the `response.stream.auto.close` property (it was only taking into account in some specific flow). Also, I think the default auto-close detection mechanism is good enough to not requiring manual should/should not close instructions. The builds are green, what do you think about this one? Thank you.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
```diff
@@ -55,7 +55,6 @@
 import javax.ws.rs.ext.ReaderInterceptor;
 import javax.ws.rs.ext.RuntimeDelegate.HeaderDelegate;
 import javax.xml.stream.XMLStreamReader;
-import javax.xml.transform.Source;
```
reta, line 58: Should `javax.xml.transform.Source` be part of the `JAXRSUtils::isStreamingOutType` ?

andymc12 (approved):
LGTM - thanks @reta!

PR #860, opened 2021-10-08
[CXF-8603]  Add null check to XPathUtils.getValue() to avoid setting null CCL

archiepatrick:
Added null check to XPathUtils.getValue() to prevent setting null CCL - similar pattern as used for other instances of ClassLoaderUtils.setThreadContextClassLoader()
Code tested against version CXF 3.1.18


reta:
@archiepatrick thank you for the pull request, I am closing this, resolved by https://github.com/apache/cxf/commit/b1753fa70a86292e0965cf570fae5d81e280080a.

PR #862, opened 2021-10-12
Fix not unique JAXRS method candidates WARN caused by CrossOriginResourceSharingFilter 

vladimirfx:
Fix 'JAXRSUtils : Both Resource#method1 and Resource#method2 are equal candidates for handling the current request which can lead to unpredictable results' warning by using real 'Accept' and 'Content-Type' values for operation lookup.


reta:
Thanks for the pull request, @vladimirfx , could you please open an issue in the JIRA tracker (https://issues.apache.org/jira/projects/CXF)?

gnodet:
Superseded by #2954 which takes a simpler approach — rather than changing the CORS filter to pass real media types, we downgrade the log message from WARNING to FINE (debug). The message fires in expected CORS preflight scenarios and the behavior is deterministic despite the "unpredictable" wording. Thank you @vladimirfx for reporting the issue.

vladimirfx:
Superseded by https://github.com/apache/cxf/pull/2954

PR #866, opened 2021-10-23
CXF-8601: [Regression] jaxrs.ee.rs.core.response tests

reta:
By default, the response entity is never closed automatically which is not compliant with JAX-RS specification and TCK. The auto-close behavior is controlled by `ResponseImpl#RESPONSE_STREAM_AUTO_CLOSE` property which is set to "false" by default.

One of the possible solutions which tries to minimize the impact of the change as much as possible is to introduce the `autoCloseHint`. It alters the default value of this property to be "true" for all non-streaming and non-streaming like responses  in order to follow the specification. The additional set of streaming-like responses, including `MultipartBody`, `Attachment`, ... and all supported reactive types have been introduced.

Fixing following tests:
- com/sun/ts/tests/jaxrs/ee/rs/core/response/JAXRSClient.java#getEntityThrowsIllegalStateExceptionWhenConsumedTest_from_standalone: JAXRSClient_getEntityThrowsIllegalStateExceptionWhenConsumedTest_from_standalone
com/sun/ts/tests/jaxrs/ee/rs/core/response
Regression causes jaxrs.ee.rs.core.response TCK tests to fail:

- /JAXRSClient.java#readEntityClassAnnotationThrowsIllegalStateExceptionTest_from_standalone: JAXRSClient_readEntityClassAnnotationThrowsIllegalStateExceptionTest_from_standalone
com/sun/ts/tests/jaxrs/ee/rs/core/response

- /JAXRSClient.java#readEntityClassThrowsIllegalStateExceptionTest_from_standalone: JAXRSClient_readEntityClassThrowsIllegalStateExceptionTest_from_standalone
com/sun/ts/tests/jaxrs/ee/rs/core/response

- /JAXRSClient.java#readEntityGenericTypeAnnotationThrowsIllegalStateExceptionTest_from_standalone: JAXRSClient_readEntityGenericTypeAnnotationThrowsIllegalStateExceptionTest_from_standalone
com/sun/ts/tests/jaxrs/ee/rs/core/response

- /JAXRSClient.java#readEntityGenericTypeThrowsIllegalStateExceptionTest_from_standalone: JAXRSClient_readEntityGenericTypeThrowsIllegalStateExceptionTest_from_standalone

### DIFF
--- a/rt/frontend/jaxrs/src/test/java/org/apache/cxf/jaxrs/impl/ResponseImplTest.java
+++ b/rt/frontend/jaxrs/src/test/java/org/apache/cxf/jaxrs/impl/ResponseImplTest.java
@@ -21,6 +21,8 @@ package org.apache.cxf.jaxrs.impl;
 
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
+import java.io.Reader;
+import java.lang.annotation.Annotation;
 import java.net.URI;
 import java.util.List;
 import java.util.Locale;
@@ -71,6 +73,7 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
 
 
@@ -598,6 +601,42 @@ public class ResponseImplTest {
         }
     }
 
+    @Test
+    public void testReadEntityAndGetEntityAfter() {
+        final String str = "ouch";
+
+        final ResponseImpl response = new ResponseImpl(500, str);
+        final Message outMessage = createMessage();
+        outMessage.put(Message.REQUEST_URI, "http://localhost");
+        response.setOutMessage(outMessage);
+
+        final MultivaluedMap<String, Object> headers = new MetadataMap<>();
+        headers.putSingle("Content-Type", "text/xml");
+        response.addMetadata(headers);
+
+        assertEquals(str, response.readEntity(String.class));
+        assertThrows(IllegalStateException.class, () -> response.getEntity());
+    }
+    
+    @Test
+    public void testReadEntityWithAnnotations() {
+        final String str = "ouch";
+
+        final ResponseImpl response = new ResponseImpl(500, str);
+        final Message outMessage = createMessage();
+        outMessage.put(Message.REQUEST_URI, "http://localhost");
+        response.setOutMessage(outMessage);
+
+        final MultivaluedMap<String, Object> headers = new MetadataMap<>();
+        headers.putSingle("Content-Type", "text/xml");
+        response.addMetadata(headers);
+
+        final Annotation[] annotations = String.class.getAnnotations();
+        assertEquals(str, response.readEntity(String.class, annotations));
+        assertThrows(IllegalStateException.class, 
+            () -> response.readEntity(Reader.class, annotations));
+    }
+    
     public static class StringBean {
         private String header;
 
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
@@ -66,7 +66,7 @@ import org.apache.cxf.message.Message;
 import org.apache.cxf.message.MessageUtils;
 
 public final class ResponseImpl extends Response {
-
+    public static final String RESPONSE_STREAM_AUTO_CLOSE = "response.stream.auto.close";
     private static final Pattern LINK_DELIMITER = Pattern.compile(",\\s*(?=\\<|$)");
 
     private StatusType status;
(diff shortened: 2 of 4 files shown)


rt/frontend/jaxrs/src/test/java/org/apache/cxf/jaxrs/impl/ResponseImplTest.java
```diff
@@ -598,6 +601,42 @@ public void testGetLinksMultipleMultiline() {
...
+        final Message outMessage = createMessage();
+        outMessage.put(Message.REQUEST_URI, "http://localhost");
+        response.setOutMessage(outMessage);
+
+        final MultivaluedMap<String, Object> headers = new MetadataMap<>();
+        headers.putSingle("Content-Type", "text/xml");
+        response.addMetadata(headers);
+
+        assertEquals(str, response.readEntity(String.class));
+        assertThrows(IllegalStateException.class, () -> response.getEntity());
```
andymc12, line 618: I think that this is an incorrect behavior.  Per the javadocs at https://javadoc.io/static/javax.ws.rs/javax.ws.rs-api/2.1.1/javax/ws/rs/core/Response.html#getEntity-- , the `response.getEntity()` method should not throw an `IllegalStateException` unless the response has been closed or if the entity type is a stream that has been fully consumed.
> Throws:
IllegalStateException - if the entity was previously fully consumed as an input stream, or if the response has been closed.

And I believe that it is expected that the `response.getEntity()` _should_ return the cached entity after calling `readEntity` - see the javadoc from `readEntity(Class)` ( https://javadoc.io/static/javax.ws.rs/javax.ws.rs-api/2.1.1/javax/ws/rs/core/Response.html#readEntity-java.lang.Class- ) :
> A message instance returned from this method will be cached for subsequent retrievals via getEntity(). Unless the supplied entity type is an input stream, this method automatically closes the an unconsumed original response entity data stream if open. In case the entity data has been buffered, the buffer will be reset prior consuming the buffered data to enable subsequent invocations of readEntity(...) methods on this response.

So my take is that the these lines should be:
 ```suggestion
        assertEquals(str, response.readEntity(String.class));
        assertEquals(str, response.getEntity());
        // ideally we would verify that the input stream that backs the entity was closed, but that the Response was not - however in this case the stream is temporary (created via `convertEntityToStreamIfPossible()`)
```

**Review by andymc12 — COMMENTED** [review]
I'm thinking that it is incorrect to close the `Response` after reading it (see my comments in the test case).   I'm wondering if we could pass the cast-type to a new `ResponseImpl` that should only be used on the client - the caller would need to specify what the response entity is intended to be cast to by proxy-based clients (JAX-RS clients would continue using the existing constructors).  Then if the proxy-based client returns a `Response`, we leave the behavior the same as JAX-RS, but otherwise we would close after reading. Wdyt?

File: rt/frontend/jaxrs/src/test/java/org/apache/cxf/jaxrs/impl/ResponseImplTest.java (hunk 1 above)
Comment (reta) [reply rt/frontend/jaxrs/src/test/java/org/apache/cxf/jaxrs/impl/ResponseImplTest.java:618]: Thanks @andymc12 , so this is exactly the way it works right now, but I am actually trying to fix this TCK test (and few others), which verifies this exact flow (unless I miss something): https://github.com/eclipse-ee4j/jakartaee-tck/blob/8.0.2/src/com/sun/ts/tests/jaxrs/ee/rs/core/response/JAXRSClient.java#L343

```
public void getEntityThrowsIllegalStateExceptionWhenConsumedTest()
      throws Fault {
    Response response = invokeGet("entity");
    response.readEntity(String.class);
    try {
      Object entity = response.getEntity();
      fault("No exception has been thrown entity=", entity);
    } catch (IllegalStateException e) {
      logMsg("#getEntity throws IllegalStateException as expected", e);
    }
  }
```

**Comment by reta** [comment]
> 
> 
> I'm thinking that it is incorrect to close the `Response` after reading it (see my comments in the test case). I'm wondering if we could pass the cast-type to a new `ResponseImpl` that should only be used on the client - the caller would need to specify what the response entity is intended to be cast to by proxy-based clients (JAX-RS clients would continue using the existing constructors). Then if the proxy-based client returns a `Response`, we leave the behavior the same as JAX-RS, but otherwise we would close after reading. Wdyt?

I would certainly prefer to keep it unchanged (and not close the response after reading), having a hard time to fit CXF into the TCK / spec space without changing that.

Also, I have consulted Jersey as the reference implementation and the  the `Response` is indeed closed after reading, so I think we are changing the right parts (see please https://github.com/eclipse-ee4j/jersey/blob/master/core-common/src/main/java/org/glassfish/jersey/message/internal/InboundMessageContext.java#L896). 

Would be great to understand what I am missing here?

**Comment by reta** [comment]
Just realized, we should probably introduce the distinction between "closing the response" and "closing underlying entity", I think these two are tangled right now in `ResponseImpl` under `close()`: the only thing "closing the response" does is "closing underlying entity".

**Comment by andymc12** [comment]
@reta I see your point, and I agree that this change gets us closer to Jersey's behavior.  I think I don't like the overall architecture (JAX-RS spec, not CXF) as it basically says things are different on the client vs server.  For example:
```
Response r1 = Response.ok("foo").build();
assert "foo".equals(r1.readEntity(String.class));
assert "foo".equals(r1.getEntity()); // since foo is not backed by an input stream, it must be cached
// it's sort of like a safeguard in case you lost a reference to the entity but still have a reference to the Response

// but on the client side...
Response r2 = someTarget.request().get();
assert "foo".equals(r2.readEntity(String.class);
assert exceptionThrown(IllegalStateException.class, () -> r2.getEntity());

// further... (unless I'm misreading the Jersey code)
Response r3 = someTarget.request().get();
assert r3.bufferEntity();
assert "foo".equals(r3.readEntity(String.class);
assert r3.getEntity() instanceof InputStream; // what?! not String??
```

In some cases the `Response` object is a permanent holder of the entity, but in others it is not. And "permanent" might change depending on methods called on the `Response` object - other than `close`.  It might also depend on methods called on the stream... for example:
```
Response r4 = someTarget.request().get();
InputStream is = r4.readEntity(InputStream.class);
assert r4.getEntity() != null && r4.getEntity() instanceof InputStream; // can even call it twice because the stream is not consumed
fullyConsume(is);
assert exceptionThrown(IllegalStateException.class, () -> r4.getEntity());
```

Personally, I don't like this inconsistency.  I think that `Response` should either be a permanent (until closed) holder of entities or it should not be a holder at all (i.e. once `readEntity` is called, the response no longer has an entity - the caller must keep a reference to it - but the response doesn't have to be closed - callers could still check things like status code or headers).  But I think the current described behavior is confusing - for users and implementors.  

I'll plan to take this up with the JAX-RS community, but for now, I think this change works.  Thanks (and sorry for the rant! :) ).

andymc12 (approved):

reta:
😄  @andymc12 , thanks a lot for the rant, please count my voice in, I am all in for consistency and following the least surprise design, thanks a mill for review. ❤️

PR #870, opened 2021-11-01
CXF-8221: Upgrade Http Components Core and Client to 5.0

reta:
Upgrade Http Components Core and Client to 5.0 by introducing new module `transports/http-hc5`. Functionally, it is interchangeable with the Http Components Core and Client to 4.x (existing `transports/http-hc` module, except the OSGi part) and could be used as in-place replacement.  The implementation is largely ported from the `transports/http-hc` and adapted to the new APIs.

- [X] Remove unnecessary abstractions.
- [X] Add More Tests

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -301,6 +301,11 @@
             </dependency>
             <dependency>
                 <groupId>org.apache.cxf</groupId>
+                <artifactId>cxf-rt-transports-http-hc5</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.cxf</groupId>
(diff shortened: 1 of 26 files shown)


reta:
@ffang would do you think about this one, we are preserving HttpClient 4.x support and introducing HttpClient 5.x side by side (new transport module), would appreciate you taking a look, thank you!

ffang (approved):
LGTM!

ffang:
> @ffang would do you think about this one, we are preserving HttpClient 4.x support and introducing HttpClient 5.x side by side (new transport module), would appreciate you taking a look, thank you!

Yes, this new transport module is a good idea, so we still have HttpClient 4.x supported for CXF in OSGi container

PR #872, opened 2021-11-07
CXF-8616: Calling oneway methods using async client hangs response indefinitely

reta:
It turns out that when async client is used with oneway methods (in CXF context, oneway is the method that returns 202), the response hangs indefinitely due to the fact that the client callback is never called / propagated. Additionally, it turned out, for sync / async flows, the response filters are also not called for oneway methods.

- Introduced new contextual property `PROPAGATE_202_RESPONSE_ONEWAY_OR_PARTIAL`
- It is set to `false` by default but JAX-RS clients override it with `true`

### DIFF
--- a/core/src/main/java/org/apache/cxf/message/Message.java
+++ b/core/src/main/java/org/apache/cxf/message/Message.java
@@ -92,6 +92,13 @@ public interface Message extends StringMap {
      * Default value is true
      */
     String PROCESS_202_RESPONSE_ONEWAY_OR_PARTIAL = "org.apache.cxf.transport.process202Response";
+    
+    /**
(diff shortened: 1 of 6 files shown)


reta:
@ffang sorry but would appreciate if you could validate this one for me as well, I run into this issue while adding test cases for HttpClient 5.x support

ffang:
> @ffang sorry but would appreciate if you could validate this one for me as well, I run into this issue while adding test cases for HttpClient 5.x support, thank you

Hi @reta ,

Looks good to me.

Could you please add affected component in the original JIRA ticket as "JAXRS" since this only affects JAXRS service?

Thanks!
Freeman

PR #878, opened 2021-11-25
CXF-8619 Prevent double URL-decoding for form parameters where the pa…

jgallimore:
…rameters are read from HttpServletRequest.getParameter()

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/FormUtils.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/FormUtils.java
@@ -31,6 +31,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.HttpMethod;
@@ -175,8 +176,20 @@ public final class FormUtils {
             && MessageUtils.getContextualBoolean(m, FORM_PARAMS_FROM_HTTP_PARAMS, true)) {
             for (Enumeration<String> en = request.getParameterNames(); en.hasMoreElements();) {
                 String paramName = en.nextElement();
-                String[] values = request.getParameterValues(paramName);
-                params.put(HttpUtils.urlDecode(paramName), Arrays.asList(values));
+                String[] parameterValues = request.getParameterValues(paramName);
+
+                // these parameters will already be URLdecoded by the servlet container on the
+                // request.getParameterValues() call above
+
+                List<String> values = Arrays.asList(parameterValues);
+
+                if (!decode) {
+                    values = values.stream()
+                            .map(s -> HttpUtils.urlEncode(s, enc))
+                            .collect(Collectors.toList());
+                }
+
+                params.put(HttpUtils.urlDecode(paramName), values);
             }
             logRequestParametersIfNeeded(params, enc);
         }
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/FormUtils.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/FormUtils.java
@@ -31,7 +31,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.logging.Logger;
-import java.util.stream.Collectors;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.HttpMethod;
@@ -57,6 +56,7 @@ import org.apache.cxf.transport.http.AbstractHTTPDestination;
 public final class FormUtils {
     public static final String FORM_PARAMS_FROM_HTTP_PARAMS = "set.form.parameters.from.http.parameters";
     public static final String FORM_PARAM_MAP = "org.apache.cxf.form_data";
+    public static final String FORM_PARAM_MAP_DECODED = "org.apache.cxf.form_data.decoded";
 
     private static final Logger LOG = LogUtils.getL7dLogger(FormUtils.class);
     private static final String MULTIPART_FORM_DATA_TYPE = "form-data";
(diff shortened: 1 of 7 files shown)


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/JAXRSUtils.java
```diff
@@ -1153,6 +1153,17 @@ private static Object processFormParam(Message m, String key,
                 String body = FormUtils.readBody(entityStream, enc);
                 // Do not decode unless the key is empty value, fe @FormParam("")
                 FormUtils.populateMapFromStringOrHttpRequest(params, m, body, enc, StringUtils.isEmpty(key) && decode);
+
+                if (!StringUtils.isEmpty(key) && decode) {
```
reta, line 1157: I think this is change is not right: it applies only when `if (params == null)`, which holds `true` for first parameter only. This is the reason why it was located at the end of the function. PS: the test cases are failing because of that.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/FormUtils.java
```diff
@@ -175,8 +176,20 @@ public static void populateMapFromString(MultivaluedMap<String, String> params,
...
-                String[] values = request.getParameterValues(paramName);
-                params.put(HttpUtils.urlDecode(paramName), Arrays.asList(values));
+                String[] parameterValues = request.getParameterValues(paramName);
+
+                // these parameters will already be URLdecoded by the servlet container on the
+                // request.getParameterValues() call above
+
+                List<String> values = Arrays.asList(parameterValues);
+
+                if (!decode) {
```
reta, line 186: I think we should not encode the input. I understand that you are trying to mitigate the decoding issue(s), I am looking into the issue, will try to find out alternative solution.

reta:
Thank you for the contribution, @jgallimore, I have the concerns regarding this change, looking into the issue in order to understand the decoding flaws.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/JAXRSUtils.java
jgallimore: I see those test failures now, I'll take a look.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/FormUtils.java
jgallimore: I'm not a fan of the solution, which is probably a slightly bizarre comment to make on my own code. There is a scenario where CXF is reading parameters from HttpServetReques.getParameter(), where the parameter itself is annotated with `@Encoded`. In that scenario, I'm not sure how we can provide the parameter correctly without re-encoding.

jgallimore:
Thank you so much for the very quick review, @reta. I'm very happy to work with you and the team to come to the right solution on this. Please feel free to nudge me in the right direction :).

On thing I am wondering is whether the systest does a good job of highlighting the issue. I can certainly add some tests with an embedded Tomcat if that is useful.

reta:
Thanks @jgallimore 

> On thing I am wondering is whether the systest does a good job of highlighting the issue. I can certainly add some tests with an embedded Tomcat if that is useful.

I think it does, we continuously enrich it with new test cases and the fact it captured the regression is a good sign

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/FormUtils.java
rmannibucau: not using getParameter but the body inputstream - think cxf already has some part of it in body readers? - can solves it.

another side note is that moving the evaluation to be lazy instead of eager in a list can also be a good thing and avoid a lot of preprocessing (think `if (param1 is ok) { processBigParam(param2); }`, if param1 is not ok you don't care of having processed param2 and worse if you have some security checks in your endpoint (`@Context HttpServletRequest` is not that rare ;)). This can be done in a custom MultivaluedMap impl (new RequestParametersMultivaluedMap<>(request)).

Hope it makes sense.

reta: > not using getParameter but the body inputstream - think cxf already has some part of it in body readers? - can solves it.

Yes, the  body inputstream is handled right before

reta:
@jgallimore I think I understood the problem and I have a very simple solution to suggest, would you mind if I push the changes to your PR so you could take a look and provide the feedback? Thank you.

jgallimore:
@reta Yes please, that would be fantastic, thank you!

reta:
@jgallimore done, please let me know what do you think, thank you!

jgallimore:
Thanks @reta - this looks good to me, I'm not seeing a double decode happen for this case anymore. Do we want to try and figure out a way to enable @Encoded endpoints where we're reading from the HTTP parameter map? (I'm not sure its possible without re-encoding).

reta:
@jgallimore thank you

> Do we want to try and figure out a way to enable @encoded endpoints where we're reading from the HTTP parameter map? 

At least for me, the uncertainty here is that we either "trust" the request params as being properly decoded or not. If we do, that basically means @Encoded is noop in this flow - we don't know if the request params were encoded in first place or not - the servlet hides this from us and just provides the final value. Contrary, if we get rid of this "trust" logic (basically, drop this pull request), then by adding `@Encoded` annotation to form params, the double decoding is also gone (to be fair, I would prefer to use only `@Encoded` as means to control decoding behavior).

jgallimore:
From my perspective, I'd keep the PR as-is. If there's a need for the original encoded data, then its best to not consume the InputStream - instead leave it alone for CXF to parse.

Is this good to merge in, or does it need anything else? I'm very happy to help in any way.

reta:
@jgallimore thank you, sounds reasonable

> Is this good to merge in, or does it need anything else? I'm very happy to help in any way.

I would like to run JAX-RS TCK tests over this PR, itis a bit sensitive change and it is better to make sure we don't introduce new failures, will update you shortly

reta:
@jgallimore just finished the TCK run, no new test case failure, I think we are good to go once build checks are done, thank you!

PR #879, opened 2021-11-30
CXF-8620: AbstractHTTPServlet should support HTTP TRACE method dispatching if it is allowed by underlying container

reta:
A number of JAX-RS TCK of tests use HTTP TRACE to verify the client / server behavior, but TRACE HTTP method dispatching is not supported by AbstractHTTPServlet, causing tests do not fail.


reta:
@ffang mind please taking a look? Just run the JAX-RS TCKs, with this change we shaved off 20 test failures:

```
[javatest.batch] Test results: passed: 2,630; failed: 33 (-> 53 right now)
```

thank you!

ffang (approved):
LGTM, thanks!

PR #885, opened 2021-12-18
CXF-8630: Remove mockwebserver dependency, use wiremock instead

reta:
Right now there is one module which uses `mockwebserver` for test scaffolding, this is not really necessary since we already have `wiremock` dependency in place. One or another should be used, although `wiremock`  is used in most places.


reta:
@andymc12 if you don't mind taking a quick look, trying to reduce the amount of competing test dependencies, thank you

PR #892, opened 2022-01-18
[CXF-8641] NPE on NamePasswordCallbackHandler

globalbus:
https://issues.apache.org/jira/projects/CXF/issues/CXF-8641


ffang (approved):
Looks good to me!

PR #893, opened 2022-01-19
CXF-8642: ResponseImpl#hasEntity return 'false' when entity is buffered but entity stream is fully consumed with processing exception

reta:
`ResponseImpl#hasEntity` returns 'false' when entity is buffered but entity stream is fully consumed with processing exception. 

No new TCK test cases are failing:
```
Test results: passed: 2,631; failed: 32
```

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
@@ -152,6 +152,9 @@ public final class ResponseImpl extends Response {
         Object actualEntity = getActualEntity();
         if (actualEntity == null) {
             return false;
+        } else if (entityBufferred) {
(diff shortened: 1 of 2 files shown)


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/ResponseImpl.java
```diff
@@ -152,6 +152,9 @@ public boolean hasEntity() {
         Object actualEntity = getActualEntity();
         if (actualEntity == null) {
             return false;
+        } else if (entityBufferred) {
```
reta, line 155: When exception occurs while reading entity, it could happen that:
 -  getActualEntity() return `entity` (since `lastEntity` is `null`)
 - but the `entity` 's `InputStream` is fully consumed and is not empty
 - however, reading over fully consumed stream in this case is equivalent to reading over empty stream: nothing left to read

As such, the `hasEntity` return `false`. Adding the short circuit logic to make sure that if entity was buffered successfully, the response is considered to have an entity.

PR #894, opened 2022-01-20
Initial removal of Abdera

coheigea:
@reta Can you help to remove the Abdera parts from the CDI systests? It looks like if it's removed then the tests break completely.


reta (approved):
Thanks a lot, @coheigea , LGTM!

coheigea:
Thanks @reta

PR #895, opened 2022-01-25
Added test case for response entity with annotations

reta:
Added test case for response entity with annotations


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/interceptor/JAXRSOutInterceptor.java
```diff
@@ -156,8 +156,8 @@ private void serializeMessage(ServerProviderFactory providerFactory,
         Annotation[] responseAnns = response.getEntityAnnotations();
         if (responseAnns != null) {
             annotations = new Annotation[staticAnns.length + responseAnns.length];
-            System.arraycopy(staticAnns, 0, annotations, 0, staticAnns.length);
-            System.arraycopy(responseAnns, 0, annotations, staticAnns.length, responseAnns.length);
+            System.arraycopy(responseAnns, 0, annotations, 0, responseAnns.length);
```
reta, line 159: The order of annotations matters for TCK

PR #908, opened 2022-02-16
[CXF-8658] CXF 3.4.5 breaks binary compatibility with 3.4.1 in TypeClassInitializer

garydgregory:


rt/frontend/simple/src/main/java/org/apache/cxf/endpoint/dynamic/TypeClassInitializer.java
```diff
@@ -50,6 +50,15 @@
     boolean isFault;
     Bus bus;
 
+    @Deprecated // CXF-8658
+    public TypeClassInitializer(ServiceInfo serviceInfo,
```
reta, line 54: @garydgregory we cannot do that, it will blow with NPE in the `TypeClassInitializer::createFaultClass`

garydgregory: Well, crud :-(

reta: There are a few options, the `bus` instance should be taken from somewhere, for example if we bring constructor back, we need to introduce `setBus(Bus b)` method which has to be called, not ideal but a solution nonetheless.

Another option - use the bus from the thread local context using BusFactory::getThreadDefaultBus, the caller context should make sure it is set.

rmannibucau: or just sthg like (`bus != null? bus.getExtension( ExceptionCreator.class) : new ExceptionCreator()).createExceptionClass(cls);`)

reta: @rmannibucau afaik does not exist anymore, `ExceptionCreator` is replaced by `ExceptionClassGenerator` which also needs `bus` ... (because of ASM Helper).

rt/frontend/simple/src/main/java/org/apache/cxf/endpoint/dynamic/TypeClassInitializer.java
```diff
@@ -50,6 +50,15 @@
...
+            S2JJAXBModel model,
+            boolean allowWr) {
+        super(serviceInfo);
```
reta, line 57:
```suggestion
        this(BusFactory.getDefaultBus(), serviceInfo, model, allowWr);
```
I think this is the best we can do there ... :-\

rt/frontend/simple/src/main/java/org/apache/cxf/endpoint/dynamic/TypeClassInitializer.java
rmannibucau: Sounds like the bus is passed just for the extensions so replacing it by a `Function<Class<?>, Object>` (real extension "lookup") could make it trivial to replace wen bus is null, no?

gnodet:
Closing as the `3.4.x-fixes` branch is no longer maintained. The binary compatibility concern was valid at the time but is no longer relevant.

PR #923, opened 2022-03-11
Custom certificate alias not being used when using HC5 AsyncHTTPConduit

reta:
The `HttpContext` is not available in the `AsyncClientConnectionOperator`, so we have to provide our own TLS strategy on construction.


reta:
@ffang  sorry, another one for you please :)

ffang (approved):

ffang:
LGTM!

PR #926, opened 2022-03-21
[CXF-8680] Encapsulating getter method resolution in ClientProxyImpl

ilgrosso:


rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/ClientProxyImpl.java
```diff
@@ -407,6 +410,23 @@ protected static int getBodyIndex(MultivaluedMap<ParameterType, Parameter> map,
         return index;
     }
 
+    protected static Optional<Method> getBeanGetter(
```
reta, line 413: Super minor, may be just `getGetter` as a bit more generic name, but no strong opinion here. On the other subject, the method is `static` which prevents it to be overridden by subclasses, it is by design?

ilgrosso: No strong opinion either, about naming.
I've set it static since it is not accessing anything local in the instance. I wasn't aware there is some actual use of subclassing, no issues to remove static.

reta: Gotcha, let's make it `private` than, wdyt?

ilgrosso: After looking again at the code, I remember why did I set the new method as `protected static`: there are already a few like

* `protected static Annotation[] getMethodAnnotations(Method aMethod, int bodyIndex)`
* `protected static void reportInvalidResourceMethod(Method m, String name)`
* `protected static int getBodyIndex(MultivaluedMap<ParameterType, Parameter> map,                                    OperationResourceInfo ori)`

reta: 👍 sounds good, LGTM!

reta (approved):

ilgrosso:
Thanks for review @reta !

PR #929, opened 2022-03-30
OSGi: temporary exclude org/apache/cxf/transport/http/osgi/**.java from compilation

reta:
OSGi: temporary exclude org/apache/cxf/transport/http/osgi/**.java from compilation

### DIFF
--- a/rt/transports/http/pom.xml
+++ b/rt/transports/http/pom.xml
@@ -281,4 +281,30 @@
             </plugins>
         </pluginManagement>
     </build>
+    
+    <!-- TODO: Remove once OSGi is supported or / and extracted -->
+    <profiles>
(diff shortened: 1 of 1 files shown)


rt/transports/http/pom.xml
```diff
@@ -281,4 +281,30 @@
             </plugins>
         </pluginManagement>
     </build>
+    
+    <!-- TODO: Remove once OSGi is supported or / and extracted -->
```
reta, line 285: @jimma unblocking `transport-http` by excluding the OSGi classes from compilation (temporary), wdyt?

jimma: @reta  This looks good to me.  Thanks.

PR #930, opened 2022-03-30
CXF-8682: x-forwarded-proto header causes cxf to not match request to java method

reta:
CXF-8682: x-forwarded-proto header causes cxf to not match request to java method


reta:
@ffang mind please taking a quick look? thank you!

PR #927, opened 2022-03-26
Merge Opensaml4 to Jakarta EE branch

reta:
Merge Opensaml4 to Jakarta EE branch

PR #931, opened 2022-04-01
Temporary exclude modules which still depend on 'javax.*' by displacing them to separate 'javax' profile

reta:
Temporary exclude modules which still depend on 'javax.*' by displacing them to separate 'javax' profile

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -229,7 +229,7 @@
         <cxf.woodstox.core.version>6.2.7</cxf.woodstox.core.version>
         <cxf.woodstox.stax2-api.version>4.2.1</cxf.woodstox.stax2-api.version>
         <cxf.wsdl4j.version>1.6.3</cxf.wsdl4j.version>
-        <cxf.jakarta.wss4j.version>3.0.0-SNAPSHOT</cxf.jakarta.wss4j.version>
+        <cxf.jakarta.wss4j.version>2.4.1</cxf.jakarta.wss4j.version>
         <cxf.xalan.version>2.7.2</cxf.xalan.version>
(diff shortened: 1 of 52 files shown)


reta:
@jimma could you please check it out? removing (temporary) modules which do not comply with `jakarta.*` and fail compilation right now, thanks!

parent/pom.xml
```diff
@@ -229,7 +229,7 @@
         <cxf.woodstox.core.version>6.2.7</cxf.woodstox.core.version>
         <cxf.woodstox.stax2-api.version>4.2.1</cxf.woodstox.stax2-api.version>
         <cxf.wsdl4j.version>1.6.3</cxf.wsdl4j.version>
-        <cxf.jakarta.wss4j.version>3.0.0-SNAPSHOT</cxf.jakarta.wss4j.version>
+        <cxf.jakarta.wss4j.version>2.4.1</cxf.jakarta.wss4j.version>
```
reta, line 232: @jimma I reverted that for now (there are artifacts we use but non-existing in 3.0.0-SNAPSHOT) so we could hopefully address that as part of https://github.com/apache/cxf/pull/927

jimma: @reta  OK. After the this security failure is fixed, we can ask for a wss4j jakarta release to include it in CXF.

jimma:
> @jimma could you please check it out? removing (temporary) modules which do not comply with `jakarta.*` and fail compilation right now, thanks!

@reta  This looks good to me.  Thanks for working this out.

parent/pom.xml
reta: 👍

PR #937, opened 2022-04-19
[CXF-8695]:Remove duplicate copied class from cxf-core in cxf-rt-mana…

jimma:
…gement


rt/management/pom.xml
```diff
@@ -156,18 +156,6 @@
                             </execution>
                         </executions>
                     </plugin>
-                    <plugin>
```
reta, line 159: @jimma I think we need this plugin here, the one which copies files is maven-compiler-plugin above

jimma: @reta  I confirmed that maven bundle-plugin copied these classes and specifically this configured instruction does this job : 
```
                            <instructions>
                                <Export-Package>org.apache.cxf.management*,org.apache.cxf.tracing*</Export-Package>
                            </instructions>
```

ffang: Hi @jimma ,

Please change as this way
```
<configuration>
                             <obrRepository>NONE</obrRepository>
                             <instructions>
-                                <Export-Package>org.apache.cxf.management*,org.apache.cxf.tracing*</Export-Package>
+                                <Export-Package>
+                                                org.apache.cxf.management.counters,
+                                                org.apache.cxf.management.interceptor,
+                                                org.apache.cxf.management.persistence,
+                                                org.apache.cxf.management.utils,
+                                                org.apache.cxf.management.jmx,
+                                                org.apache.cxf.management.jmx.export.runtime,
+                                                org.apache.cxf.tracing*
+                                </Export-Package>
                             </instructions>
                             <niceManifest>true</niceManifest>
                         </configuration>
```

The root cause of this issue is that the previous org.apache.cxf.management* isn't specific enough, it include package from core, we should use the more fine-grained export package of cxf-rt-management module to avoid it.

ffang (approved):

reta (approved):

PR #942, opened 2022-04-30
[CXF-8697] Fix order-dependent flaky tests

jchen8460:
**Issue**
[Jira Issue](https://issues.apache.org/jira/browse/CXF-8697?jql=project%20%3D%20CXF%20AND%20resolution%20%3D%20Unresolved%20AND%20text%20~%20%22CachedOutputStream%22%20ORDER%20BY%20priority%20DESC%2C%20updated%20DESC)

The tests `org.apache.cxf.io.CacheAndWriteOutputStreamTest.testDeleteTmpFile` and `org.apache.cxf.io.CachedOutputStreamTest.testDeleteTmpFile` fail when run by themselves with the command `mvn test` because the temporary file is null. The reason is because the method `createFileOutPutStream` is never called in the method `enforceLimits` in `CachedOutputStream` since the threshold instance variable is not greater than the totalLength variable. The threshold variable depends on the static variable defaultThreshold. The defaultThreshold also depends on the system property. Therefore, these tests pass when the whole test suite is run because a different test, such as `testUseSysProps`, changes the static variable along with the system property before `testDeleteTmpFile` is run.

**Solution**
In order to allow these two tests to pass on their own, the system property must first be set to "4". Next, the static variables need to be updated with a call to `reloadDefaultProperties`. Since the previous version of these tests do not change system properties and static variables, the test should then reset the system property to the original state once finished executing. Please let me know if you would like me to reset the state differently.

### DIFF
--- a/core/src/test/java/org/apache/cxf/io/CachedStreamTestBase.java
+++ b/core/src/test/java/org/apache/cxf/io/CachedStreamTestBase.java
@@ -69,16 +69,27 @@ public abstract class CachedStreamTestBase {
 
     @Test
     public void testDeleteTmpFile() throws IOException {
-        Object cache = createCache();
-        //ensure output data size larger then 64k which will generate tmp file
-        String result = initTestData(65);
-        File tempFile = getTmpFile(result, cache);
-        assertNotNull(tempFile);
-        //assert tmp file is generated
-        assertTrue(tempFile.exists());
-        close(cache);
-        //assert tmp file is deleted after close the CachedOutputStream
-        assertFalse(tempFile.exists());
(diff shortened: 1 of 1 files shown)


amarkevich:
or use explicit threshold value in the test instead:
`Object cache = createCache(64 * 1024);`

jchen8460:
Yes, that would work as well. I suggested the other way because almost all the tests use the default `createCache()`, so I thought that the test was trying to test the system property and the threshold as well. Since that is not the case, I have made a new commit that uses your suggestion. Please let me know if you have any other comments.

amarkevich:
@jchen8460 its fine, thanks!

PR #944, opened 2022-05-02
NPE fix in addClaims() method

lukepasek:
In AbstractSTSClient NPE is thrown when instance claims is null, and claimsCallbackHandler returns claims as ClaimCollection.
In the original code the wrong variable was cast on the claimCollection local variable, was claims and should be claimsToSerialize.

### DIFF
--- a/rt/ws/security/src/main/java/org/apache/cxf/ws/security/trust/AbstractSTSClient.java
+++ b/rt/ws/security/src/main/java/org/apache/cxf/ws/security/trust/AbstractSTSClient.java
@@ -1453,7 +1453,7 @@ public abstract class AbstractSTSClient implements Configurable, InterceptorProv
         if (claimsToSerialize instanceof Element) {
             StaxUtils.copy((Element)claimsToSerialize, writer);
         } else if (claimsToSerialize instanceof ClaimCollection) {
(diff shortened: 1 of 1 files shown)


rt/ws/security/src/main/java/org/apache/cxf/ws/security/trust/AbstractSTSClient.java
```diff
@@ -1453,7 +1453,7 @@ protected void addClaims(XMLStreamWriter writer) throws Exception {
         if (claimsToSerialize instanceof Element) {
             StaxUtils.copy((Element)claimsToSerialize, writer);
         } else if (claimsToSerialize instanceof ClaimCollection) {
-            ClaimCollection claimCollection = (ClaimCollection)claims;
+            ClaimCollection claimCollection = (ClaimCollection)claimsToSerialize;
```
reta, line 1456: Thanks for the fix @lukepasek , seems like we have a gap in tests, could you please add at least one to capture that? Thank you!

coheigea: I'll add the test to the STS systests

coheigea (approved):

PR #950, opened 2022-05-14
CXF-8698: Content-ID of attachments for outgoing requests are URL-decoded instead of URL-encoded

reta:
RFC-2392 (https://datatracker.ietf.org/doc/html/rfc2392) says:
```
  content-id = url-addr-spec
  url-addr-spec = addr-spec ; URL encoding of RFC 822 addr-spec
```

RFC-822 addr-spec (https://datatracker.ietf.org/doc/html/rfc822#appendix-D) says:
```
    addr-spec = local-part "@" domain ; global address
```

In the same vein, as RFC-2392 (https://datatracker.ietf.org/doc/html/rfc2392) says:
```
A "cid" URL is converted to the corresponding Content-ID message
header [MIME] by removing the "cid:" prefix, converting the % encoded
character to their equivalent US-ASCII characters, and enclosing the
remaining parts with an angle bracket pair, "<" and ">".  
```

### DIFF
--- a/core/src/main/java/org/apache/cxf/attachment/AttachmentSerializer.java
+++ b/core/src/main/java/org/apache/cxf/attachment/AttachmentSerializer.java
@@ -25,6 +25,7 @@ import java.io.OutputStream;
 import java.io.StringWriter;
 import java.io.Writer;
 import java.net.URLDecoder;
+import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.Iterator;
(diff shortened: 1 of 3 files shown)


reta:
@deki @ffang @dkulp guys may I ask you please to take a look? Attempt to make Content-ID behavior more compliant to the specification (+ fixing some bug reports), thank you!

deki:
Thanks for looking into that @reta. I remember I was looking into a proposed fixed a few years back and missed the errata doc for the RFC: https://www.rfc-editor.org/errata/rfc2392
To me the change looks good but we don't have that many testcases covering all the use-cases.

I noticed there is another issue related to cid: https://issues.apache.org/jira/browse/CXF-8706 
We should probably fix it within the same release.

reta:
> To me the change looks good but we don't have that many testcases covering all the use-cases.

Thanks @deki , I added a few but still the coverage indeed could be improved

> I noticed there is another issue related to cid: https://issues.apache.org/jira/browse/CXF-8706

Yes, I saw this one, I will take care of it once we this fix gets in, thank you.

mash-sap:
Yes, this change is causing outgoing content-ids to be ENCODED. But unfortunately it also breaks the DECODE of incoming message. Content-ID headers do not contain "cid:-prefix as assumed here:
https://github.com/apache/cxf/pull/950/commits/221456aa5a515589b3bcda07a879de1b6e5e458c#diff-e3efb80d0a98bbbd7f6eddd3c021c5fb5ab05ea2ee8d97dc68026f6345e5a509R227

So after picking this change, I can see that outgoing headers are encoded but incoming messages are not decoded anymore. So I think there has to be a different condition to decide between en/decoding.

reta:
Thanks @mash-sap , may I ask you please for to confirm, I am trying to understand the case: with this change non-`cid:` prefixed `Content-ID` header is encoded on serializer side and **is not** decoded on deserializer side (when header extraction happens), correct?

mash-sap:
Eaxctly @reta.

Outgoing Content-ID is encoded, but incoming message Content-ID is not decoded.

reta:
@mash-sap thanks for heads up, I will take a look shortly, we may fix that by properly decoding  in the deserializer.

PR #954, opened 2022-06-08
CXF-8712 One-off issue with retries

kwin:
Immediately switch to the next alternate (address or endpoint) if the
current one has already exceeded the maxNumberOfRetries

### DIFF
--- a/rt/features/clustering/src/main/java/org/apache/cxf/clustering/RetryStrategy.java
+++ b/rt/features/clustering/src/main/java/org/apache/cxf/clustering/RetryStrategy.java
@@ -41,6 +41,7 @@ public class RetryStrategy extends SequentialStrategy {
         return getEndpoints(exchange, stillTheSameAddress());
     }
 
+    @Override
     protected <T> T getNextAlternate(List<T> alternates) {
         // is the amount of retries for the first alternate already exceeded?
         if (!stillTheSameAddress()) {
--- a/rt/features/clustering/src/main/java/org/apache/cxf/clustering/RetryStrategy.java
+++ b/rt/features/clustering/src/main/java/org/apache/cxf/clustering/RetryStrategy.java
@@ -42,7 +42,11 @@ public class RetryStrategy extends SequentialStrategy {
     }
 
     protected <T> T getNextAlternate(List<T> alternates) {
-        return stillTheSameAddress() ? alternates.get(0) : alternates.remove(0);
+        // is the amount of retries for the first alternate already exceeded?
+        if (!stillTheSameAddress()) {
+            alternates.remove(0);
+        }
+        return alternates.isEmpty() ? null : alternates.get(0);
     }
 
     protected boolean stillTheSameAddress() {
--- a/rt/features/clustering/src/main/java/org/apache/cxf/clustering/AbstractStaticFailoverStrategy.java
+++ b/rt/features/clustering/src/main/java/org/apache/cxf/clustering/AbstractStaticFailoverStrategy.java
@@ -73,12 +73,14 @@ public abstract class AbstractStaticFailoverStrategy implements FailoverStrategy
      * Select one of the alternate addresses for a retried invocation.
      *
      * @param alternates a List of alternate addresses if available
-     * @return the selected address
+     * @return the selected address or {@code null} if no alternate address is available
      */
     public String selectAlternateAddress(List<String> alternates) {
         String selected = null;
         if (alternates != null && !alternates.isEmpty()) {
             selected = getNextAlternate(alternates);
+        }
+        if (selected != null) {
             Level level = getLogLevel();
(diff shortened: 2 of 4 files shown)


rt/features/clustering/src/main/java/org/apache/cxf/clustering/RetryStrategy.java
```diff
@@ -42,7 +42,11 @@ public List<Endpoint> getAlternateEndpoints(Exchange exchange) {
     }
 
     protected <T> T getNextAlternate(List<T> alternates) {
-        return stillTheSameAddress() ? alternates.get(0) : alternates.remove(0);
+        // is the amount of retries for the first alternate already exceeded?
+        if (!stillTheSameAddress()) {
+            alternates.remove(0);
+        }
+        return alternates.isEmpty() ? null : alternates.get(0);
```
reta, line 49: This is dangerous one, `getNextAlternate()` is not supposed to return `null` and is calling for trouble [1], we should accommodate that.

[1] https://github.com/apache/cxf/blob/master/rt/features/clustering/src/main/java/org/apache/cxf/clustering/AbstractStaticFailoverStrategy.java#L113

reta:
Thank you for the fix, @kwin , I think you are right, there is max retries + 1 attempts actually being made, we should better respect this setting.

rt/features/clustering/src/main/java/org/apache/cxf/clustering/RetryStrategy.java
kwin: Any other idea how to fix without returning null? Or do we need to change the contract for of selectNextAlternate to allow null return values?

reta: This is a great question, the suggestion I have in mind is not 100% bullet proof but I would love to hear your opinion:

```
protected <T> T getNextAlternate(List<T> alternates) {
        String lastUsedAddress = null;
        // is the amount of retries for the first alternate already exceeded?
        if (!stillTheSameAddress()) {
            lastUsedAddress  = alternates.remove(0);
        }
        return alternates.isEmpty() ? lastUsedAddress : alternates.get(0);
}
```

In this case we would only go off the rails when there are no more alternatives to try out. WDYT?

kwin: This won't work for me, as I actually use this strategy with only one endpoint, so I would always run into the edge case.

reta: > This won't work for me, as I actually use this strategy with only one endpoint, so I would always run into the edge case.

Agree. Just a suggestion, in this case you actually have an easy workaround (not a solution, just temporary shortcut): reduce the number of max retries by 1. I will try to look for a better fix meantime, will update you shortly.

kwin: Thanks, if you modify the SPI anyways maybe you can also consider my comment from https://issues.apache.org/jira/browse/CXF-8288. Would be nice to always pass the message/exchange before selecting an alternate.

reta: @kwin for this particular issue (have not looked into CXF-8288) we should be able to keep SPI unchanged by documenting `getNextAlternate` could return `null` and making sure `AbstractStaticFailoverStrategy` could deal with `null` in `selectAlternateXxx` methods.

coheigea:
@reta What's the status of this PR?

reta:
> @reta What's the status of this PR?

@coheigea It looks abandoned, in general we need to make changes in a few places to prevent NPEs from happening

coheigea:
> > @reta What's the status of this PR?
> 
> @coheigea It looks abandoned, in general we need to make changes in a few places to prevent NPEs from happening

Thanks @reta , in that case I removed the fix versions from the corresponding JIRA.

kwin:
@reta What is necessary for this PR to be merged? Just https://github.com/apache/cxf/pull/954#discussion_r895181297 or anything else?

reta:
> @reta What is necessary for this PR to be merged? Just [#954 (comment)](https://github.com/apache/cxf/pull/954#discussion_r895181297) or anything else?

@kwin just that, thank you

rt/features/clustering/src/main/java/org/apache/cxf/clustering/RetryStrategy.java
kwin: Done in https://github.com/apache/cxf/pull/954/commits/cfca0e575821f577a9034d33d9f15b1638d94be2.

reta (approved):

PR #960, opened 2022-06-19
CXF-8706: CXF MTOM handler allow content injection

reta:
CXF MTOM handler allow content injection

### DIFF
--- a/core/src/main/java/org/apache/cxf/attachment/AttachmentUtil.java
+++ b/core/src/main/java/org/apache/cxf/attachment/AttachmentUtil.java
@@ -576,9 +576,9 @@ public final class AttachmentUtil {
                 return loadDataSource(contentId, atts);
             } else {
                 try {
-                    final String followUrls = SystemPropertyAction
-                        .getProperty(ATTACHMENT_XOP_FOLLOW_URLS_PROPERTY, "false");
-                    if ("true".equals(followUrls)) {
+                    final boolean followUrls = Boolean.valueOf(SystemPropertyAction
+                        .getProperty(ATTACHMENT_XOP_FOLLOW_URLS_PROPERTY, "false"));
+                    if (followUrls) {
                         return new URLDataSource(new URL(contentId));
                     } else {
                         return loadDataSource(contentId, atts);
--- a/core/src/main/java/org/apache/cxf/attachment/AttachmentUtil.java
+++ b/core/src/main/java/org/apache/cxf/attachment/AttachmentUtil.java
@@ -546,24 +546,40 @@ public final class AttachmentUtil {
     }
 
     public static DataSource getAttachmentDataSource(String contentId, Collection<Attachment> atts) {
-        // Is this right? - DD
+        //
+        // RFC-2392 (https://datatracker.ietf.org/doc/html/rfc2392) says:
+        //
+        // A "cid" URL is converted to the corresponding Content-ID message
+        // header [MIME] by removing the "cid:" prefix, converting the % encoded
+        // character to their equivalent US-ASCII characters, and enclosing the
+        // remaining parts with an angle bracket pair, "<" and ">".  
+        //
         if (contentId.startsWith("cid:")) {
             try {
                 contentId = URLDecoder.decode(contentId.substring(4), StandardCharsets.UTF_8.name());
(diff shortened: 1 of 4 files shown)


reta:
@deki wondering if you have time to check this one out, thank you!

dkulp:
I don't think this is the "right" fix for this.   The Soap w/ Attachement spec allows for http URL's for the cid and actually shows that use in the examples ( https://www.w3.org/TR/SOAP-attachments/ ).  Thus, I think the better fix would be to pass a flag into LazyDataSource to allow flipping to URLDataSource if the CID isn't present as a direct attachment.

reta:
> Thus, I think the better fix would be to pass a flag into LazyDataSource to allow flipping to URLDataSource if the CID isn't present as a direct attachment.

@dkulp thanks, I think I got your idea, but it solves a bit different problem: it does not help with the original issue, namely to limit href scheme `cid:` usage for `xop:include`, unless I am missing something ...

dkulp:
The main issue is that SOAP w/ Attachments doesn't have that limitation.  Thus, applying that limitation in AttachmentUtils is not the right place unless an extra flag is passed in to enable/disable the check and that's just set for the xop cases.

reta:
> The main issue is that SOAP w/ Attachments doesn't have that limitation. Thus, applying that limitation in AttachmentUtils is not the right place unless an extra flag is passed in to enable/disable the check and that's just set for the xop cases.

Exactly, I am looking at how to distinguish xop / non-xop cases, this information is "lost" when MTOM decorator calls `AttachmentUnmarshaller::getAttachmentAsDataHandler` (if you have something in mind, please share), thanks!

reta:
@dkulp relied on `isXOPPackage()` flag of the `AttachmentUnmarshaller` to distinguish xop / non-xop cases, what do you think of it? Thank you!

core/src/main/java/org/apache/cxf/attachment/AttachmentUtil.java
```diff
@@ -546,24 +546,40 @@ public static Attachment createMtomAttachmentFromDH(
...
+            // See please https://www.w3.org/TR/xop10/
+            //
+            if (contentId.indexOf("://") == -1) {
+                return loadDataSource(contentId, atts);
+            } else {
+                try {
+                    return new URLDataSource(new URL(contentId));
+                } catch (MalformedURLException e) {
+                    throw new Fault(e);
+                }
```
coheigea, line 578: Hi Andriy, let's control this behavior with a system property, where it only allows to read message attachments by default.

coheigea (changes requested):

core/src/main/java/org/apache/cxf/attachment/AttachmentUtil.java
```diff
@@ -546,24 +550,46 @@ public static Attachment createMtomAttachmentFromDH(
...
+            // 
+            // See please https://www.w3.org/TR/xop10/
+            //
+            if (contentId.indexOf("://") == -1) {
+                return loadDataSource(contentId, atts);
+            } else {
+                try {
+                    final String followUrls = SystemPropertyAction
+                        .getProperty(ATTACHMENT_XOP_FOLLOW_URLS_PROPERTY, "false");
+                    if ("true".equals(followUrls)) {
```
coheigea, line 581: Maybe use Boolean.valueOf(followUrls) instead to allow for someone setting "True"?

coheigea (changes requested):

core/src/main/java/org/apache/cxf/attachment/AttachmentUtil.java
reta: Certainly, thanks @coheigea , changed

coheigea: Did you push the update @reta ?
Also this test is failing for me:

[ERROR] Failures:
[ERROR]   MtomServerTest.testURLBasedAttachment:196 expected:<1> but was:<0>
[INFO]

reta: Oops, sorry, just did @coheigea (was fixing exactly this test)

coheigea (approved):
LGTM

PR #964, opened 2022-06-21
CXF-8725: Allow RetryStrategy to optionally not retry for 404 responses

kwin:
"org.apache.cxf.transport.service_not_available" for 404 responses

### DIFF
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
@@ -32,6 +32,7 @@ import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
```diff
@@ -1600,7 +1605,7 @@ protected int doProcessResponseCode() throws IOException {
             }
             if (exchange != null) {
                 exchange.put(Message.RESPONSE_CODE, rc);
-                if (rc == 404 || rc == 503 || rc == 429) {
+                if ((rc == 404 && MessageUtils.getContextualBoolean(outMessage, SERVICE_NOT_AVAILABLE_ON_HTTP_404, true)) || rc == 503 || rc == 429) {
```
reta, line 1608: Thanks @kwin , I am wondering if we could use list of status codes as contextual property instead (`404`, `503`, `429` being the default)?

kwin: Makes sense, I am gonna change the PR

kwin: Done in https://github.com/apache/cxf/pull/964/commits/191bea69cff1ff282d6565d0ba1e4b49daa6d95b

reta (approved):
Thank you, @kwin !

PR #971, opened 2022-07-19
restore TestUtilities#invokeBytes(String,String,byte[])

mpretzer:
In our Tests, we use `TestUtilities#invokeBytes(String,String,byte[])` to send dynamically generated content to our services.

This is no longer possible with CXF 3.5 since it was removed in b902794f187035a6efe39e3632ccc7be0adc8a89.

This PR reintroduces the method (and a variant accepting InputStreams).


amarkevich:
Since this code is not used in CXF tests - there is no reason to keep it in the sources.

PR #977, opened 2022-08-08
CXF-8746: Support force GET on redirections

ypiel-talend:
CXF-8746: Support force GET on 301 & 302 redirections for historical reasons as explained in https://www.rfc-editor.org/rfc/rfc7231.html section 6.4.2 & 6.4.3.

### DIFF
--- a/rt/transports/http-hc/src/main/java/org/apache/cxf/transport/http/asyncclient/AsyncHTTPConduit.java
+++ b/rt/transports/http-hc/src/main/java/org/apache/cxf/transport/http/asyncclient/AsyncHTTPConduit.java
@@ -130,10 +130,11 @@ public class AsyncHTTPConduit extends URLConnectionHTTPConduit {
     }
 
     @Override
-    protected void setupConnection(Message message, Address address, HTTPClientPolicy csPolicy) throws IOException {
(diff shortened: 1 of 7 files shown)


reta:
@ypiel-talend my apologies, I have difficulties to understand this change, the sections you are referring to note that:

>  Note: For historical reasons, a user agent MAY change the request
      method from POST to GET for the subsequent request.  If this
      behavior is undesired, the 307 (Temporary Redirect) status code
      can be used instead.

This is the behavior driven by user agent (client) and not the server, I am not sure why CXF needs change here. Also, seems like this is really outdated behavior (from notes on spec), why do we need to support it? What is the use case?

Thank you.

gnodet:
Superseded by #2952 with a cleaner, non-API-breaking approach. Instead of threading a `boolean forceGET` parameter through abstract method signatures across multiple transport implementations, the new PR uses a contextual property (`http.redirect.force.GET`) and sets `Message.HTTP_REQUEST_METHOD` on the outMessage before retransmit, requiring changes only in `HTTPConduit.java`.

gnodet:
Superseded by #2952 which implements this with a minimal, non-API-breaking approach using a contextual property.

PR #978, opened 2022-08-09
Fix Content ID and attachment ID equality check

atnyxp:
During deserialization of MTOM, received Content ID was URL decoded for LazyDataSource but attachment ID stayed URL encoded and failed the ID equality check in LazyDataSource#load. This patch provides the fix

### DIFF
--- a/core/src/main/java/org/apache/cxf/attachment/LazyDataSource.java
+++ b/core/src/main/java/org/apache/cxf/attachment/LazyDataSource.java
@@ -22,6 +22,9 @@ package org.apache.cxf.attachment;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/cxf/attachment/LazyDataSource.java
```diff
@@ -47,7 +50,13 @@ public LazyDataSource(String id, Collection<Attachment> attachments) {
     private synchronized void load() {
         if (dataSource == null) {
             for (Attachment a : attachments) {
-                if (a.getId().equals(id)) {
+                String attachmentId = a.getId();
```
reta, line 53: @atnyxp thanks for highlighting the issue, I think we should not pursue this route and fix the encoding issue (so `attachmentId` would match `id`), I will be looking into it shortly

neseleznev: It seems that this change https://github.com/apache/cxf/pull/1028 should fix it as well. Given the `attachmentId` which is valid domain, it should have no URL encode/decode issue in equality checks

coheigea: @reta Did you ever look into this again?

reta: @coheigea I suspect this change is not needed anymore, we've altered the way encode/decode is done

coheigea: Thanks @reta , let's close the PR

PR #979, opened 2022-08-11
[CXF-8749]:Existing handler file fails the compile generated java sou…

jimma:
…rce files

### DIFF
--- a/tools/common/src/main/java/org/apache/cxf/tools/common/ClassUtils.java
+++ b/tools/common/src/main/java/org/apache/cxf/tools/common/ClassUtils.java
@@ -21,6 +21,7 @@ package org.apache.cxf.tools.common;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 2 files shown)


tools/common/src/main/java/org/apache/cxf/tools/common/ClassUtils.java
```diff
@@ -97,8 +99,9 @@ public void compile(ToolContext context) throws ToolException {
 
                                 try {
                                     Files.copy(otherFile,
-                                        Files.createDirectories(Paths.get(targetDir, dirName)).resolve(str));
-
+                                            Files.createDirectories(Paths.get(targetDir, dirName)).resolve(str));
+                                } catch (FileAlreadyExistsException existsException) {
+                                    LOG.log(Level.WARNING, "EXIST_RESOURCE_FILE", otherFile);
```
reta, line 104: @jimma seems the message is not referring to the correct file: `otherFile` is a source and should exist. The target is `Files.createDirectories(Paths.get(targetDir, dirName)).resolve(str)` (probably needs a new variable)

jimma: @reta  You are right and fixed.Thanks for the review.

PR #981, opened 2022-08-12
CXF-8750: only add trailing slash if necessary when locating webjars …

awrb:
…resources for Swagger UI in OSGi environment

### DIFF
--- a/rt/rs/description-swagger-ui/src/main/java/org/apache/cxf/jaxrs/swagger/ui/OsgiSwaggerUiResolver.java
+++ b/rt/rs/description-swagger-ui/src/main/java/org/apache/cxf/jaxrs/swagger/ui/OsgiSwaggerUiResolver.java
@@ -22,7 +22,6 @@ import java.lang.annotation.Annotation;
 import java.net.URL;
 
 import org.osgi.framework.Bundle;
-import org.osgi.framework.FrameworkUtil;
 
 public class OsgiSwaggerUiResolver extends SwaggerUiResolver {
     private static final String DEFAULT_COORDINATES = "org.webjars/swagger-ui";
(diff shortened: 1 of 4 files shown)


rt/rs/description-swagger-ui/src/main/java/org/apache/cxf/jaxrs/swagger/ui/OsgiSwaggerUiResolver.java
```diff
@@ -32,18 +31,24 @@ public class OsgiSwaggerUiResolver extends SwaggerUiResolver {
     };
     
     private final Class<? extends Annotation> annotationBundle;
+    private FrameworkUtilWrapper frameworkUtilWrapper;
```
reta, line 34: I believe the the introduction of the wrapper is for testability reasons? I would suggest to go with integration tests instead, it would also help to catch any changes in Karaf (I could help with that).

reta:
Thanks for the pull request @awrb , the change looks good

rt/rs/description-swagger-ui/src/main/java/org/apache/cxf/jaxrs/swagger/ui/OsgiSwaggerUiResolver.java
awrb: Thanks for the review @reta. Yes, the wrapper is for testability as FrameworkUtil is static. But I agree pax exam tests are a better idea than mocking OSGi. I added some but I couldn't get them to work on the main branch, due to some dependencies missing such as spring 6.0.0-M5 (I couldn't find it in central maven repo either - not sure how I can fetch these). So I opened a new PR based on 3.5.x-fixes: https://github.com/apache/cxf/pull/982

reta: Thank you @awrb , yes, `main` is in transition to Jakarta 9.1+ support, we could close this one (I will port the fix from 3.5.x to all other branches), thanks again!

PR #987, opened 2022-08-18
feat(CXF-8752): list of redirectable verbs configurable.

ypiel-talend:
https://issues.apache.org/jira/browse/CXF-8752

### DIFF
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
@@ -37,13 +37,11 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.Executor;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.stream.Collectors;
 
 import javax.xml.namespace.QName;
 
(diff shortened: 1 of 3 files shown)


rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
```diff
@@ -1427,9 +1432,22 @@ private <T extends Exception> T mapException(String msg,
          * @throws IOException
          */
         protected void handleRetransmits() throws IOException {
+
+            String defaultVerbs = KNOWN_HTTP_VERBS_WITH_NO_CONTENT.stream().collect(Collectors.joining(","));
```
reta, line 1436: @ypiel-talend could you please add `getContextualStrings` to `MessageUtils` class (where we have a set of helper methods to deal with contextual properties):

```
    public static Set<String> getContextualStrings(Message m, String key, Set<String> defaultValue) {
        if (m != null) {
            Object o = m.getContextualProperty(key);
            if (o instanceof String) {
                Set<String> values = new TreeSet<>();
                for (String value : ((String) o).split(",")) {
                    if (!StringUtils.isEmpty(value)) {
                        values.add(value.trim());
                    }
                }
                return values;
            }
        }
        return defaultValue;
    }
```

In this case, your implementation would be just:

```
            final Set<String> allowedVerbsSet = MessageUtils
                    .getContextualStrings(outMessage, AUTHORIZED_REDIRECTED_HTTP_VERBS, 
                        KNOWN_HTTP_VERBS_WITH_NO_CONTENT);
```

Thank you.

reta:
> Looks OK to me, if we can add a test for it. @reta any major objections?

Left a comment, also agree that tests are needed, thanks @coheigea !

rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
ypiel-talend: Thanks, you're right.
Done : https://github.com/apache/cxf/pull/987/commits/af29d1e467a36726d4f8a6f1ee8eab5668f4806e

reta (approved):

PR #989, opened 2022-08-19
Make opaque access token generation extensible

arthurchan35:


arthurchan35:
In case of you wonder why is it necessary:
In our code, after CXF generates the opaque token, we have some additional customization code to apply to the cxf generated token.
Previously it was fine but now we are taking advantage of cxf's JWT access token generation, however we need our customization code go after opaque token generation but also before JWT token generation.
The only CLEAN way I see is to extract out the opaque token generation to a method.

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/provider/AbstractOAuthDataProvider.java
```diff
@@ -394,35 +419,31 @@ protected void linkRefreshTokenToAccessToken(RefreshToken rt, ServerAccessToken
...
         }
 
+        ServerAccessToken at =
+            doCreateAccessToken(
+                oldRefreshToken.getAudiences() != null ?
+                    new ArrayList<String>(oldRefreshToken.getAudiences()) : null,
+                client, oldRefreshToken.getClientCodeVerifier(),
+                oldRefreshToken.getExtraProperties(), oldRefreshToken.getGrantCode(),
+                oldRefreshToken.getGrantType(), oldRefreshToken.getNonce(),
+                oldRefreshToken.getResponseType(), theNewScopes,
```
coheigea, line 442: It looks like the response type was not set on the old access token on refresh - is it correct to set it here?

arthurchan35: Hi @coheigea ,
Thanks for spotting, I have removed it in:
https://github.com/apache/cxf/pull/989/commits/4b2c2a27b357fe07354fee6db9fe005e8e2a9a50

coheigea (approved):

PR #990, opened 2022-08-19
specs says type in header should at+jwt

arthurchan35:
https://datatracker.ietf.org/doc/html/rfc9068#section-2.1

### DIFF
--- a/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/provider/AbstractOAuthDataProvider.java
+++ b/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/provider/AbstractOAuthDataProvider.java
@@ -27,6 +27,7 @@ import java.util.Map;
 import jakarta.ws.rs.core.MultivaluedMap;
 import org.apache.cxf.jaxrs.ext.MessageContext;
 import org.apache.cxf.rs.security.jose.common.JoseConstants;
+import org.apache.cxf.rs.security.jose.jws.JwsHeaders;
 import org.apache.cxf.rs.security.jose.jwt.JwtClaims;
 import org.apache.cxf.rs.security.jose.jwt.JwtConstants;
 import org.apache.cxf.rs.security.jose.jwt.JwtToken;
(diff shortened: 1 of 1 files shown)


arthurchan35:
@amarkevich
@reta

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/provider/AbstractOAuthDataProvider.java
```diff
@@ -646,7 +647,12 @@ protected String processJwtAccessToken(JwtClaims jwtCliams) {
         // It will JWS-sign (default) and/or JWE-encrypt
         OAuthJoseJwtProducer processor =
             getJwtAccessTokenProducer() == null ? new OAuthJoseJwtProducer() : getJwtAccessTokenProducer();
-        return processor.processJwt(new JwtToken(jwtCliams));
+
+        JwsHeaders jwsHeaders = new JwsHeaders();
```
reta, line 651: @arthurchan35 it does not seem to be the solving the problem at large:
 - as you may see in the comment [1], it could be JWS or JWE
 - for `JWE`, the `JwsHeaders` are not used, the `JweHeaders` are 

Looking into the right place to apply the spec recommendation, but on more general note, we need to introduce  a member to `JoseType` for `at+JWT` and respective constant to `JoseConstants`.

[1] https://github.com/apache/cxf/pull/990/files#diff-1c24cdb27ac335b1f77f921093e723cfeeda77ce8e14d3196d2d1977a3d1effaR648

arthurchan35: Hi @reta, thanks for the review!

According to RFC 9068 sections 2.1 and section 4,  a JWT access token must be signed, **optionally** encrypted. As I interpret the comment, it means the same thing as specs required?

Regarding JoseType and JoseConstants, I will look into them bit more.

gnodet:
Superseded by #2948 which implements RFC 9068 at+jwt support using the JoseType enum approach as suggested by @reta.

PR #991, opened 2022-08-19
use getter in case of overwritten from child class

arthurchan35:


arthurchan35:
@amarkevich
@reta

reta:
Makes sense, thanks @arthurchan35 !

PR #992, opened 2022-08-19
Cxf 8710

awrb:


awrb:
The change is that `queryConfigEnabled` will automatically switch after calling either `getConfigParameters` or `isQueryConfigEnabled`, which is what `SwaggerUiService` will do.

It'd probably be cleanest if `SwaggerUiConfig` had to be constructed with a builder (so it could set  `queryConfigEnabled` if necessary in the final `build()` method, rather than relying on external calls) but that would be an API change.

systests/jaxrs/src/test/java/org/apache/cxf/systest/jaxrs/description/SwaggerUiConfigurationQueryConfigTest.java
```diff
@@ -57,7 +59,7 @@ protected org.apache.cxf.endpoint.Server createServer(Bus bus) throws Exception
             sf.setProvider(new JacksonJsonProvider());
             final Swagger2Feature feature = new Swagger2Feature();
             feature.setRunAsFilter(false);
-            feature.setSwaggerUiConfig(new SwaggerUiConfig().url("/swagger.json").queryConfigEnabled(true));
```
awrb, line 60: I removed the extra call since it should be set automatically. Now, with this removed, there would not be any relevant difference between this and `SwaggerUiConfigurationTest`, so perhaps this test would also be removed or some other difference should be introduced.

rt/rs/description-swagger-ui/src/main/java/org/apache/cxf/jaxrs/swagger/ui/SwaggerUiConfig.java
```diff
@@ -260,6 +261,10 @@ public SwaggerUiConfig tryItOutEnabled(boolean tryItOut) {
     }
 
     public Boolean isQueryConfigEnabled() {
+        if (!Boolean.TRUE.equals(queryConfigEnabled) && !getConfigParameters().isEmpty()) {
```
reta, line 264: @awrb this is a getter, which is not supposed to modify any other field, the `queryConfigEnabled` is 3 state property: `null`, `true`, `false`. We should only change this property implicitly when its value is not provided (means, it is `null`). If it is set to `true` or `false` already, we have to respect that.

systests/jaxrs/src/test/java/org/apache/cxf/systest/jaxrs/description/SwaggerUiConfigurationQueryConfigTest.java
reta: We could test the behavior when the `queryConfigEnabled` is set to `false` explicitly fe.

rt/rs/description-swagger-ui/src/main/java/org/apache/cxf/jaxrs/swagger/ui/SwaggerUiConfig.java
awrb: Good point @reta, where do you think is the best place to modify it implicitly? The other place I put the modification was in `getConfigParameters()`, but it's also a kind of getter (just an aggregative one) and you probably wouldn't expect it to modify fields. An alternative would be in each setter (other than  `setQueryConfigEnabled`), but the drawback would be some repetition, e.g:
```
 public void setUrl(final String url) {
        this.url = url;
        setQueryConfigEnabledIfNecessary();
    }

public void setMaxDisplayedTags(Integer maxDisplayedTags) {
        this.maxDisplayedTags = maxDisplayedTags;
        setQueryConfigEnabledIfNecessary();
    }

private void setQueryConfigEnabledIfNecessary() {
   // perhaps also check if a sensible, non-blank value was set?
   if (this.queryConfigEnabled == null) {
      setQueryConfigEnabled(true);
   }
}
```

reta: > An alternative would be in each setter

Thanks @awrb  I think repetition is not as bad, it is still isolated  to a function call (and this is one of the reasons setters are there for, to have an ability to introduce some hooks if necessary).

awrb:
Hi @reta, I updated the pull request after the code review.

reta (approved):

reta:
Thanks a lot, @awrb !

PR #993, opened 2022-08-29
CXF-8698: Content-ID of attachments for outgoing requests are URL-decoded instead of URL-encoded (limiting decoding only to % encoded characters)

reta:
Limiting decoding only to % encoded characters (as per  RFC-2392).

// A "cid" URL is converted to the corresponding Content-ID message
// header [MIME] by removing the "cid:" prefix, converting the % encoded
// character to their equivalent US-ASCII characters


reta:
@atnyxp would it fix your issue with `+` encoding / decoding?

PR #996, opened 2022-09-07
[CXF-8761] DigestAuthSupplier: Must not decode URL encoded URI parts

tfaller:
DigestAuthSupplier currently decodes URL encoded URI parts. This breaks digest auth because servers do digest auth based of the still encoded URI.


reta:
Thanks @tfaller , the change looks right to me

reta (approved):

PR #998, opened 2022-09-19
CXF-8764: Upgrade to Jetty 10

reta:
Upgrade to Jetty 10. Getting rid of legacy Jetty continuations and Wiremock (where applicable) since it is unlikely Jetty 10 is going to be supported (https://github.com/wiremock/wiremock/issues/1967)


rt/rs/microprofile-client/src/test/java/org/apache/cxf/microprofile/client/AsyncTest.java
```diff
@@ -18,79 +18,76 @@
...
 
-import com.github.tomakehurst.wiremock.common.FileSource;
-import com.github.tomakehurst.wiremock.extension.Parameters;
-import com.github.tomakehurst.wiremock.extension.ResponseTransformer;
-import com.github.tomakehurst.wiremock.http.Request;
-import com.github.tomakehurst.wiremock.http.Response;
-import com.github.tomakehurst.wiremock.junit.WireMockRule;
+import com.sun.net.httpserver.HttpExchange;
+import com.sun.net.httpserver.HttpHandler;
+import com.sun.net.httpserver.HttpServer;
```
coheigea, line 32: Will it work on other JDKs @reta  ? e.g. IBM JDK.

reta: @coheigea in case of OpenJ9 - yes, it is part of the `jdk.httpserver` module

PR #999, opened 2022-09-21
Remove osgi, blueprint and karaf from CXF

jimma:


ffang (approved):
LGTM!
Thanks Jim!

reta:
@jimma massive one, thanks a lot for doing it, a couple of questions / suggestion please:
 - for commented code snippets, could we mark them all with some unique label, like `// TODO: [OSGi+Jakarta] ...`, it will help us to identify all places in code which needs to be revisited
 - for META-INF manifests, do you think we could keep the `</cxf.osgi.export>`, `<cxf.osgi.import>` and others? I would certainly not make deployable (no Jakarta support for OSGi) but would save us some time when we would need to bring this back

WDYT? In any case, LGTM!

jimma:
Thanks for the review @reta 

> @jimma massive one, thanks a lot for doing it, a couple of questions / suggestion please:
> 
> * for commented code snippets, could we mark them all with some unique label, like `// TODO: [OSGi+Jakarta] ...`, it will help us to identify all places in code which needs to be revisited

Good idea. I already updated the comment. 

> * for META-INF manifests, do you think we could keep the `</cxf.osgi.export>`, `<cxf.osgi.import>` and others? It would certainly not make deployable (no Jakarta support for OSGi) but would save us some time when we would need to bring this back

From these META-INF manifests, we'll have to review and modify them when osgi jakarta is ready. For example, there is no need to add the jakarta.jws package and it has been already included in jakarta.xml.ws since 4.0.0(Jakarta EE10). There will be probably some changes in other package name . These osgi import/export menifests should be reviewed when we brings the osgi code back.   
```
        <cxf.osgi.import>
            jakarta.jws*;version="${cxf.osgi.jws.version}",
            jakarta.xml.ws*;version="${cxf.osgi.jakarta.xml.ws.version}",
            jakarta.xml.bind*;version="${cxf.osgi.jakarta.bind.version}"
        </cxf.osgi.import>
```

jimma:
@reta @ffang  I already squashed these changes to one commit.

reta (approved):

PR #1000, opened 2022-09-27
CXF-8714: Migration path for Swagger v1.x

reta:
It seems like Swagger 1.x is not planned to support Jakarta, we would need to drop this integration


reta:
@jimma @ffang could you please take a look guys, if it makes sense to you, thank you!

jimma:
@reta  Should these swagger test resource files under https://github.com/reta/cxf/tree/CXF-8714/systests/jaxrs/src/test/resources/org/apache/cxf/systest/jaxrs/description be removed too ?

deki (approved):
Makes sense to drop it for 1.x

reta:
> @reta LGTM . More test resource files we can probably remove them too ?
> 
>     * Files under https://github.com/reta/cxf/tree/CXF-8714/systests/jaxrs/src/test/resources/org/apache/cxf/systest/jaxrs/description
> 
>     * https://github.com/reta/cxf/blob/CXF-8714/systests/jaxrs/src/test/resources/files/swagger.properties.

@jimma absolutely, missed that, thank you for noticing, removed resources as well

coheigea:
LGTM, need to rebase

PR #1001, opened 2022-09-27
Fix ws-discovery test on non-M1 Macs

coheigea:


coheigea:
Closing after review, we will revert the original commit

PR #1003, opened 2022-09-30
Fix SortedArraySet remove, containsAll

knut-estalea:
Remove would fail when removing the not-lowest entry.
The implementation of `containsAll` was "containsAny".
Also tweaks `add` to use binary search + insert instead
of a full sort every time.

### DIFF
--- a/core/src/main/java/org/apache/cxf/common/util/SortedArraySet.java
+++ b/core/src/main/java/org/apache/cxf/common/util/SortedArraySet.java
@@ -39,7 +39,7 @@ import java.util.concurrent.atomic.AtomicReference;
  * If no data is stored in the Set, it uses very little memory.  The backing
  * array is created on demand.
  *
- * This class is primarly useful for stuff that will be setup at startup, but
+ * This class is primarily useful for stuff that will be setup at startup, but
  * then iterated over MANY times during runtime.
  *
  * @param <T>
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/cxf/common/util/SortedArraySet.java
```diff
@@ -71,25 +71,29 @@ private T[] newArray(int size) {
     }
 
     public boolean add(T o) {
```
knut-estalea, line 73: The changes to `add` aren't fixing any bug. It just seemed like a waste to do a full sort every time.

core/src/main/java/org/apache/cxf/common/util/SortedArraySet.java
```diff
@@ -99,9 +103,9 @@ public boolean addAll(Collection<? extends T> c) {
         return val;
     }
     public boolean containsAll(Collection<?> c) {
```
knut-estalea, line 105: This method is unused in the CXF code base as far as I can tell.

core/src/main/java/org/apache/cxf/common/util/SortedArraySet.java
```diff
@@ -138,20 +142,21 @@ public boolean remove(Object o) {
             return false;
         }
         int idx = Arrays.binarySearch(tmp, o);
-        if (idx != -1) {
```
knut-estalea, line 141: The bug was here (idx == -1). Every negative value means "not found".

reta: 👍

core/src/main/java/org/apache/cxf/common/util/SortedArraySet.java
reta: We probably could make it even better by returning as early as possible (upon finding first non-containing element)?

core/src/main/java/org/apache/cxf/common/util/SortedArraySet.java
reta: 👍

core/src/main/java/org/apache/cxf/common/util/SortedArraySet.java
reta: Since class is public, it could be used outside of CXF codebase ...

reta (approved):

reta:
Thank you @knut-estalea

PR #1007, opened 2022-10-09
Migrating cxf-wadl2java-plugin & cxf-tools-wadlto-jaxrs modules

reta:
Migrating `cxf-wadl2java-plugin` & `cxf-tools-wadlto-jaxrs` modules. Temporary excluding ` jaxb2-basics` extensions till the equivalent Jakarta replacement becomes available.

### DIFF
--- a/maven-plugins/pom.xml
+++ b/maven-plugins/pom.xml
@@ -37,16 +37,6 @@
         <module>wsdl-validator-plugin</module>
         <module>corba</module>
         <module>archetypes</module>
+        <module>wadl2java-plugin</module>
     </modules>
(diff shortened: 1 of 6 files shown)


reta:
@jimma what do you think about this? the `jaxb2-basics` is used for tests only (as extension) so I removed it till the replacement becomes available

jimma:
> @jimma what do you think about this? the `jaxb2-basics` is used for tests only (as extension) so I removed it till the replacement becomes available
This looks good.  Let's go back to check this after the jaxb2-basics jakarta version is available.

PR #1009, opened 2022-10-10
[CXF-8775] fixed 3 rt Messages.properties file containing text that i…

rsearls:
…s javax. but should be jakarta


reta:
Thank you @rsearls

reta (approved):

PR #1012, opened 2022-10-13
Include back jaxrs systests module

reta:
Include back `jaxrs` systests module


reta:
> Hi @reta , is there a reason we can't just re-use this class to allocate ports?
> 
> https://github.com/apache/cxf/blob/654444433f3b17123cbc00fd17ea1c10d7de7a9c/testutils/src/main/java/org/apache/cxf/testutil/common/TestUtil.java#L123

Ah right, thhanks @coheigea , no there are no reasons not to use it, changing ....

coheigea (commented):
There are some instances of "javax.ws" Strings in the source of these tests that need to be migrated.

reta:
> There are some instances of "javax.ws" Strings in the source of these tests that need to be migrated.

@coheigea thanks for looking, correct, I will try to make all `jaxrs` systests pass before merging in this pull request, thanks!

PR #1013, opened 2022-10-18
[CXF-8724]:Add codegen module back

jimma:


systests/wsdl_maven/codegen/src/it/cxf-4004/pom.xml
```diff
@@ -37,12 +37,12 @@
                         <artifactId>cxf-xjc-ts</artifactId>
                         <version>${cxf.xjc-utils.version}</version>
                     </dependency>
-                    <dependency>
+                    <!--dependency-->
```
reta, line 40: thanks @jimma, LGTM, could you please add reference to https://issues.apache.org/jira/browse/CXF-8774 here, once the replacement is available, will be easy to trace the changes back.

jimma: @reta  Do you mean change the jira id in the commit message ? Or add some comment to https://issues.apache.org/jira/browse/CXF-8774  ?

reta: @jimma just a comment, thanks, so we know why it is commented out

jimma: Done. Thanks for review @reta

reta (approved):

PR #1018, opened 2022-10-31
[CXF-8758]:Add wiremock-standalone test dependency to fix more systests

jimma:

PR #1019, opened 2022-11-02
Fixing Microprofile test suites (temporary adding javax.servlet dependency and downgrading to Jetty 9)

reta:
Fixing Microprofile test suites (temporary adding javax.servlet dependency and downgrading to Jetty 9), @jimma it may help you with https://issues.apache.org/jira/browse/CXF-8781

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -61,8 +61,8 @@
         <cxf.activemq.artemis.version>2.26.0</cxf.activemq.artemis.version>
         <cxf.ahc.version>2.12.3</cxf.ahc.version>
         <cxf.apacheds.version>2.0.0.AM26</cxf.apacheds.version>
-        <cxf.arquillian.version>1.1.14.Final</cxf.arquillian.version>
-        <cxf.arquillian.weld.container.version>2.0.1.Final</cxf.arquillian.weld.container.version>
+        <cxf.arquillian.version>1.7.0.Alpha10</cxf.arquillian.version>
+        <cxf.arquillian.weld.container.version>3.0.0.Final</cxf.arquillian.weld.container.version>
         <cxf.aspectj.version>1.9.1</cxf.aspectj.version>
         <cxf.assertj.version>3.22.0</cxf.assertj.version>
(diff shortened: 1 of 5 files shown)


jimma:
@reta From the CI result, there are still some test failures under  `cxf-systests-microprofile-async` and `cxf-systests-microprofile-nocdi` caused by ` java.lang.ClassNotFoundException: javax.json.JsonException`
The `wiremock` or `wiremock-standalone` dependency should both work for test scope ,  but it seems it needs to add more javax dependencies for `wiremock` dependency. 

For `cxf-systests-microprofile-weld`, I did the similar change in weld module to get more tests passed yesterday , but there are still some failures I still need to have a look: https://github.com/apache/cxf/pull/1018/commits/94ce0543f4167d75dc2c9352b124e8358aec72be. 
And the latest CI shows there are 23 failures(in my branch) to fix : https://ci-builds.apache.org/job/CXF/job/CXF-JDK17-PR/110/. Do you have an idea what I need to have a look ?

reta:
Thanks @jimma 

> The `wiremock` or `wiremock-standalone` dependency should both work for test scope , but it seems it needs to add more javax dependencies for `wiremock` dependency.

I tried to fix that but went into rabbit hole of dependency mess, probably we need to wait for wiremock to catch up ...

> Do you have an idea what I need to have a look ?

I haven't yet into the failures yet, but I will help you out on https://github.com/apache/cxf/pull/1018

jimma:
> Thanks @jimma
> 
> > The `wiremock` or `wiremock-standalone` dependency should both work for test scope , but it seems it needs to add more javax dependencies for `wiremock` dependency.
> 
> I tried to fix that but went into rabbit hole of dependency mess, probably we need to wait for wiremock to catch up ...
> 

+1, now adding `wiremock-standalone` seems a bit easier and it doesn't need to add much other dependencies.  This will make it easier too when we replace the `wiremock-standalone` with the new `wiremock` release.

> > Do you have an idea what I need to have a look ?
> 
> I haven't yet into the failures yet, but I will help you out on #1018

Thanks @reta.  Let's discuss there.

PR #1022, opened 2022-11-07
[CXF-8787]:Fix sample build failures

jimma:


distribution/src/main/release/samples/pom.xml
```diff
@@ -210,6 +216,49 @@
             </plugins>
         </pluginManagement>
     </build>
+    <repositories>
+        <repository>
```
reta, line 220: I think we could remove snapshots, it was temporary before the milestones became available

distribution/src/main/release/samples/pom.xml
```diff
@@ -210,6 +216,49 @@
...
+            <id>netflix.candidates</id>
+            <url>https://artifactory-oss.prod.netflix.net/artifactory/maven-oss-candidates</url>
+            <snapshots>
+                <enabled>false</enabled>
+            </snapshots>
+        </repository>
+    </repositories>
+    <pluginRepositories>
+        <pluginRepository>
```
reta, line 245: Same here, I think we could get rid of snapshots repo

distribution/src/main/release/samples/pom.xml
```diff
@@ -240,108 +289,8 @@
...
-                </dependency>
-            </dependencies>
-        </profile>
-        
-        <!-- 
-            TODO: Please remove, temporary repository settings in order to enable snapshots for
-            Spring Boot 3.0.0
-        -->
-        <profile>
-            <id>jakarta</id>
```
reta, line 297: @jimma The introduction of the `jakarta` profile is on purpose since we depend on milestone releases (and before - even on snapshots). In general, we should not have these repositories and once we have the Spring train releases (mid-November by plan), the `jakarta` profile could be phased away from main tree and samples. Hope it make sense.

distribution/src/main/release/samples/pom.xml
jimma: Done.

distribution/src/main/release/samples/pom.xml
jimma: Done.

distribution/src/main/release/samples/pom.xml
jimma: Thanks for the update @reta . I added back this profile with a better profile name `spring`. The https://issues.apache.org/jira/browse/CXF-8788 is created to remove this profile after the final release comes out in the coming weeks.

reta: @jimma the `jakarta` name is being used in Jenkins builds and the same profile is also present in [pom.xml](https://github.com/apache/cxf/blob/main/pom.xml#L426). Could you please rename it back to `jakarta` so we don't need to change jobs configurations? (fe we run `-Peverything,jakarta`)

jimma: @reta Thanks for review. Changed back . I saw this profile is set to `activeByDefault`. Do we still need to active this profile with -P flag ?

reta: I think when profiles are used explicitly (like fe `-P`), the `activeByDefault` is not respected (since it is not default anymore). If you are curious, it is simple to check:

```
$ mvn help:active-profiles
```

vs explicit profiles with `-P` (this command fails for me, the `jakarta` profile is not picked up anymore)

```
$ mvn help:active-profiles -Peverything
```

pom.xml
```diff
@@ -217,10 +217,9 @@
                             <artifactId>maven-compiler-plugin</artifactId>
                             <version>3.10.1</version>
                             <configuration>
-                                <source>${cxf.jdk.version}</source>
-                                <target>${cxf.jdk.version}</target>
                                 <maxmem>256M</maxmem>
                                 <fork>${cxf.compiler.fork}</fork>
+                                <release>11</release>
```
reta, line 222: Super minor, we could set it to `${cxf.jdk.version}` (it is 11)

jimma: @reta I wrongly committed some experimental stuff. Reverted.  I should changed this with another PR and JIRA issue.

distribution/src/main/release/samples/pom.xml
jimma: @reta I didn't know this and learned new thing.  Thanks.

reta (approved):

PR #1029, opened 2022-11-16
CXF-8791: Make the Spring / Spring Boot dependencies optional

reta:
CXF-8791: Make the Spring / Spring Boot dependencies optional

### DIFF
--- a/distribution/src/main/release/samples/corba/bank/pom.xml
+++ b/distribution/src/main/release/samples/corba/bank/pom.xml
@@ -59,6 +59,17 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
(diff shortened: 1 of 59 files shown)


asoldano (approved):

distribution/src/main/release/samples/corba/bank/pom.xml
```diff
@@ -59,6 +59,18 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
```
reta, line 62: @jimma @asoldano I have updated all samples which use Spring 6 / Spring Boot 3 to require JDK-17, otherwise they are not runnable

jimma (approved):
@reta  Great stuff to improve the usability!

PR #1038, opened 2022-12-07
Allow building with maven 4

gnodet:
This upgrade is important in order to be able to build CXF with maven 4.x.

### DIFF
--- a/systests/wsdl_maven/codegen/src/it/settings.xml
+++ b/systests/wsdl_maven/codegen/src/it/settings.xml
@@ -30,9 +30,11 @@ under the License.
                     <url>@localRepositoryUrl@</url>
                     <releases>
                         <enabled>true</enabled>
+                        <checksumPolicy>ignore</checksumPolicy>
                     </releases>
                     <snapshots>
                         <enabled>true</enabled>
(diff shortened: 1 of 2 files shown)


systests/wsdl_maven/codegen/src/it/settings.xml
```diff
@@ -30,9 +30,11 @@ under the License.
                     <url>@localRepositoryUrl@</url>
                     <releases>
                         <enabled>true</enabled>
+                        <checksumPolicy>ignore</checksumPolicy>
```
reta, line 33: Why the `checksumPolicy` needs a change?

gnodet: Because maven 4 enforces the checksum by default.  
Checksums are computed at deploy time, not install time, so they are never present in the local repository for artifacts that have been installed during a build.  But this repository is used as a fake remote repositories, for which maven 4 expects checksums by default.  So we just need to switch the default policy to ignore the fact that the checksums are not present.

reta: Got it, thank you for clarifying, I didn't know that

reta (approved):

PR #1040, opened 2022-12-11
Fix org.apache.cxf.transport.jms.RequestResponseTest test cases which fail with 'Can't receive the Conduit Message in 10 seconds'

reta:
Fix `org.apache.cxf.transport.jms.RequestResponseTest` test cases which fail with `Can't receive the Conduit Message in 10 seconds`


rt/transports/jms/src/test/java/org/apache/cxf/transport/jms/AbstractJMSTester.java
```diff
@@ -94,7 +94,7 @@ public static void startServices() throws Exception {
         addressSettings.setAutoCreateQueues(true);
         addressSettings.setAutoDeleteQueues(false);
         addressSettings.setAutoDeleteAddresses(false);
-        config.setAddressesSettings(Collections.singletonMap("#", new AddressSettings()));
+        config.setAddressSettings(Collections.singletonMap("#", new AddressSettings()));
```
reta, line 97: Not related (replaced deprecated API usage)

PR #1043, opened 2022-12-16
Some toString methods to help with debugging

dblevins:


coheigea:
@dblevins Can you rebase this please so that we can trigger the github action builder for it?

PR #1046, opened 2023-01-04
CXF-8810: Fix http-hc osgi startup

jmylly:


rt/transports/http-hc/src/test/java/org/apache/cxf/transport/http/asyncclient/ActivatorTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.cxf.Bus;
+import org.osgi.framework.BundleContext;
+
+import org.easymock.EasyMock;
+import org.junit.Test;
+
+public class ActivatorTest {
+
+    @Test
+    public void testConduitConfigurerUpdates() throws NoSuchMethodException {
```
reta, line 34: Thanks a lot for the fix, @jmylly, I think we could craft much simpler test case:

```
    @Test
    public void testConduitConfigurerUpdates() throws Exception {
        final Bus b = BusFactory.getThreadDefaultBus();
        final BundleContext context = EasyMock.mock(BundleContext.class);

        final ConduitConfigurer configurer = new ConduitConfigurer(context) {
            @Override
            public Object[] getServices() {
                return new Object[] {
                    b 
                };
            }
        };
        
        configurer.updated(new Hashtable<>());
    }
```

What do you think?

jmylly: Thanks, much cleaner than PartialMock. However, this test doesn't much verify the functionality. Pushed another version taking best of both.

reta (approved):

PR #1151, opened 2023-02-22
CXF-8821: Remove org.eclipse.tycho plugins since OSGi support is removed

reta:
Remove org.eclipse.tycho plugins since OSGi support is removed


jimma (approved):
Thanks for the cleanup @reta !

PR #1158, opened 2023-03-07
CXF-8822: AsyncHTTPConduit removes query-parameters when path is empty

reta:
AsyncHTTPConduit removes query-parameters when path is empty


rt/transports/http-hc/src/main/java/org/apache/cxf/transport/http/asyncclient/AsyncHTTPConduit.java
```diff
@@ -190,9 +190,17 @@ protected void setupConnection(Message message, Address address, HTTPClientPolic
...
         }
+
         if (StringUtils.isEmpty(uri.getPath())) {
-            //hc needs to have the path be "/"
-            uri = uri.resolve("/");
+            try {
+                //hc needs to have the path be "/"
+                uri = new URI(uri.getScheme(),
+                    uri.getUserInfo(), uri.getHost(), uri.getPort(),
+                    uri.resolve("/").getPath(), uri.getQuery(),
```
reta, line 199: The `URI::resolve` strips the query path off

PR #1239, opened 2023-04-18
[CXF-8837] check algorithm instead of class instance type

janbernhardt:
It would be great if this patch can also be backported to previous versions of CXF, down to 3.4.x.


coheigea:
FYI CXF 3.4.x is no longer maintained.

coheigea (approved):

PR #1246, opened 2023-04-21
CXF-8839

timsacre:


reta:
@timsacre thanks for the change, could you please change the commit message to `CXF-8839: Missing dependency on plexus-utils in cxf-wsdl-validator-plugin`. Also, shouldn't the same change be applied to other CXF Apache Maven plugins? Thank you.

timsacre:
@reta I've changed the commit message. Looks like most of the cxf plugins already have a dependency on plexus-utils.

PR #1299, opened 2023-06-12
[CXF-8886] Support scraping javadocs generated by Java 17

ilgrosso:


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/model/doc/JavaDocProvider.java
```diff
@@ -74,12 +83,12 @@ public JavaDocProvider(Bus bus, String... paths) throws Exception {
         }
     }
 
-    private static double getVersion() {
+    protected static double getVersion() {
```
reta, line 86: Could we keep this methods protected or there is a reason to open them up?

reta:
Thanks @ilgrosso , LGTM! Just a question regarding visibility change, keeping the APIs private would give us more freedom to change them

reta (approved):

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/model/doc/JavaDocProvider.java
ilgrosso: I tend to open up all methods for classes that might be extended by 3rd party apps; being also `static`, moreover, there should be no danger with override.
Anyway, I've reverted this.

ilgrosso:
@reta thanks. I've restored the original visibility as requested.

reta:
> @reta thanks. I've restored the original visibility as requested.

Thanks @ilgrosso , do you mind if we restore the original visibility level to other methods? (changed from `private` to `protected`), I could take care of it, just wanted to know if that's really needed, thank you!

PR #1301, opened 2023-06-13
[CXF-8886] Support scraping javadocs generated by Java 17 (#1299)

ilgrosso:


reta:
> Can you please have a look at the failed tests? I don't think it relates to this specific change.

Yeah, we sadly have a few flaky ones (https://issues.apache.org/jira/browse/CXF-8393), trying to fix them but for some of them is it very difficult to reproduce

reta:
> @reta ok so I will wait for this one once you have restored the `private` visibility as mentioned in #1299 so I'll include it here too

Should be in, thanks!

PR #1340, opened 2023-07-11
CXF-8457 Provide OpenTelemetry support

mederel:
A new module under integration/tracing is introduced to add support for tracing with OpenTelemetry (short OTel).

A few additional changes have been done:

* Add the OTel BOMs in the parent/pom.xml dependency management
* The new module is strongly inspired by the integration/tracing/tracing-opentracing module
* The tracing systest was enriched with tests dedicated to OpenTelemetry in the JaxRS and JaxWS contexts
* The necessary OTel dependencies are added to distribution/javadoc to be able to produce the documentation
* A new Jax-RS sample showcases span creation on client call and server call

### DIFF
--- a/distribution/src/main/release/samples/jax_rs/tracing_opentelemetry/pom.xml
+++ b/distribution/src/main/release/samples/jax_rs/tracing_opentelemetry/pom.xml
@@ -1,0 +1,167 @@
+<?xml version="1.0"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements. See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership. The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License. You may obtain a copy of the License at
+ 
+  http://www.apache.org/licenses/LICENSE-2.0
+ 
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied. See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>jax_rs_tracing_opentelelemtry</artifactId>
+    <name>JAX-RS Demo Using Distributed Tracing with OpenTelemetry</name>
+    <description>JAX-RS Demo Using Distributed Tracing with OpenTelemetry</description>
+    <parent>
+        <groupId>org.apache.cxf.samples</groupId>
+        <artifactId>cxf-samples</artifactId>
+        <version>4.0.3-SNAPSHOT</version>
+        <relativePath>../..</relativePath>
+    </parent>
+    <properties>
+        <cxf.version>${project.version}</cxf.version>
+    </properties>
+    <profiles>
+        <profile>
+            <id>server</id>
+            <build>
+                <defaultGoal>test</defaultGoal>
+                <plugins>
+                    <plugin>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>exec-maven-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <phase>test</phase>
+                                <goals>
+                                    <goal>java</goal>
+                                </goals>
+                                <configuration>
+                                    <mainClass>demo.jaxrs.tracing.server.Server</mainClass>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+        <profile>
+            <id>client</id>
+            <build>
+                <defaultGoal>test</defaultGoal>
+                <plugins>
+                    <plugin>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>exec-maven-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <phase>test</phase>
+                                <goals>
+                                    <goal>java</goal>
+                                </goals>
+                                <configuration>
+                                    <mainClass>demo.jaxrs.tracing.client.Client</mainClass>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.cxf</groupId>
+            <artifactId>cxf-rt-transports-http</artifactId>
+            <version>${cxf.version}</version>
+        </dependency>
+        <!-- This dependency is needed if you're using the Jetty container -->
+        <dependency>
+            <groupId>org.apache.cxf</groupId>
+            <artifactId>cxf-rt-transports-http-jetty</artifactId>
+            <version>${cxf.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.cxf</groupId>
+            <artifactId>cxf-rt-frontend-jaxrs</artifactId>
+            <version>${cxf.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.cxf</groupId>
+            <artifactId>cxf-rt-rs-client</artifactId>
+            <version>${cxf.version}</version>
+        </dependency>    
+        <dependency>
+            <groupId>org.apache.cxf</groupId>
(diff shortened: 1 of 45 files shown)


reta:
Thanks @mederel for the contribution, to help you with question a bit

> * Add the OTel BOMs in the parent/pom.xml dependency management (too much? should keep only the dependencies that are used?)

It is fine to add CXF deps to BOM (but not external deps like OTEL etc)

distribution/src/main/release/samples/jax_rs/tracing_opentelemetry/pom.xml
```diff
@@ -0,0 +1,162 @@
...
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.cxf</groupId>
+            <artifactId>cxf-rt-transports-http</artifactId>
+            <version>4.0.3-SNAPSHOT</version>
```
reta, line 86: Could you please replace the exact version with `cxf.version`  in this pom.xml?
```suggestion
            <version>${cxf.version}</version>
```

distribution/src/main/release/samples/jax_rs/tracing_opentelemetry/pom.xml
```diff
@@ -0,0 +1,162 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-sdk</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-semconv</artifactId>
+        </dependency>
+        <dependency>
```
reta, line 157: Just to clarify, we need this shim because Feign does not support OTel (but only opentracing)?

distribution/src/main/release/samples/jax_rs/tracing_opentelemetry/src/main/java/demo/jaxrs/tracing/ConsoleSpanExporter.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.util.Collection;
+
+import io.opentelemetry.sdk.common.CompletableResultCode;
+import io.opentelemetry.sdk.trace.data.SpanData;
+import io.opentelemetry.sdk.trace.export.SpanExporter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ConsoleSpanExporter implements SpanExporter {
```
reta, line 11: May be we could use `LoggingSpanExporter` here?

integration/tracing/tracing-opentelemetry/src/main/java/org/apache/cxf/tracing/opentelemetry/AbstractOpenTelemetryClientProvider.java
```diff
@@ -0,0 +1,129 @@
...
+        Context parentContext = Context.current();
+        Span activeSpan = tracer.spanBuilder(buildSpanDescription(uri.toString(), method))
+            .setParent(parentContext).setSpanKind(SpanKind.CLIENT)
+            .setAttribute(SemanticAttributes.HTTP_METHOD, method)
+            .setAttribute(SemanticAttributes.HTTP_URL, uri.toString())
+            // TODO: Enhance with semantics from request
+            .startSpan();
+        Scope scope = activeSpan.makeCurrent();
+
+        GlobalOpenTelemetry.getPropagators().getTextMapPropagator().inject(Context.current(), requestHeaders,
```
reta, line 61: The reference to statics (`GlobalOpenTelemetry`) should not be used, but it seems like in case of `OpenTelemetry`, having the instance of the `Tracer` is not sufficient. Should the `OpenTelemetry` + `instrumentationScopeName` be used instead?

integration/tracing/tracing-opentelemetry/src/main/java/org/apache/cxf/tracing/opentelemetry/OpenTelemetryFeature.java
```diff
@@ -0,0 +1,65 @@
...
+    public OpenTelemetryFeature(final Tracer tracer) {
+        super(new Portable(tracer));
+    }
+
+    public static class Portable implements AbstractPortableFeature {
+        private OpenTelemetryStartInterceptor in;
+        private OpenTelemetryStopInterceptor out;
+
+        public Portable() {
+            this(GlobalOpenTelemetry.getTracer("CXFServer"));
```
reta, line 48: For "CXFServer", we probably need to extract the constant like DEFAULT_TRACER_NAME

parent/pom.xml
```diff
@@ -1870,6 +1870,18 @@
                 <version>${cxf.opentracing.version}</version>
                 <scope>test</scope>
             </dependency>
+            <dependency>
+                <groupId>io.opentelemetry</groupId>
+                <artifactId>opentelemetry-bom-alpha</artifactId>
```
reta, line 1875: Why do we need opentelemetry-bom-alpha and not opentelemetry-bom?

distribution/src/main/release/samples/jax_rs/tracing_opentelemetry/src/main/java/demo/jaxrs/tracing/ConsoleSpanExporter.java
mederel: I did and it works, great suggestion!

integration/tracing/tracing-opentelemetry/src/main/java/org/apache/cxf/tracing/opentelemetry/AbstractOpenTelemetryClientProvider.java
mederel: In the end, an OpenTelemetry object is passed around, and the tracer is retrieved only where it is used using a utility class.

parent/pom.xml
mederel: The only reason but not the least is to have the version of opentelemetry-semconv handled by this BOM. This module has static variables with the standard attribute names from the OpenTelemetry semantics conventions. They are however not "stable" yet, hence the alpha version.

integration/tracing/tracing-opentelemetry/src/main/java/org/apache/cxf/tracing/opentelemetry/OpenTelemetryTracerUtils.java
```diff
@@ -0,0 +1,17 @@
+package org.apache.cxf.tracing.opentelemetry;
+
+import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.api.trace.Tracer;
+import org.apache.cxf.version.Version;
+
+public final class OpenTelemetryTracerUtils {
```
reta, line 7: I would inclined to drop this class and use constant for default name but also provide the way to specify the tracer name (fe `BraveFeature` follows this approach). Please let me know what do you think, I am happy to take that as post merge action, thank you.

parent/pom.xml
reta: If this is just one single artifact, it would be better to have `opentelemetry-semconv` defined separately (like `opentelemetry-opentracing-shim`), it is a bit risky to make a releases with alpha artifacts (that are not test related).

integration/tracing/tracing-opentelemetry/src/main/java/org/apache/cxf/tracing/opentelemetry/OpenTelemetryFeature.java
reta: Thank you, shared my thoughts here https://github.com/apache/cxf/pull/1340#discussion_r1276905543

reta (approved):

parent/pom.xml
mederel: i switch for opentelemetry-bom with semconv on the side, and introduced maven properties for versioning

integration/tracing/tracing-opentelemetry/src/main/java/org/apache/cxf/tracing/opentelemetry/OpenTelemetryTracerUtils.java
mederel: I removed the utility class, and added new constructors with an instrumentationName argument. It defaults to a constant "org.apache.cxf". I remove all notions of instrumentation versions (looking in opentelemetry-java-instrumentation, I see that it is not always set). Custom names are tested in the systests.

systests/tracing/src/test/java/org/apache/cxf/systest/jaxrs/tracing/opentelemetry/HasAttribute.java
reta, line 1: Ah apologies @mederel , the license header is missing, could you please add one?

mederel: Hi, header added on HasAttribute, and also OpenTelemetrySetup from the samples.

reta:
Thanks a lot @mederel !

PR #1344, opened 2023-07-18
[CXF-8905]:TransportURIResolver doesn't honor the property setting in…

jimma:
… bus


reta:
LGTM @jimma ! (apologies, was on vacation last week)

PR #1346, opened 2023-07-19
CXF-8817: Micrometer Observation

marcingrzejszczak:
This PR adds support for Micrometer Observation instrumentation of Apache CXF.

NOTE: This PR is heavily influenced by Brave's implementation.

Dependencies
- Adds micrometer tracing
- Adds zipkin sender

CXF metrics project
- Now also does tracing (TODO: maybe this should be done in a separate module?)
- Adds dependency on `rt-management` since metrics will now also be related to tracing 
- Adds dependency to Micrometer Observation
- Is a copy of the Brave Tracing module but adopted to Micrometer Observation API
- Uses low and high cardinality tags basing on OpenTelemetry Semantic Conventions for RPC and HTTP (JaxRs) from 19.07.2023. VERY IMPORTANT: This doesn't mean that this project is OTel Semantic Convention compliant!!!
- Uses `ObservationDocumentation` that using https://micrometer.io/docs/observation#_documentation_generation could generate asciidoctor documentation of observability data

Spring Boot
- adds `MicrometerObservationAutoConfiguration` that registers `Feature`s as beans

Tests
- reused existing Brave tests with Micrometer Observation, altered existing tagging and naming defaults to make the test pass with Brave tagging and naming defaults
- reused existing Boot tests to additionally assert observation based metrics and traces

TODO ? :
- maybe the code shouldn't be in CXF metrics module but a completely separate one?
- add Javadocs?
- add documentation - where should it be added?

Screenshots from Zipkin:

Screenshot from 2023-07-19 13-25-48

Screenshot from 2023-07-19 13-25-55

Fixes https://issues.apache.org/jira/browse/CXF-8817


reta:
Thanks a lot for the pull request @marcingrzejszczak ! 

> Now also does tracing (TODO: maybe this should be done in a separate module?)

This is the way CXF structures the integrations: the tracing is separated from metrics (integration/tracing). Could you please split them into separate modules? Thank you.

marcingrzejszczak:
@reta I've moved the Micrometer Observation integration under the tracing folder. The integration allows you to actually not have tracing at all and in essence is tracing agnostic, however for the sake of consistency I put it under the tracing folder

reta:
> @reta I've moved the Micrometer Observation integration under the tracing folder. The integration allows you to actually not have tracing at all and in essence is tracing agnostic, however for the sake of consistency I put it under the tracing folder

Thanks a lot @marcingrzejszczak , there are some failures in the build, could you please take a look? thank you

```
Caused by: java.lang.IllegalStateException: Failed to introspect Class [org.apache.cxf.spring.boot.autoconfigure.micrometer.MicrometerObservationAutoConfiguration] from ClassLoader [jdk.internal.loader.ClassLoaders$AppClassLoader@5cb0d902]
	at org.springframework.util.ReflectionUtils.getDeclaredMethods(ReflectionUtils.java:483)
	at org.springframework.util.ReflectionUtils.doWithMethods(ReflectionUtils.java:360)
```

integration/tracing/tracing-micrometer/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+            <artifactId>cxf-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.cxf</groupId>
+            <artifactId>cxf-rt-management</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
```
reta, line 40: I think these two artifacts shouldn't be optional since they are always required for this integration, right?

systests/spring-boot/src/test/java/org/apache/cxf/systest/jaxrs/spring/boot/SpringJaxrsComponentScanningTest.java
```diff
@@ -51,8 +52,8 @@ public void testCxfComponentScan() {
         // The component scanner only looks for CXF's @Provider annotations, 
         // not JAX-RS Features/@Provider.
         assertThat(scanner.getFeatures())
-                .hasSize(2)
-                .hasOnlyElementsOfTypes(OpenApiFeature.class, JAXRSBeanValidationFeature.class);
+                .hasSize(3)
+                .hasOnlyElementsOfTypes(ObservationFeature.class, OpenApiFeature.class, OpenApiFeature.class, JAXRSBeanValidationFeature.class);
```
reta, line 56: Seems like `OpenApiFeature.class` is included twice:

```suggestion
                .hasOnlyElementsOfTypes(ObservationFeature.class, OpenApiFeature.class, JAXRSBeanValidationFeature.class);
```

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/MicrometerObservationAutoConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.context.annotation.Configuration;
+
+import io.micrometer.observation.ObservationRegistry;
+
+@Configuration
+@AutoConfigureAfter(ObservationAutoConfiguration.class)
+@ConditionalOnClass({ ObservationRegistry.class, ObservationFeature.class })
+@ConditionalOnProperty(name = "cxf.observation.enabled", matchIfMissing = true)
+@ConditionalOnBean(ObservationRegistry.class)
+public class MicrometerObservationAutoConfiguration {
```
reta, line 43: I am unsure about this particular auto configuration, it conflicts with the way CXF uses scans (coming from CxfAutoConfiguration).

marcingrzejszczak: The `MessageInObservationConvention` and `MessageOutObservationConvention` *MAY BE* provided by the user. This is to override the defaults that CXF would provide. It is perfectly valid *NOT* to have these defined by the user

marcingrzejszczak:
Hey @reta . I've applied all the suggestions and rebased the branch against `main`. Anything else you would like me to do?

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/MicrometerObservationAutoConfiguration.java
marcingrzejszczak: So `CxfAutoConfiguration` will create a bean for the `ObservationFeature` ? I can't find the code that does that.

reta: Yes, there are 2 way it works (managed by property):
 -  scanning classpath and extracting JAX-RS annotated classes (done by https://github.com/apache/cxf/blob/main/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/spring/AbstractJaxrsClassesScanServer.java)
 - scanning Spring components (done by https://github.com/apache/cxf/blob/main/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/spring/AbstractSpringComponentScanServer.java)

reta: Correct, that is why I was suggesting to have pre-created all needed beans (guarded by @ConditionalOnMissingBean, `ObservationRegistry` falls into this category) in this auto configuration (if there are chances in may not be present. But you think it is redundant, the configuration could be removed.

marcingrzejszczak: I tried removing the autoconfiguration but from what I see the `AbstractSpringComponentScanServer` is trying to call a default constructor that the `ObservationProvider` or `ObservationClientProvider` just don't have. I wonder if it makes sense to have providers that have a non default constructor :shrug:

marcingrzejszczak: I will modify the component scanning logic to allow trying to autowire by constructor. That way if you have a provider that has constructors Spring will try to autowire using it

marcingrzejszczak: OK I gave it another thought and I don't see any difference between `MicrometerMetricsProvider` which is defined as a Spring bean and `ObservationProvider` or `ObservationClientProvider`. Trying to change the existing code to allow construction injection is not enough for all test to pass. I would leave the current auto-configuration cause that prepares the providers the way they should be set up given that the ObservationRegistry is on the classpath. From my perspective this PR is ready.

reta: >  Trying to change the existing code to allow construction injection is not enough for all test to pass.

Thanks @marcingrzejszczak , do you mind if I push this specific change? Thank you.

marcingrzejszczak: Oh be my guest :) 

What I did is I modified the `JAXRSServerFactoryBeanDefinitionParser` to have its last static method to fallback to constructor autowiring.

```java
static List<Object> createBeansFromDiscoveredClasses(ApplicationContext context,
                                                         Collection<Class<?>> classes,
                                                         Class<? extends Annotation> serviceClassAnnotation) {
        AutowireCapableBeanFactory beanFactory = context.getAutowireCapableBeanFactory();
        final List< Object > providers = new ArrayList<>();
        for (final Class< ? > clazz: classes) {
            if (serviceClassAnnotation != null && clazz.getAnnotation(serviceClassAnnotation) == null) {
                continue;
            }
            Object bean;
            try {
                bean = beanFactory.createBean(clazz, AutowireCapableBeanFactory.AUTOWIRE_BY_TYPE, true);
            } catch (Exception ex) {
                String stackTrace = ExceptionUtils.getStackTrace(ex);
                LOG.fine("Autowire failure for a " + clazz.getName() + " bean: " + stackTrace);
                try {
                    bean = beanFactory.createBean(clazz);
                } catch (Exception ex2) {
                    // Fallback to constructor based injection
                    String stackTrace2 = ExceptionUtils.getStackTrace(ex2);
                    LOG.fine("Bean creation failure for a " + clazz.getName() + " bean: " + stackTrace2);
                    GenericBeanDefinition definition = new GenericBeanDefinition();
                    definition.setBeanClass(clazz);
                    definition.setAutowireMode(AbstractBeanDefinition.AUTOWIRE_CONSTRUCTOR);
                    String name = clazz.getName() + "_" + System.currentTimeMillis();
                    ((BeanDefinitionRegistry) context).registerBeanDefinition(name, definition);
                    bean = context.getBean(name);
                }
            }
            providers.add(bean);
        }
        return providers;
    }
```

reta: @marcingrzejszczak pushed my changes
  - the `MicrometerObservationAutoConfiguration` is removed (the features are created manually now but I will take care of it)
  - refactored a bit test cases for systests/tracing (to separate Brave and Micrometer)
  - added more test cases for Spring Boot (and here was confused, asking for your help)

Hope you have no objections, thank you.

systests/spring-boot/src/test/java/org/apache/cxf/systest/jaxrs/spring/boot/SpringJaxrsObservabiityTest.java
```diff
@@ -0,0 +1,173 @@
...
+               .hasASpanWithName("GET", spanAssert -> {
+                   spanAssert.hasKindEqualTo(Kind.CLIENT)
+                             .hasTag("rpc.service", "WebClient")
+                             .hasTag("rpc.system", "cxf")
+                             .hasTagWithKey("server.address")
+                             .hasTagWithKey("server.port")
+                             .isStarted()
+                             .isEnded();
+               })
+               .hasASpanWithName("rpc.server.duration", spanAssert -> {
```
reta, line 145: The server span name is very unexpected `rpc.server.duration`, do you know why is that?

systests/spring-boot/src/test/java/org/apache/cxf/systest/jaxrs/spring/boot/SpringJaxrsObservabiityTest.java
```diff
@@ -0,0 +1,173 @@
...
+    @Autowired
+    private ObservationRegistry observationRegistry;
+
+    @Autowired
+    private ArrayListSpanReporter arrayListSpanReporter;
+
+    @LocalServerPort
+    private int port;
+
+    @EnableAutoConfiguration(exclude = WebMvcObservationAutoConfiguration.class)
```
reta, line 80: For JAX-RS (aka WebMvc :D), the most common scenario would be to have HTTP traces but I have not found the way to trade `rpc` for `http`, without `WebMvcObservationAutoConfiguration` I do have both on server side, could you help me please to understand the proper configuration?

marcingrzejszczak: You did it right - I think I did the same thing. So `WebMvcObservationAutoConfiguration` is the configuration that hooks in for MVC observations on the Servlet level. The Cxf instrumentation is on a different level and should take precedense for that matter. You would have the same problem with e.g. Tomcat Valve instrumentation. Tomcat would hook in earlier and we would have to disable the `WebMvcAutoConfiguration` (or just the observation filter) so as not to do double instrumentation.

systests/spring-boot/src/test/java/org/apache/cxf/systest/jaxrs/spring/boot/SpringJaxrsObservabiityTest.java
```diff
@@ -0,0 +1,173 @@
...
+    @ComponentScan(basePackageClasses = Library.class)
+    static class TestConfig {
+        @Bean
+        public Feature observationFeature(ObservationRegistry observationRegistry) {
```
marcingrzejszczak, line 84: Here you are creating the feature manually and I guess we want that to happen out of the box?

systests/spring-boot/src/test/java/org/apache/cxf/systest/jaxrs/spring/boot/SpringJaxrsObservabiityTest.java
marcingrzejszczak: Interesting. That would mean that the default, "metric" related name hooked in and the contextual name wasn't called. Here you can see how the contextual name should be set (https://github.com/apache/cxf/pull/1346/files#diff-05b29f9cca6ab07f827c4985ccf14e61f6c878510b4e18c37634b360041ed658R42) and here (https://github.com/apache/cxf/pull/1346/files#diff-3a451c2c9618eb5753ffd9960d384ffa6685f97fedf6605d787c7c018556d871R77) but maybe I messed sth up 😬

systests/spring-boot/src/test/java/org/apache/cxf/systest/jaxrs/spring/boot/SpringJaxrsObservabiityTest.java
marcingrzejszczak: > the MicrometerObservationAutoConfiguration is removed (the features are created manually now but I will take care of it)

I haven't noticed this messed from you so disregard the comment

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/micrometer/MicrometerObservationAutoConfiguration.java
marcingrzejszczak: > added more test cases for Spring Boot (and here was confused, asking for your help)

I've replied to the comment 

> Hope you have no objections, thank you.

Oh absolutely not, I'm just the guest here :) Do all the changes that you think make sense!

systests/spring-boot/src/test/java/org/apache/cxf/systest/jaxrs/spring/boot/SpringJaxrsObservabiityTest.java
reta: I will try to debug that, but besides that - I think we should be all set

reta: Should be all set, the JAX-RS HTTP instrumentations works pretty well.

reta (approved):

marcingrzejszczak:
LGTM!

PR #1363, opened 2023-07-30
CXF-8860: Get rid of EasyMock in cxf-rt-ws-rm

reta:
Get rid of EasyMock in cxf-rt-ws-rm


mederel:
I discovered the keyword mock with the mockito library, so for me, these changes make sense of course. My side question is: what happened to easymock to motivate the switch?

Either way, this change is welcome I think.

PR #1377, opened 2023-08-10
[CXF-8885] Eliminate hidden reference to `HttpClient` in `ProxySelector`

LeoWoerteler:
The SelectorManager thread of the WebClient would normally shut down after the HttpClient's outer shell, the `HttpClientFacade` has been garbage collected. Unfortunately the `ProxySelector` instance created in `HttpClientHTTPConduit` is a non-static (anonymous) subclass, so it retains a hard reference to the enclosing `HttpClientHTTPConduit` instance and therefore also its `client` field. This gives the SelectorManager thread a hard reference to the `HttpClientFacade`, so the facade can never be garbage collected and the thread never shuts down. Making the class static solves the problem.


reta (approved):

reta:
Thank you @LeoWoerteler !

cachescrubber:
Thank you @LeoWoerteler, I just verified the fix in my reproducible example:

https://github.com/cachescrubber/cxf-8885-demo/commit/fda7c8a5248689357f76effe200b34f38fa83126

PR #1380, opened 2023-08-12
CXF-8885: Eliminate references to `HttpClient` via `HTTPClientPolicy`

LeoWoerteler:
There are additional potential references to the `HttpClientFacade` hidden in the property change listeners of `HTTPClientPolicy`. To prevent those from keeping the SelectorManager thread alive, we have store a copy of the policy with no listeners attached.

### DIFF
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HttpClientHTTPConduit.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HttpClientHTTPConduit.java
@@ -303,19 +303,34 @@ public class HttpClientHTTPConduit extends URLConnectionHTTPConduit {
      * This class <i>must</i> be static so it doesn't capture a reference to {@code HttpClientHTTPConduit.this} and
      * through that to {@link HttpClientHTTPConduit#client}. Otherwise the client can never be garbage collected, which
      * means that the companion "SelectorManager" thread keeps running indefinitely (see CXF-8885).
+     * <p>
+     * The {@link HTTPClientPolicy} must be copied here because its property change listeners can still have hard
+     * references to the {@code HttpClientHTTPConduit}.
      */
     private static final class ProxyFactoryProxySelector extends ProxySelector {
         private final ProxyFactory proxyFactory;
-        private final HTTPClientPolicy csPolicy;
+        private final HTTPClientPolicy clientPolicy;
 
         ProxyFactoryProxySelector(ProxyFactory proxyFactory, HTTPClientPolicy csPolicy) {
             this.proxyFactory = proxyFactory;
-            this.csPolicy = csPolicy;
+            this.clientPolicy = new HTTPClientPolicy();
+            if (csPolicy.isSetProxyServerType()) {
+                clientPolicy.setProxyServerType(csPolicy.getProxyServerType());
+            }
+            if (csPolicy.isSetProxyServer()) {
+                clientPolicy.setProxyServer(csPolicy.getProxyServer());
+            }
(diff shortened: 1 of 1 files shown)


rt/transports/http/src/main/java/org/apache/cxf/transport/http/HttpClientHTTPConduit.java
```diff
@@ -303,19 +303,34 @@ protected OutputStream createOutputStream(Message message, boolean needToCacheRe
...
      */
     private static final class ProxyFactoryProxySelector extends ProxySelector {
         private final ProxyFactory proxyFactory;
-        private final HTTPClientPolicy csPolicy;
+        private final HTTPClientPolicy clientPolicy;
 
         ProxyFactoryProxySelector(ProxyFactory proxyFactory, HTTPClientPolicy csPolicy) {
             this.proxyFactory = proxyFactory;
-            this.csPolicy = csPolicy;
+            this.clientPolicy = new HTTPClientPolicy();
```
LeoWoerteler, line 316: If there is a better, more elegant way to copy the `HTTPClientPolicy`'s contents without the listeners, please let me know.

reta: My first reaction - the HTTP client policy outlives the any client facades so the no need to copy, do you happen to have heap roots to confirm the potential hypothesis? Thank you.

LeoWoerteler: Do you want me to send you the heap dump? I can certainly do that: [heapdump-1691833659800.hprof.gz](https://github.com/apache/cxf/files/12327502/heapdump-1691833659800.hprof.gz) I created it from my repro here: https://gist.github.com/LeoWoerteler/ecf7cc7d975990615a0a3b46152fa0b4

Just keep it running for a minute or so and then check for the GC roots of `HttpClientFacade`. This is how it looks for me:
image

LeoWoerteler: By the way, I'm also experimenting with registering the `HTTPConduit` as listener via a `SoftReference<PropertyChangeListener>` to break the reference chain, but that doesn't fix the problem in my experiments: https://github.com/apache/cxf/commit/dd7131416971723868d95c7daf0153125937dfc4

It's really weird actually, the `HttpClientFacade`s are not being garbage collected even though no GC roots are being found.

reta: Thank you @LeoWoerteler , I think it is sufficient to see the problem, will try to look shortly

reta: > It's really weird actually, the `HttpClientFacade`s are not being garbage collected even though no GC roots are being found.

Hm ... that's weird I agree

LeoWoerteler: @reta I've figured it out, it turns out I mixed up `SoftReference` and `WeakReference`. Softly referenced objects are only garbage collected as a last resort before OOM, while weak references objects are ignored completely by the GC. I've opened PR #1381 as an alternative to this PR, feel free to reject the one you like less of the two.

reta:
Closing in favor of https://github.com/apache/cxf/pull/1381

PR #1381, opened 2023-08-12
CXF-8885: Eliminate hard refs to `HttpClient` via `HTTPClientPolicy`

LeoWoerteler:
There are additional potential references to the `HttpClientFacade` hidden in the property change listeners of `HTTPClientPolicy`. To prevent those from keeping the SelectorManager thread alive, the `HTTPConduit` now registers itself as a listener only via a `WeakReference`, so the conduit can be garbage collected even if the client policy is still referenced inside the SelectorManager thread.

This is an alternative to PR #1380, the `WeakReference` approach taken here allows the `HTTPClientPolicy` to change dynamically.

### DIFF
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
@@ -26,6 +26,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import java.lang.ref.WeakReference;
 import java.net.HttpRetryException;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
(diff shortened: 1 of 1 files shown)


reta:
Thank @LeoWoerteler, I am still unsure it addresses the problem: the client should be closed explicitly and when it happens, all property listeners it registered are removed (at least, that's the idea). @dkulp could you please chime in since you've been working in the HttpClient conduit implementation, thank you.

LeoWoerteler:
@reta @dkulp I guess that is the core of the question then. In CXF 3.5 it was possible to close only the `Response` but not the `WebClient` without severe consequences. After we had to upgrade to CXF 4.0.X because of the `javax.*` to `jakarta.*` move, now every `WebClient` that isn't being closed leaks a zombie thread, which eventually leads to OOM crashes. Was this always a misuse of the API (even though `WebClient` isn't even `AutoCloseable` and many examples/answers online say otherwise) and therefore the breakage is fine or is this a serious regression?

In my opinion the resource leak is easily fixable (for example via the `WeakReference`), so leaving it in would be very unfortunate. Missing a single unclosed `WebClient` or proxy anywhere in the code base during the CXF upgrade would introduce a memory leak. That makes the upgrade extremely risky.

reta:
> @reta @dkulp I guess that is the core of the question then. In CXF 3.5 it was possible to close only the `Response` but not the `WebClient` without severe consequences

Could you elaborate more on that please? `WebClient::close` could be called when client is not used anymore. It is indeed not `AutoCloseable` (I think the initial version predates JDK 1.7 when this interface was introduced) but it should I think.

LeoWoerteler:
> Could you elaborate more on that please? `WebClient::close` could be called when client is not used anymore. It is indeed not `AutoCloseable` (I think the initial version predates JDK 1.7 when this interface was introduced) but it should I think.

It could be called, it it practically didn't have to be, so lots of existing code doesn't. Until CXF 3.5 it was evidently enough to close the `Response` to avoid leaking resources, and there are plenty of sources online that claim that. Now in 4.0 the same pattern is suddenly leaky. That is at the very least a giant foot-gun for anyone upgrading CXF on older code bases.

reta:
It is very valuable feedback, thank you @LeoWoerteler , I hope @dkulp gets engaged

PR #1399, opened 2023-08-23
CXF-8919 - JAXRSUtils references javax.activation.Datasource

rzo1:
# What does this PR do?

JAXRSUtils has a LazyClass reference to `javax.activation.Datasource`. Since we are using jakarta namespace for cxf4, we need to fix it.

The wrong reference is a problem for

`com.sun.ts.tests.jaxrs.spec.provider.standardnotnull.JAXRSClient#java#clientDataSourceProviderTest_from_standalone`

if running the ee9.1 plattform tck for TomEE.


reta:
Thank you @rzo1 !

PR #1408, opened 2023-08-30
CXF-8920: use the same approach with Temporal.class.isAssignableFrom …

vershnik:
…check as used for non-collection fields to recognize and convert temporal values


reta (approved):
Thank you @vershnik !

PR #1454, opened 2023-09-26
CXF-8936 Fix h2 protocol negotiation in Jetty Transport

huw0:
# Summary 

I have been trying to update one of our apps using the Jetty backend to use http2. While doing this I've noticed that the protocol negotiation is in the wrong order which results in http/1.1 always being preferred where the client includes this in the ALPN handshake. This means that both browsers and cURL will never successfully negotiate h2/http2.

https://issues.apache.org/jira/browse/CXF-8936

[output-jetty-original.txt](https://github.com/apache/cxf/files/12729574/output-jetty-original.txt)
[output-netty.txt](https://github.com/apache/cxf/files/12729575/output-netty.txt)

### DIFF
--- a/rt/transports/http-jetty/src/main/java/org/apache/cxf/transport/http_jetty/JettyHTTPServerEngine.java
+++ b/rt/transports/http-jetty/src/main/java/org/apache/cxf/transport/http_jetty/JettyHTTPServerEngine.java
@@ -684,7 +684,6 @@ public class JettyHTTPServerEngine implements ServerEngine, HttpServerEngineSupp
             result = new org.eclipse.jetty.server.ServerConnector(server);
 
             if (tlsServerParameters != null) {
-                connectionFactories.add(httpFactory);
                 httpConfig.addCustomizer(new SecureRequestCustomizer(tlsServerParameters.isSniHostCheck()));
 
                 if (isHttp2Enabled(bus)) {
@@ -704,10 +703,14 @@ public class JettyHTTPServerEngine implements ServerEngine, HttpServerEngineSupp
                         }
                     }
                 }
-                if (connectionFactories.size() == 1) {
+                
+                if (connectionFactories.isEmpty()) {
                     final SslConnectionFactory scf = new SslConnectionFactory(sslcf, httpFactory.getProtocol());
                     connectionFactories.add(scf);
                 }
+                
+                // Ensure http/1.1 is last in the list so it is the lowest priority in ALPN
+                connectionFactories.add(httpFactory);
 
                 // Has to be set before the default protocol change
                 result.setConnectionFactories(connectionFactories);


rt/transports/http-jetty/src/main/java/org/apache/cxf/transport/http_jetty/JettyHTTPServerEngine.java
```diff
@@ -697,7 +697,8 @@ AbstractConnector createConnectorJetty(SslContextFactory.Server sslcf, String ho
                         final SslConnectionFactory scf = new SslConnectionFactory(sslcf, alpn.getProtocol());
                         connectionFactories.add(scf);
```
reta, line 698: Thanks for the fix, @huw0 , I believe the ALPN should go before HTTP/2, wdyt?

```
connectionFactories.add(alpn);
connectionFactories.add(new HTTP2ServerConnectionFactory(httpConfig));
connectionFactories.add(scf);
```

huw0: As far as I can tell, the ordering of the ALPN connection factory doesn't matter. From what I can see within Jetty's `AbstractConnector` the ordering of the connection factories doesn't matter at the point where ALPN is started.

```
        if (ssl != null)
        {
            String next = ssl.getNextProtocol();
            ConnectionFactory cf = getConnectionFactory(next);
            if (cf == null)
                throw new IllegalStateException("No protocol factory for SSL next protocol: '" + next + "' in " + this);
        }
```

_In this case, `ssl.getNextProtocol()` will return ALPN_

However I am not that familiar with Jetty's internals and whether there is anywhere else where this would matter.

The key thing is just that http2 precedes http/1.1 in the connection factories list. There are four connection factories in total. The first of these is on line 691 and is outside of the `isHttp2Enabled` block.

```
            if (tlsServerParameters != null) {
                connectionFactories.add(httpFactory);
                httpConfig.addCustomizer(new SecureRequestCustomizer(tlsServerParameters.isSniHostCheck()));
                if (isHttp2Enabled(bus)) {
                ....
```

I would be happy to refactor so that the first connector (which is http/1.1) is added after the `isHttp2Enabled` block, which would remove the need to explicitly add http2 as the first item in the list.

reta: > I would be happy to refactor so that the first connector (which is http/1.1) is added after the isHttp2Enabled block, which would remove the need to explicitly add http2 as the first item in the list.

I think that would make sense, thank you. It also is aligned with Jetty examples [1]

[1] https://github.com/eclipse/jetty.project/blob/jetty-9.4.14.v20181114/examples/embedded/src/main/java/org/eclipse/jetty/embedded/Http2Server.java

huw0: I've made these changes. When running the CXF Jetty demo, the ordering now matches that of the Jetty example.

`INFO: Started ServerConnector@11eadcba{ssl, (ssl, alpn, h2, http/1.1)}{0.0.0.0:9000}`

reta (approved):

PR #1472, opened 2023-10-14
[JAX-WS] The client out interceptor chain is not called in case of oneway operation

reta:
The client out interceptor chain is not called in case of oneway operation

### DIFF
--- a/core/src/main/java/org/apache/cxf/endpoint/ClientImpl.java
+++ b/core/src/main/java/org/apache/cxf/endpoint/ClientImpl.java
@@ -36,6 +36,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executor;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 import javax.xml.namespace.QName;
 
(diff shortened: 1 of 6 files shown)


reta:
@coheigea may I ask you to take a second look if what I am trying to do here is making sense (the change is non-breaking), thank you

coheigea:
I'll defer to @dkulp on this one @reta

dkulp:
Why would the out interceptor chain be called when there isn't a message?   That seems wrong.   This just "fakes" a message which doesn't seem right.    My gut feeling is that this would also break the JSX-WS TCK if this ends up calling the JAX-WS handlers.   

OpenTelemetry likely should have an interceptor at the end of the in chain which checks the operation and if it's a one way, close whatever it needs to close at that point.

reta:
> Why would the out interceptor chain be called when there isn't a message? That seems wrong. This just "fakes" a message which doesn't seem right. My gut feeling is that this would also break the JSX-WS TCK if this ends up calling the JAX-WS handlers.

Thanks @dkulp! There is `out` message, always, but not `in` message. And we call only `OneWayInterceptor` in this case, as I mentioned the change is non-breaking - none of the existing interceptors in the chain will be called unless opted in (by implementing `OneWayInterceptor`).

> OpenTelemetry likely should have an interceptor at the end of the in chain which checks the operation and if it's a one way, close whatever it needs to close at that point.

It could be done as well, the issue is that it someone has to know that
a) message is one way, it should be checked
b) the `in` interceptors won't be called 

This is very easy to forget and catch, with  `OneWayInterceptor` at least the handling could be formalized and generalized.

reta:
> Sorry. I missed the filterOneway calls.

Not a problem, thanks a lot @dkulp ! 

>  Not sure if it would make sense to add a "bool empty()" to InterceptorChain and skip setting the thread bus and everything for the common case.

I will try to prototype that, my initial attempt was to have a similar hook but it exploded the changeset since the interceptor chain had to be modified to to support this additional type of interceptor call.

PR #1473, opened 2023-10-14
[CXF-8940]ensure we have PolicyBasedWSS4JInterceptors for UsernameTok…

ffang:
…enInterceptorProvider


systests/ws-security/src/test/resources/org/apache/cxf/systest/ws/ut/plaintext-pass-policy.xml
```diff
@@ -2,27 +2,6 @@
 <wsp:Policy xmlns:wsu="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-utility-1.0.xsd" xmlns:wsp="http://www.w3.org/ns/ws-policy" wsu:Id="TransportUsernameTokenPolicy">
     <wsp:ExactlyOne>
         <wsp:All>
```
coheigea, line 4: Couldn't we add a new policy rather that modifying an existing testcase?

coheigea (changes requested):

systests/ws-security/src/test/resources/org/apache/cxf/systest/ws/ut/plaintext-pass-policy.xml
ffang: Hi @coheigea,

Revised testcase accordingly.

Cheers
Freeman

coheigea (commented):
@ffang Why do we want the policy based interceptors added and what relation does it have to https://issues.apache.org/jira/browse/CXF-8940 ?

ffang:
> @ffang Why do we want the policy based interceptors added and what relation does it have to https://issues.apache.org/jira/browse/CXF-8940 ?

Hi @coheigea ,

CXF-8940 complains the issue that property "ws-security.must-understand" can't be honoured when Dom based ws-secuirty API is used, while it's working in Stax based ws-secuirty. 

After debugging, I realized that this issue only occurs when UsernameToken is used alone but no Security Binding(like HTTPS TransportBinding) in policy file. Actually in PolicyBasedWSS4JOutInterceptor.java we have code like
```
              boolean mustUnderstand =
                MessageUtils.getContextualBoolean(
                    message, SecurityConstants.MUST_UNDERSTAND, true
                );
            String actor = (String)message.getContextualProperty(SecurityConstants.ACTOR);

            // extract Assertion information
            AbstractBinding binding = PolicyUtils.getSecurityBinding(aim);

            if (binding == null && isRequestor(message)) {
                Policy policy = new Policy();
                binding = new TransportBinding(org.apache.wss4j.policy.SPConstants.SPVersion.SP11,
                                                 policy);
            }

            if (binding != null) {
                WSSecHeader secHeader = new WSSecHeader(actor, mustUnderstand, saaj.getSOAPPart());
                final Element el;
                try {
                    el = secHeader.insertSecurityHeader();
                } catch (WSSecurityException e) {
                    throw new SoapFault(
                        new Message("SECURITY_FAILED", LOG), e, message.getVersion().getSender()
                    );
                }
```
to honour property "ws-security.must-understand" when using dom based ws-security API, however, this interceptor isn't added OOTB if the policy file only contains UsernameToken but no Security Binding. 

Take a close look at UsernameTokenInterceptorProvider.java, the PolicyBasedWSS4JStaxOutInterceptor and PolicyBasedWSS4JStaxInInterceptor have been there already, that's why Stax based ws-secuirty can honour property "ws-security.must-understand" when the policy file only contains UsernameToken. And I believe PolicyBasedWSS4JOutInterceptor and PolicyBasedWSS4JInInterceptor should be added there to support DOM based ws-security.

Also, I believe we have a similar issue https://issues.apache.org/jira/browse/CXF-2890,
 and the fix for it is
https://github.com/apache/cxf/commit/48b7dce3d977a1630cefb633fe88480ee0746b8f

Freeman

coheigea:
@ffang The problem is that adding the DOM interceptors means that you are using the whole benefit of the UsernameTokenInterceptor, which is a pure-streaming interceptor. It will cause the message to be converted to DOM. Instead IMO the fix is to add the must understand header as part of the streaming code in UsernameTokenInterceptor.

coheigea:
@ffang The "SecurityConstants.ENABLE_STREAMING_SECURITY" refers to the streaming WS-Security implementation in WSS4J. The UsernameTokenInterceptor doesn't use WSS4J at all, it's just a pure-streaming interceptor for when we have the simple case of a UsernameToken with no security binding.

ffang:
> @ffang The "SecurityConstants.ENABLE_STREAMING_SECURITY" refers to the streaming WS-Security implementation in WSS4J. The UsernameTokenInterceptor doesn't use WSS4J at all, it's just a pure-streaming interceptor for when we have the simple case of a UsernameToken with no security binding.

Hi @coheigea !

Thanks for the response!

So you are suggesting that we need to accommodate the property "ws-security.must-understand" in AbstractTokenInterceptor.findSecurityHeader method, right?

Currently we have code in that method like
{code}
        SoapHeader sh = new SoapHeader(new QName(WSS4JConstants.WSSE_NS, "Security"), el);
        sh.setMustUnderstand(true);
{code}
where mustUnderstand is hard-coded as true, but we should check property "ws-security.must-understand"  here and setMustUnderstand accordingly. Is my understanding correct?

Freeman

coheigea:
@ffang Yes exactly, thanks!

ffang:
> @ffang Yes exactly, thanks!

Thanks @coheigea !

Revised accordingly!

coheigea (approved):

PR #1483, opened 2023-10-20
[CXF-8947] - Avoid expensive regex operations in Rfc3986UriValidator

WhiteCat22:
https://issues.apache.org/jira/browse/CXF-8947

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/Rfc3986UriValidator.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/Rfc3986UriValidator.java
@@ -39,7 +39,7 @@ final class Rfc3986UriValidator {
 
     private static final String LAST = "#(.*)";
 
-    private static final Pattern HTTP_URL = Pattern.compile("^" + SCHEME 
+    private static final Pattern HTTP_URL = Pattern.compile("^" + SCHEME
         + "(//(" + USERINFO + "@)?" + HOST  + ")?" + PATH
         + "(\\?" + QUERY + ")?" + "(" + LAST + ")?");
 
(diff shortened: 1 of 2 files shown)


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/Rfc3986UriValidator.java
```diff
@@ -39,21 +39,28 @@ final class Rfc3986UriValidator {
...
      * @param uri HTTP schemed URI to validate
      * @return "true" if URI matches RFC-3986 validation rules, "false" otherwise
      */
     public static boolean validate(final URI uri) {
         // Only validate the HTTP(s) URIs
-        if (HttpUtils.isHttpScheme(uri.getScheme())) { 
+        if (HttpUtils.isHttpScheme(uri.getScheme())) {
+        // If URI.getHost() returns a host name, validate it and
+            // skip the expensive regular expression logic.
+            final String uriHost = uri.getHost();
```
reta, line 59: @WhiteCat22 the reason for this validator to exists sadly is the fact that Java's URI is not RFC-3986 complaint. The `host` is not trustful source here hence we validate it against the pattern.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/HttpUtils.java
```diff
@@ -93,13 +93,18 @@ public final class HttpUtils {
...
     private static final String PATH_RESERVED_CHARACTERS = "=@/:!$&\'(),;~";
     private static final String QUERY_RESERVED_CHARACTERS = "?/,";
-    
+
     private static final Set<String> KNOWN_HTTP_VERBS_WITH_NO_REQUEST_CONTENT =
         new HashSet<>(Arrays.asList(new String[]{"GET", "HEAD", "OPTIONS", "TRACE"}));
     private static final Set<String> KNOWN_HTTP_VERBS_WITH_NO_RESPONSE_CONTENT =
         new HashSet<>(Arrays.asList(new String[]{"HEAD", "OPTIONS"}));
-    
-    private static final Pattern HTTP_SCHEME_PATTERN = Pattern.compile("^(?i)(http|https)$");
```
reta, line 102: This pattern is super straightforward to look up, what is exactly the gains here (vs adding the set + comparator + lowecasing)?

WhiteCat22: Hi @reta , Sorry for the delay. The reason we changed this was because `HashSet.contains(String)` uses "WAY" less CPU than `Pattern.matcher(String).matches()`

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/impl/Rfc3986UriValidator.java
WhiteCat22: That is a good point that we had not considered.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/HttpUtils.java
reta: Fair enough, may be we could just have two constants instead?

gnodet:
Hi @reta, I've pushed a commit addressing your review feedback:

- **Removed** the `URI.getHost()` shortcut in `Rfc3986UriValidator` — you were right that Java's `URI.getHost()` is not RFC-3986 compliant, so it can't be trusted here.
- **Simplified** `HttpUtils.isHttpScheme()` to use `equalsIgnoreCase()` with two string constants instead of a `HashSet`, as you suggested.

PR #1490, opened 2023-10-23
Removing owasp maven plugin as we rely on dependabot now

coheigea:


reta:
@coheigea if you don't mind, I would prefer to keep the plugin for local runs (I do that from time to time), we don't trigger the plugin on any builds by default. Thank you.

coheigea:
OK no problem @reta

PR #1504, opened 2023-11-02
Use Collection.Sort and compare the contents of the list to fix the flaky test

prathyushreddylpr:
### Description
Fixed the flaky test `testGetFiles` inside `FileUtilsTest.java` class. 

https://github.com/apache/cxf/blob/8b205c8086aeb1f5514472f0802c1fb7d510a50e/core/src/test/java/org/apache/cxf/helpers/FileUtilsTest.java#L70

#### Root Cause
The test `testGetFiles` has been reported as flaky when run with the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool. The test failed because it is trying to compare contents in two lists, but since the order is not maintained in lists the assert is failing. The List in Java is implemented in such a way that it does not store the order in which the values are inserted. As a result, when both the lists are compared, it caused the failure. 

#### Fix
This test is fixed by first sorting the two lists and then comparing both lists.

### How this has been tested?

**Maven:** Apache Maven 3.6.3

Command used - 
```
```

Command used - 
```
mvn -pl core test -Dtest=org.apache.cxf.helpers.FileUtilsTest#testGetFiles
```

3) **NonDex test**  - Failed
Command used - 
```
mvn -pl core edu.illinois:nondex-maven-plugin:2.1.7-SNAPSHOT:nondex -Dtest=org.apache.cxf.helpers.FileUtilsTest#testGetFiles
```

### DIFF
--- a/core/src/test/java/org/apache/cxf/helpers/FileUtilsTest.java
+++ b/core/src/test/java/org/apache/cxf/helpers/FileUtilsTest.java
@@ -24,6 +24,7 @@ import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 
@@ -77,6 +78,8 @@ public class FileUtilsTest {
 
         List<File> foundFiles2 = FileUtils.getFiles(directory, ".*\\.class$");
 
+        Collections.sort(foundFiles);
+        Collections.sort(foundFiles2);
         assertEquals(foundFiles, foundFiles2);
     }
 }


core/src/test/java/org/apache/cxf/helpers/FileUtilsTest.java
```diff
@@ -77,6 +78,8 @@ public void testGetFiles() throws URISyntaxException {
 
         List<File> foundFiles2 = FileUtils.getFiles(directory, ".*\\.class$");
 
+        Collections.sort(foundFiles);
```
reta, line 81: @prathyushreddylpr thank you for the fix, although I've never seen this test failing, the File System API indeed does not guarantee an order. There is an issue with this patch, the `Collections.sort` requires the collection to be modifiable which we cannot guarantee.

prathyushreddylpr: Thanks @reta for the feedback. I have changed the code to check the contents of two lists in a better way which would ignore the order of the elements in the list.

reta (approved):

PR #1516, opened 2023-11-10
Cxf dev2

prathyushreddylpr:
### Description
Fixed the flaky tests named `testDeserialization` inside `SerializationTest.java` and `AttributeTest.java` classes. 

https://github.com/apache/cxf/blob/3048785bf03baf76750f448777dfb003b5b1c98e/rt/javascript/javascript-tests/src/test/java/org/apache/cxf/javascript/types/SerializationTest.java#L91

https://github.com/apache/cxf/blob/3048785bf03baf76750f448777dfb003b5b1c98e/rt/javascript/javascript-tests/src/test/java/org/apache/cxf/javascript/types/AttributeTest.java#L73

#### Root Cause
The test `testDeserialization` has been reported as flaky when run with the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool. The tests failed because of the serialization of Testbean and AttributeTestBean classes, where the order of elements in the serialized XML changes between runs. The contents of the serialized strings do not remain constant and hence the tests are failing.  

#### Fix
To define a specific order for the XML elements when the classes TestBean1 and AttributeTestBean are serialized, we need to add the propOrder attribute to the @XmlType annotation. This attribute explicitly specifies the order in which fields should be serialized.

### How this has been tested?

**Maven:** Apache Maven 3.6.3

Command used - 
```
```

Command used - 
```
mvn -pl rt/javascript/javascript-tests test -Dtest=org.apache.cxf.javascript.types.SerializationTest#testDeserialization
```

```
mvn -pl rt/javascript/javascript-tests test -Dtest=org.apache.cxf.javascript.types.AttributeTest#testDeserialization
```

3) **NonDex test**  - Failed
Command used - 
```
mvn -pl rt/javascript/javascript-tests edu.illinois:nondex-maven-plugin:2.1.7-SNAPSHOT:nondex -Dtest=org.apache.cxf.javascript.types.SerializationTest#testDeserialization
```

```

mvn -pl rt/javascript/javascript-tests edu.illinois:nondex-maven-plugin:2.1.7-SNAPSHOT:nondex -Dtest=org.apache.cxf.javascript.types.AttributeTest#testDeserialization -DnondexRuns=10
```

### DIFF
--- a/core/src/test/java/org/apache/cxf/helpers/FileUtilsTest.java
+++ b/core/src/test/java/org/apache/cxf/helpers/FileUtilsTest.java
@@ -24,7 +24,6 @@ import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 
@@ -78,8 +77,6 @@ public class FileUtilsTest {
 
         List<File> foundFiles2 = FileUtils.getFiles(directory, ".*\\.class$");
 
-        Collections.sort(foundFiles);
-        Collections.sort(foundFiles2);
         assertEquals(foundFiles, foundFiles2);
     }
 }
--- a/core/src/test/java/org/apache/cxf/helpers/FileUtilsTest.java
+++ b/core/src/test/java/org/apache/cxf/helpers/FileUtilsTest.java
@@ -24,6 +24,7 @@ import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 
@@ -77,6 +78,8 @@ public class FileUtilsTest {
 
         List<File> foundFiles2 = FileUtils.getFiles(directory, ".*\\.class$");
 
+        Collections.sort(foundFiles);
+        Collections.sort(foundFiles2);
         assertEquals(foundFiles, foundFiles2);
     }
 }
--- a/rt/javascript/javascript-tests/src/test/java/org/apache/cxf/javascript/fortest/AttributeTestBean.java
+++ b/rt/javascript/javascript-tests/src/test/java/org/apache/cxf/javascript/fortest/AttributeTestBean.java
@@ -28,7 +28,13 @@ import jakarta.xml.bind.annotation.XmlType;
  *
  */
 @XmlRootElement(namespace = "uri:org.apache.cxf.javascript.testns")
-@XmlType(namespace = "uri:org.apache.cxf.javascript.testns")
+@XmlType(namespace = "uri:org.apache.cxf.javascript.testns",
+        propOrder = {
+        "element1",
+        "attribute1",
+        "element2",
+        "attribute2"
+        })
 public class AttributeTestBean {
 
     //CHECKSTYLE:OFF
--- a/rt/javascript/javascript-tests/src/test/java/org/apache/cxf/javascript/fortest/TestBean1.java
+++ b/rt/javascript/javascript-tests/src/test/java/org/apache/cxf/javascript/fortest/TestBean1.java
@@ -29,7 +29,21 @@ import jakarta.xml.bind.annotation.XmlType;
  * Bean with a selection of elements suitable for testing the JavaScript client.
  */
 @XmlRootElement(namespace = "uri:org.apache.cxf.javascript.testns")
-@XmlType(namespace = "uri:org.apache.cxf.javascript.testns")
+@XmlType(namespace = "uri:org.apache.cxf.javascript.testns",
+        propOrder = {
+        "stringItem",
+        "intItem",
+        "longItem",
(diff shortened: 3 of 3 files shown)


reta:
@prathyushreddylpr I don't see any comparison logic in `org.apache.cxf.javascript.types.SerializationTest.testDeserialization` that relies on order of the properties. where this hypothesis is coming from?

prathyushreddylpr:
> @prathyushreddylpr I don't see any comparison logic in `org.apache.cxf.javascript.types.SerializationTest.testDeserialization` that relies on order of the properties. where this hypothesis is coming from?

In the 'testDeserialization' there is a call to 'rhinoCallInContext' which internally calls deserializeTestBean3_1(serialized string) here: 'https://github.com/apache/cxf/blob/74069a140737603a0a59435de308401c4351a80b/rt/javascript/javascript-tests/src/test/resources/deserializationTests.js#L33' and the actual comparison is happening here.

reta:
> > @prathyushreddylpr I don't see any comparison logic in `org.apache.cxf.javascript.types.SerializationTest.testDeserialization` that relies on order of the properties. where this hypothesis is coming from?
> 
> In the 'testDeserialization' there is a call to 'rhinoCallInContext' which internally calls deserializeTestBean3_1(serialized string) here: '
> 
> https://github.com/apache/cxf/blob/74069a140737603a0a59435de308401c4351a80b/rt/javascript/javascript-tests/src
> ' and the actual comparison is happening here.

But it compares deserialized form of each attribute, the order does not matter here

prathyushreddylpr:
> > > @prathyushreddylpr I don't see any comparison logic in `org.apache.cxf.javascript.types.SerializationTest.testDeserialization` that relies on order of the properties. where this hypothesis is coming from?
> > 
> > 
> > In the 'testDeserialization' there is a call to 'rhinoCallInContext' which internally calls deserializeTestBean3_1(serialized string) here: '
> > https://github.com/apache/cxf/blob/74069a140737603a0a59435de308401c4351a80b/rt/javascript/javascript-tests/src
> > ' and the actual comparison is happening here.
> 
> But it compares deserialized form of each attribute, the order does not matter here

Yes but when the class is serialized to a string, the attributes does not follow any particular and when the string is compared with certain value it is causing the issue. For example, below are two serialized strings for two different runs:
First:
image

Second:
image

reta:
> Yes but when the class is serialized to a string, the attributes does not follow any particular and when the string is compared with certain value it is causing the issue. For example, below are two serialized strings for two different runs:

Could you please point out where this serialized string is compared to another serialized strings? I have found none such places, and the order does not matter here as far as all attributes have the same values.

prathyushreddylpr:
> > Yes but when the class is serialized to a string, the attributes does not follow any particular and when the string is compared with certain value it is causing the issue. For example, below are two serialized strings for two different runs:
> 
> Could you please point out where this serialized string is compared to another serialized strings? I have found none such places, and the order does not matter here as far as all attributes have the same values.

The comparison is happening here: https://github.com/apache/cxf/blob/74069a140737603a0a59435de308401c4351a80b/rt/javascript/javascript-tests/src/test/resources/deserializationTests.js#L33

image


In the above code bean.getStringItem is expecting "bean1>stringItem". But serialized string contains two string items(example is provided in the above comment) and one of them contained the value "required=true"

reta:
> But serialized string contains two string items(example is provided in the above comment) and one of them contained the value "required=true" and hence the order is important

I don't see to items 

image

and

image

The other two string items belong `beanTwoItem` which is nested serialized bean

reta:
> When propOrder is defined, JAXB takes additional steps to ensure all properties are accounted for and processed systematically, which it might not do when relying on the default (implicit) order.

Thanks @prathyushreddylpr , I am very surprised by your findings but if that is indeed the case, this  definitely not a CXF but JAXB implementation issue (we use Glassfish one), could you please report it there?

prathyushreddylpr:
> Glassfish

@reta , can you please let me know how I can report this to JAXB? Is this:https://github.com/eclipse-ee4j/glassfish the glassfish repo that you mentioned?

reta:
> @reta , can you please let me know how I can report this to JAXB? Is this:https://github.com/eclipse-ee4j/glassfish the glassfish repo that you mentioned?

@prathyushreddylpr  I suppose you could open an issue there with the test case that demonstrates the problem with JAXB implementation (you may need to craft the reproducer without Apache CXF context since as per your explanation, it is purely dictated by JAXB with propOrder setting).

gnodet:
Closing this PR — the root cause has been identified and fixed in #2945.

The non-deterministic serialization order is caused by CXF's own `Utils.java` using `HashSet` to collect bean properties from reflection. `HashSet` has non-deterministic iteration order, so when `@XmlType(propOrder)` is not specified, serialized XML element order varies between runs. The fix switches to `LinkedHashSet` to preserve insertion order.

Adding `propOrder` to the test beans was a valid workaround, but the proper fix is in the serialization infrastructure itself.

Thanks @prathyushreddylpr for identifying the issue!

PR #1528, opened 2023-11-17
[CXF-8959] Fixed order dependent flakiness in AttachmentUtilTest.java in core module

yijut2:
**Description**

Fixed the order dependent flaky test in AttachmentUtilTest.java in core module.

**Root Cause**

The polluter that caused the flakiness in `AttachmentUtilTest.java` was the test class `CachedOutputStreamTest.java`. In `AttachmentUtilTest`, there were multiple tests that used Mockito.spy() to verify the behavior of `CachedOutputStream` object. In those tests, the spy `CachedOutputStream` object "cos" was generated (`CachedOutputStream cos = spy(CachedOutputStream.class`). Then after this, there was a line `verify(cos).setThreshold(102400L)`. Therefore, after the cos being initialized, we need to make sure the setThreshold has been called once. However, it doesn't execute as expected, and the root cause is that 
1. the threshold is not provided when calling `testSetStreamedAttachmentProperties()` method -> so the setThreshold() won't be called in line 198 in AttachmentUtil.java and 
2. the `thresholdSysPropSet` (the boolean that indicated whether the default threshold has been set or not) has been set to true when we run the `CachedOutputStreamTest.java`. Inside the `CachedOutputStreamTest.java`, the unit test `testUseSysPropsWithAttachmentDeserializer()` called the `AttachmentUtil.setStreamedAttachmentProperties(message, cache`), so the `thresholdSysPropSet`has been set to true then. -> so the etThreshold(AttachmentDeserializer.THRESHOLD) won't be called in  line 215 in  AttachmentUtil.java.

Since none of them being called, the verify(cos).setThreshold(102400L) will complained that setThreshold() was not invoked.

 Therefore, if we run `CachedOutputStreamTest.java` before running the `AttachmentUtilTest.java`, the `verify(cos).setThreshold(102400L)` statement will lead to the error message 
```
Wanted but not invoked:
cachedOutputStream.setThreshold(102400L);
-> at org.apache.cxf.attachment.AttachmentUtilTest.bigIntAsAttachmentMaxSize(AttachmentUtilTest.java:279)
```

**Fix log**

After debugging the code, it is clear that the `thresholdSysPropSet` should be `False` every time the CachedOutputStream object was initialized. Therefore, I add the method `resetDefaultThreshold()` in the `CachedOutputStream.java`. In the test class `AttachmentUtilTest`, I called this reset method after initializing the new CachedOutputStream(). After this fix, the failure has been solved. 

**How this has been tested**

Java: openjdk version "17.0.8.1"
Maven: Apache Maven 3.9.5

**How to reproduce the failure**
`mvn install -pl core -am -DskipTests`
(must setup iDFlakies on a Maven project - https://github.com/UT-SE-Research/iDFlakies)
`mvn -pl core idflakies:detect -Ddetector.detector_type=random-class-method -Ddt.randomize.rounds=10 -Ddt.detector.original_order.all_must_pass=false`

### DIFF
--- a/core/src/main/java/org/apache/cxf/io/CachedOutputStream.java
+++ b/core/src/main/java/org/apache/cxf/io/CachedOutputStream.java
@@ -576,6 +576,10 @@ public class CachedOutputStream extends OutputStream {
         this.cipherTransformation = cipherTransformation;
     }
 
+    public void resetDefaultThreshold() {
+        thresholdSysPropSet = false;
+    }
+
     public static void setDefaultMaxSize(long l) {
         if (l == -1) {
             String s = SystemPropertyAction.getProperty(CachedConstants.MAX_SIZE_SYS_PROP, "-1");
--- a/core/src/main/java/org/apache/cxf/io/CachedOutputStream.java
+++ b/core/src/main/java/org/apache/cxf/io/CachedOutputStream.java
@@ -576,10 +576,6 @@ public class CachedOutputStream extends OutputStream {
         this.cipherTransformation = cipherTransformation;
     }
 
-    public void resetDefaultThreshold() {
-        thresholdSysPropSet = false;
-    }
-
     public static void setDefaultMaxSize(long l) {
         if (l == -1) {
             String s = SystemPropertyAction.getProperty(CachedConstants.MAX_SIZE_SYS_PROP, "-1");
@@ -593,9 +589,12 @@ public class CachedOutputStream extends OutputStream {
             i = SystemPropertyAction.getInteger(CachedConstants.THRESHOLD_SYS_PROP, -1);
             if (i <= 0) {
                 i = 128 * 1024;
+                thresholdSysPropSet = false; /* we not using system property value */
             } else {
                 thresholdSysPropSet = true;
             }
+        } else {
+            thresholdSysPropSet = false; /* we not consulting system properties at all */
         }
         defaultThreshold = i;
     }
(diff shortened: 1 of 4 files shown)


reta:
@yijut2 thank you for finding out this issue, I could certainly confirm that this is the case. Regarding the approach, I would suggest to explore the different path which tries to address the cause (vs trying to mitigate the issue). 

It is clear that (as per your investigation) that CachedOutputStreamTest / CachedStreamTestBase leave the `CachedOutputStream` in "non-default" state, I would suggest to fix that by making sure the test cases that manipulate system properties restore the original state (happy to help here).

Thank you.

yijut2:
Thanks for your reply! So I realized that it was because static variables will be shared between instances of the same class within a JVM, and I found this order dependent issue by running multiple tests under the same JVM. I am not sure what would be the best way to address that. Is removing static a possible solution..? Do you have any suggestion?

reta:
> Thanks for your reply! So I realized that it was because static variables will be shared between instances of the same class within a JVM, and I found this order dependent issue by running multiple tests under the same JVM. 

👍 , this is 100% the case

> I am not sure what would be the best way to address that. Is removing static a possible solution..? Do you have any suggestion?

Sure, do you mind if I push some changes to your pull request? Thank you.

yijut2:
Please feel free to make any changes. Thank you!

reta:
> Please feel free to make any changes. Thank you!

Thank you, pushed the change, basically there are 2 fixes:
 1) the `thresholdSysPropSet` flips back to `false` if not set from system properties
 2) the CachedOutputStreamTest / CachedStreamTestBase leave the CachedOutputStream in default state

yijut2:
It makes sense and it is more intuitive. The change really addresses the root cause, thanks for you correction!

reta:
Thank you @yijut2 !

PR #1558, opened 2023-12-06
CXF-8966 : Validation of nillable tags fails 

f2par0:
Change xml validation strategy to use the Woodstox validation when reading instead of having a reader and a writer to a nulloutput because this method is not able to handle the xsi:nil=true attributes.

**Previous  Algorithm :**

- Transform the input to a DOMSource Object
- create a XMLReader from this new object and a XMLWriter to a nulloutput
- setup the wookdstock validation on the writer if possible
- Create a filtered Reader to remove external references
- Copy the filterReader to the Null Writer
- if setup fails, use XOP validation

**Suggested algorithm :**

- create a filtered reader from the input to remove external references
- setup the wookdstock validation on the reader if possible
- call read
- if setup fails, use XOP validation

### DIFF
--- a/core/src/main/java/org/apache/cxf/databinding/source/XMLStreamDataReader.java
+++ b/core/src/main/java/org/apache/cxf/databinding/source/XMLStreamDataReader.java
@@ -28,7 +28,6 @@ import java.util.logging.Logger;
 import javax.xml.namespace.QName;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamReader;
-import javax.xml.stream.XMLStreamWriter;
 import javax.xml.transform.Source;
 import javax.xml.transform.dom.DOMSource;
 import javax.xml.transform.sax.SAXSource;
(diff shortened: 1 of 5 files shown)


reta:
@f2par0 thank you for attempt to fix the issue, I believe the issue is identified here https://github.com/FasterXML/woodstox/issues/179 and it seems like the issue is not within CXF or Woodstox but MSV

f2par0:
Yes, thanks for finding this issue. So this PR is more a workaround to avoid to call MSV writer.

reta:
> Yes, thanks for finding this issue. So this PR is more a workaround to avoid to call MSV writer.

I would prefer the wait for fix to come from upstream if possible and not merge any workarounds

ppalaga:
https://github.com/FasterXML/woodstox/issues/179 is a Woodstox issue after all - see https://github.com/FasterXML/woodstox/issues/179#issuecomment-1879372276

ppalaga:
A fix for https://github.com/FasterXML/woodstox/issues/179 is underway: https://github.com/FasterXML/woodstox/pull/187

NiasSt90:
Why not change in general the approach to run the validation on reader-side instead of on a (dummy) writer during a additional copy process?

I like the approach of these PR here to change.....

I've tried it by myself [here in a similar way.](https://github.com/apache/cxf/compare/main...NiasSt90:cxf:CXF-8979)

ppalaga:
Sorry for the delay, https://github.com/FasterXML/woodstox/pull/187 got stuck due to legal issues with the CLA. I hope those get resolved soonish.

ppalaga:
This issue was fixed in Woodstox via https://github.com/FasterXML/woodstox/pull/245#event-25455969220
The fix is available in Woodstox 7.2.0

reta:
Bumped the dependency here https://github.com/apache/cxf/pull/3138, do we want to keep the tests from this pull request? thanks @ppalaga !

ppalaga:
> do we want to keep the tests from this pull request?

Yeah, that would definitely be useful!

reta:
Closing, tests are incorporated https://github.com/apache/cxf/pull/1558

PR #1563, opened 2023-12-12
only load the lifecycle, bus and event once to make test faster

TestBoost:
The test class `ClassUnwrapperTest` tries to load the `lifecycle`, `bus` and `event` before every test runs in this test class.

For `event`, the test class `ServletContextEvent` creates it with a mocked class and it is only used when starting the application using `lifecycle`.

For `lifecycle`, it is only used to start an application with the `event` and called with `getBeanManager()`.

For `bus`, it is only used to get certain properties of certain class name.

For all these three fields, the two tests inside the `ClassUnwrapperTest` are not trying to modify them but just accessing them. Repeatly creating `lifecycle`, `bus` and `event` object will increase the cost of running tests. When run on our own machine, the test runtime can jump from `5.4858 s` to `4.5919 s` after applying the changes.


reta:
Thank you for contribution @TestBoost !

reta (approved):

PR #1607, opened 2023-12-23
CXF-8955: Custom timeout with Message.RECEIVE_TIMEOUT not working with hc5 / hc

reta:
Custom timeout with Message.RECEIVE_TIMEOUT not working with hc5 / hc


rt/transports/http-hc/src/main/java/org/apache/cxf/transport/http/asyncclient/AsyncHTTPConduit.java
```diff
@@ -245,18 +244,6 @@ public boolean isRepeatable() {
         message.put(CXFHttpRequest.class, e);
     }
 
-    private void propagateJaxwsSpecTimeoutSettings(Message message, HTTPClientPolicy csPolicy) {
```
reta, line 248: The `csPolicy` is shared within same HTTP Async Conduit instance, however the timeouts are per message and override each other (in case each message has contextual timeouts set).

reta:
@ffang I am wondering if you could look at this change, thank you

PR #1608, opened 2023-12-24
Scoped SBOM (cyclonedx-maven-plugin) to release and deploy Apache Maven profiles only

reta:
At the moment, we always generate SBOMs on every build, it significantly slows the builds down. The suggestion is to generate SBOMs at deploy / release time only, when we really need those artifacts to be published (somewhat Inspired by Apache Camel approach https://github.com/apache/camel/blob/main/pom.xml#L801)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -368,6 +368,10 @@
                             </gpgArguments>
                         </configuration>
                     </plugin>
+                    <plugin>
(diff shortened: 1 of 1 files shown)


reta:
@coheigea would love to hear your opinion please, thank you

coheigea (approved):
LGTM, we only need SBOMs at release anyway

PR #1615, opened 2023-12-27
CXF-8968: OpenTelemetryClientFeature cannot be used as a CDI bean unless it has a no-args constructor

ppalaga:
https://issues.apache.org/jira/browse/CXF-8968


reta (approved):
Thank you @ppalaga !

PR #1616, opened 2023-12-28
CXF-8834: "Operation [{.../wsdl}Issue] already exists" when calling JaxWsServerFactoryBean.create() for SecurityTokenService

reta:
Fix "Operation [{.../wsdl}Issue] already exists" when calling JaxWsServerFactoryBean.create() for SecurityTokenService


rt/ws/security/src/main/java/org/apache/cxf/ws/security/sts/provider/SecurityTokenService.java
```diff
@@ -73,7 +73,7 @@ RequestSecurityTokenResponseCollectionType issue(
               partName = "response")
     @Action(input = "http://docs.oasis-open.org/ws-sx/ws-trust/200512/RST/Issue",
             output = "http://docs.oasis-open.org/ws-sx/ws-trust/200512/RSTRC/IssueFinal")
-    @WebMethod(operationName = "Issue")
+    @WebMethod(operationName = "IssueSingle")
```
reta, line 76: The operation was introduced manually (see please d3c73f2af973a875febe3bc6f3f34c251fa5339b), not part of the WS-Trust spec

PR #1633, opened 2024-01-10
Update to Jetty 12

ffang:


rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/HttpUtils.java
```diff
@@ -475,6 +475,9 @@ public static String getProtocolHeader(Message m, String name, String defaultVal
 
     public static String getBaseAddress(Message m) {
         String endpointAddress = getEndpointAddress(m);
+        if (endpointAddress.startsWith("ws")) {
```
reta, line 478: Curios why we substitute `ws` with `http`? I believe the call will fail (plus we also could have `wss://`)

rt/rs/microprofile-client/pom.xml
```diff
@@ -182,14 +182,18 @@
         </dependency>
         <dependency>
             <groupId>org.wiremock</groupId>
-            <artifactId>wiremock</artifactId>
+            <artifactId>wiremock-standalone</artifactId>
```
reta, line 185: Ah yeah, wiremock still uses Jetty 11

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/filters/OAuthRequestFilter.java
```diff
@@ -202,7 +202,7 @@ protected boolean checkRequestURI(HttpServletRequest request, List<String> uris,
         if (uris.isEmpty()) {
             return true;
         }
-        String servletPath = request.getPathInfo();
+        String servletPath = request.getServletPath() + request.getPathInfo();
```
reta, line 205: Should we be vigilant about presence of `/` ?

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/utils/HttpUtils.java
ffang: This change is required to fix some websocket related tests failures with Jetty 12.

We need this change because in Jetty 12 architect and API changes.
For example, if the request url is "ws://localhost:31604/websocket" or "http://localhost:31604/websocket", we want getBaseAddress() returns "/websocket".

I pushed another commit here to honor "wss" protocol as well
https://github.com/apache/cxf/pull/1633/commits/20a0f1c5230c4907b2b03db1a7a12d7a0024baaf

rt/rs/microprofile-client/pom.xml
ffang: Yes, actually this is the major reason why I hold this PR that long. This is the last piece that requires Jetty11, until I found other projects which face the similar issue using wiremock-standalone

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/filters/OAuthRequestFilter.java
ffang: I don't think we need to worry about it.

From the java doc of servlet-api 6
```
/**
     * Returns any extra path information associated with the URL the client sent when it made this request. The extra path
     * information follows the servlet path but precedes the query string and will start with a "/" character.
     *
     * <p>
     * This method returns <code>null</code> if there was no extra path information.
     *
     * @return a <code>String</code> specifying extra path information that comes after the servlet path but before the
     * query string in the request URL; or <code>null</code> if the URL does not have any extra path information. The path
     * will be canonicalized as per section 3.5 of the specification. This method will not return any encoded characters
     * unless the container is configured specifically to allow them.
     * @throws IllegalArgumentException In standard configuration, this method will never throw. However, a container may be
     * configured to not reject some suspicious sequences identified by 3.5.2, furthermore the container may be configured
     * to allow such paths to only be accessed via safer methods like {@link #getRequestURI()} and to throw
     * IllegalArgumentException if this method is called for such suspicious paths.
     */
    public String getPathInfo();
```

So a "/" character is guaranteed here.

rt/rs/microprofile-client/pom.xml
reta: I will open a ticket (but yeah, `wiremock-standalone` is a way to go)

reta:
@ffang just fyi, I had time to go over the half of the change (sorry about that), will try to finish it till the end of the week but please don't wait for me if you need it asap.

ffang:
@reta , no rush, take your time. I will keep this PR open for a while

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/filters/OAuthRequestFilter.java
reta: @ffang just reread the javadoc `This method returns <code>null</code> if there was no extra path information.` - so we still have an issue here, concatenating with `null` would not produce the expected results.

rt/transports/http-jetty/src/main/java/org/apache/cxf/transport/http_jetty/JettyHTTPServerEngine.java
```diff
@@ -456,10 +459,23 @@ public synchronized void addServant(URL url, JettyHTTPHandler handler) {
                             //then it need be on top of JettyHTTPHandler
                             //set JettyHTTPHandler as inner handler if
                             //inner handler is null
-                            ((SecurityHandler)h).setHandler(handler);
+                            //((SecurityHandler)h).setHandler(handler);
```
reta, line 462: It is commented out accidentally or not needed?

rt/transports/http-jetty/src/main/java/org/apache/cxf/transport/http_jetty/JettyHTTPServerEngine.java
```diff
@@ -937,25 +1035,81 @@ private ThreadPool getThreadPool() {
...
                         try {
-                            contexts.removeHandler(handler);
-                            handler.stop();
-                            handler.destroy();
+                            servletHolder.stop();
+                            contextHandler.getContext().destroy(servlet);
+                            //need to remove path from ServletHandler._servletMappings
+                            //and has to access the private field.
+                            
+                            Field privateField
```
reta, line 1088: @ffang could we avoid using reflection? If we really really need that - please use `org.apache.cxf.common.util.ReflectionUtil` (but I would love to see the solution without reflection)

rt/transports/http-jetty/src/main/java/org/apache/cxf/transport/http_jetty/JettyHTTPServerEngine.java
```diff
@@ -1180,5 +1336,137 @@ public int getSessionTimeout() {
     public void setSessionTimeout(int sessionTimeout) {
         this.sessionTimeout = sessionTimeout;
     }
+    
+    class CxfJettyErrorHandler extends ErrorHandler {
```
reta, line 1340:
```suggestion
    static class CxfJettyErrorHandler extends ErrorHandler {
```

rt/transports/websocket/pom.xml
```diff
@@ -156,16 +145,8 @@
...
-            <artifactId>websocket-jakarta-server</artifactId>
-            <version>${cxf.jetty11.version}</version>
-            <optional>true</optional>
-            <exclusions>
-                <exclusion>
-                    <groupId>jakarta.servlet</groupId>
-                    <artifactId>jakarta.servlet-api</artifactId>
-                </exclusion>
-            </exclusions>
+            <groupId>org.eclipse.jetty.ee10.websocket</groupId>
```
reta, line 148:
```
            <optional>true</optional>
```

rt/transports/websocket/src/main/java/org/apache/cxf/transport/websocket/jetty/WebSocketVirtualServletRequest.java
```diff
@@ -545,14 +543,31 @@ public <T extends HttpUpgradeHandler> T upgrade(Class<T> arg0) throws IOExceptio
...
 
-    @Override
+    
     public boolean isRequestedSessionIdFromUrl() {
         LOG.log(Level.FINE, "isRequestedSessionIdFromUrl");
         return false;
     }
+
+    public String getRequestId() {
+        // TODO Auto-generated method stub
```
reta, line 558: Let's get rid of this comments?

rt/transports/websocket/src/main/java/org/apache/cxf/transport/websocket/jetty12/Jetty12WebSocketDestination.java
```diff
@@ -335,5 +342,36 @@ private HttpServletRequest getHttpServletRequest(final UpgradeRequest upgradeReq
         }
     }
 
+    @org.eclipse.jetty.websocket.api.annotations.WebSocket
+    public class Jetty12WebSocket {
```
reta, line 346: Do we need this class to be public?

```suggestion
   static class Jetty12WebSocket {
```

rt/transports/websocket/src/main/java/org/apache/cxf/transport/websocket/undertow/WebSocketUndertowServletRequest.java
```diff
@@ -570,14 +571,31 @@ public <T extends HttpUpgradeHandler> T upgrade(Class<T> arg0) throws IOExceptio
...
 
-    @Override
+
     public boolean isRequestedSessionIdFromUrl() {
         LOG.log(Level.FINE, "isRequestedSessionIdFromUrl");
         return false;
     }
+
+    public String getRequestId() {
+        // TODO Auto-generated method stub
```
reta, line 586: Same, let's get rid of this comments please

systests/microprofile/client/weld/pom.xml
```diff
@@ -178,7 +188,7 @@
                     <dependency>
                         <groupId>com.github.tomakehurst</groupId>
                         <artifactId>wiremock-jre8-standalone</artifactId>
-                        <version>2.35.1</version>
+                        <version>3.0.1</version>
```
reta, line 191: The 2.35.x line was used for Jetty 9 support, if we don't need that, let's use the latest one (please notice group change, `org.wiremock` now):

```
<dependency>
    <groupId>org.wiremock</groupId>
    <artifactId>wiremock-standalone</artifactId>
    <version>3.3.1</version>
    <scope>test</scope>
</dependency>
```

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/filters/OAuthRequestFilter.java
ffang: a Null checker added to address this

rt/transports/http-jetty/src/main/java/org/apache/cxf/transport/http_jetty/JettyHTTPServerEngine.java
ffang: this isn't needed, a new commit removed it

rt/transports/http-jetty/src/main/java/org/apache/cxf/transport/http_jetty/JettyHTTPServerEngine.java
ffang: We have to use ReflectionUtil here. A new commit addressed it

rt/transports/http-jetty/src/main/java/org/apache/cxf/transport/http_jetty/JettyHTTPServerEngine.java
ffang: A new commit addressed it

rt/transports/websocket/pom.xml
ffang: A new commit addressed it

rt/transports/websocket/src/main/java/org/apache/cxf/transport/websocket/jetty/WebSocketVirtualServletRequest.java
ffang: A new commit addressed it

rt/transports/websocket/src/main/java/org/apache/cxf/transport/websocket/jetty12/Jetty12WebSocketDestination.java
ffang: Yes, this needs to be public, so that the annotation of this class
    @org.eclipse.jetty.websocket.api.annotations.WebSocket
can take effect

rt/transports/websocket/src/main/java/org/apache/cxf/transport/websocket/undertow/WebSocketUndertowServletRequest.java
ffang: a new commit addressed it

systests/microprofile/client/weld/pom.xml
ffang: a new commit addressed it.

And thanks for the review and input @reta !

reta:
Thanks @ffang LGTM!

ffang:
Thanks for the review @reta !
Merged

PR #1645, opened 2024-01-12
[CXF-8970]ensure we can build and run without bouncycastle dependencies

ffang:


rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwe/PbesHmacAesWrapKeyEncryptionAlgorithm.java
```diff
@@ -129,19 +128,23 @@ static int getKeySize(String keyAlgoJwt) {
...
+            PBEKeySpec pbeSpec = new PBEKeySpec(new String(password).toCharArray(), saltValue, pbesCount, keySize * 8);
+            SecretKeyFactory keyFact = SecretKeyFactory.getInstance(algorithm); 
+            Key sKey = keyFact.generateSecret(pbeSpec);
+            byte[] ret = new byte[keySize];
+            byte[] key = sKey.getEncoded();
+            System.arraycopy(key, 0, ret, 0, keySize);
+            return ret;
+           
+        } catch (Exception ex) {
+            ex.printStackTrace();
```
andrei-ivanov, line 145: should this be here?

ffang: Good catch,  a new commit addressed this.

Thanks!

rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwa/AlgorithmUtils.java
```diff
@@ -88,9 +88,7 @@ public final class AlgorithmUtils {
     public static final String RS_SHA_256_JAVA = "SHA256withRSA";
     public static final String RS_SHA_384_JAVA = "SHA384withRSA";
     public static final String RS_SHA_512_JAVA = "SHA512withRSA";
-    public static final String PS_SHA_256_JAVA = "SHA256withRSAandMGF1";
-    public static final String PS_SHA_384_JAVA = "SHA384withRSAandMGF1";
-    public static final String PS_SHA_512_JAVA = "SHA512withRSAandMGF1";
```
coheigea, line 93: Why are these removed?

rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwa/AlgorithmUtils.java
```diff
@@ -65,7 +65,7 @@ public final class AlgorithmUtils {
     public static final String A256GCM_ALGO = "A256GCM";
     // Java
     public static final String AES_GCM_ALGO_JAVA = AES + "/GCM/NoPadding";
-    public static final String AES_CBC_ALGO_JAVA = AES + "/CBC/PKCS7Padding";
+    public static final String AES_CBC_ALGO_JAVA = AES + "/CBC/PKCS5Padding";
```
coheigea, line 68: What's the explanation for this change please?

rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwa/SignatureAlgorithm.java
```diff
@@ -31,9 +31,9 @@ public enum SignatureAlgorithm {
     RS384(AlgorithmUtils.RS_SHA_384_ALGO, AlgorithmUtils.RS_SHA_384_JAVA, -1),
     RS512(AlgorithmUtils.RS_SHA_512_ALGO, AlgorithmUtils.RS_SHA_512_JAVA, -1),
 
-    PS256(AlgorithmUtils.PS_SHA_256_ALGO, AlgorithmUtils.PS_SHA_256_JAVA, -1),
-    PS384(AlgorithmUtils.PS_SHA_384_ALGO, AlgorithmUtils.PS_SHA_384_JAVA, -1),
-    PS512(AlgorithmUtils.PS_SHA_512_ALGO, AlgorithmUtils.PS_SHA_512_JAVA, -1),
+    PS256(AlgorithmUtils.PS_SHA_256_ALGO, AlgorithmUtils.PS_SHA_JAVA, -1),
+    PS384(AlgorithmUtils.PS_SHA_384_ALGO, AlgorithmUtils.PS_SHA_JAVA, -1),
+    PS512(AlgorithmUtils.PS_SHA_512_ALGO, AlgorithmUtils.PS_SHA_JAVA, -1),
```
coheigea, line 36: Is it correct to be downgrading from SHA2 to SHA-1 like this?

rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwa/AlgorithmUtils.java
ffang: IIUC, "PKCS7Padding" is only available from BouncyCastle provider, in JDK default security provider, we need to use the name "PKCS5Padding" instead, but this doesn't mean that JDK default provider doesn't support "PKCS7Padding" since AES which is supported by JDK default provider actually requires "PKCS7Padding".  "PKCS5Padding" is a legacy name convention which also covers "PKCS7Padding" if I read a few related discussions correctly.

For example, this one here
https://crypto.stackexchange.com/questions/9043/what-is-the-difference-between-pkcs5-padding-and-pkcs7-padding
```
Many cryptographic libraries use an identifier indicating PKCS#5 or PKCS#7 to define the same padding mechanism. The identifier should indicate PKCS#7 if block sizes other than 8 are used within the calculation. Some cryptographic libraries such as the [SUN provider in Java](http://docs.oracle.com/javase/7/docs/technotes/guides/security/StandardNames.html#Cipher) indicate PKCS#5 where PKCS#7 should be used - "PKCS5Padding" should have been "PKCS7Padding". This is - with high probability - a legacy from the time that only 8 byte block ciphers such as (triple) DES symmetric cipher were available.
```

And this one here
https://copyprogramming.com/howto/java-security-nosuchalgorithmexception-cannot-find-any-provider-supporting-aes-ecb-pkcs7padding
```
 The Sun JCE provider, which is Java's implementation, required a padding method for block size of 16 bytes. To fulfill this requirement, PKCS#7 specifies a padding method that is similar to PKCS#5 padding. The only difference is that PKCS#7 padding is defined for block sizes ranging from 2 to 255 bytes, which is the maximum value of a byte if it encodes a zero-based unsigned integer. However, the padding method was already available with the name "PKCS5Padding" . Instead of introducing a new name, the existing name "PKCS5Padding" was utilized.

At this point, it is crucial that the Sun provider offers support for "PKCS7Padding" since PKCS#5 padding is not appropriate. This is not only a problem for Java Naming issue , but it affects all developers who are attempting to implement cryptographic protocols or transfer other software to Java. While waiting for the Sun provider to offer support, it is recommended to use "PKCS5Padding" instead of "PKCS7Padding" . 
```

rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwa/SignatureAlgorithm.java
ffang: Please see my comment here
https://github.com/apache/cxf/pull/1645#discussion_r1453510256
With PSSParameterSpec specified correctly, actually SHA2 family is used.

rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwa/AlgorithmUtils.java
coheigea: Thanks @ffang

rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwe/PbesHmacAesWrapKeyEncryptionAlgorithm.java
```diff
@@ -129,19 +128,23 @@ static int getKeySize(String keyAlgoJwt) {
...
+            PBEKeySpec pbeSpec = new PBEKeySpec(new String(password).toCharArray(), saltValue, pbesCount, keySize * 8);
+            SecretKeyFactory keyFact = SecretKeyFactory.getInstance(algorithm); 
+            Key sKey = keyFact.generateSecret(pbeSpec);
+            byte[] ret = new byte[keySize];
+            byte[] key = sKey.getEncoded();
+            System.arraycopy(key, 0, ret, 0, keySize);
+            return ret;
+           
+        } catch (Exception ex) {
+            LOG.warning("cannot create Derived Key");
```
coheigea, line 145: Shouldn't the exception be logged here?

rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jws/PublicKeyJwsSignatureVerifier.java
```diff
@@ -59,6 +85,30 @@ public PublicKeyJwsSignatureVerifier(X509Certificate cert,
...
+                break;
+            case "384" : 
+                spec = new PSSParameterSpec(MGF1ParameterSpec.SHA384.getDigestAlgorithm(),  
+                                            "MGF1", MGF1ParameterSpec.SHA384, Integer.valueOf(size) / 8, 1);
+                break;
+            case "512" : 
+                spec = new PSSParameterSpec(MGF1ParameterSpec.SHA512.getDigestAlgorithm(), 
+                                            "MGF1", MGF1ParameterSpec.SHA512, Integer.valueOf(size) / 8, 1);
+                break;
+            default :
```
coheigea, line 106: Could we maybe have some shared code for this PSSParameterSpec creation, as it's re-used 3 times?

rt/security/src/main/java/org/apache/cxf/rt/security/crypto/CryptoUtils.java
```diff
@@ -592,6 +578,11 @@ public static Cipher initCipher(Key secretKey, KeyProperties keyProps, int mode)
         try {
             String algorithm = keyProps != null && keyProps.getKeyAlgo() != null
                 ? keyProps.getKeyAlgo() : secretKey.getAlgorithm();
+            if (algorithm.equals("AESWrap")) {
+                int keySize = secretKey.getEncoded().length;
+                algorithm = "AESWrap_" + keySize * 8;
+                secretKey = new SecretKeySpec(secretKey.getEncoded(), 0, keySize, "AES");
+            }
```
coheigea, line 585: Why is this change needed?

rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jws/PublicKeyJwsSignatureVerifier.java
ffang: addressed by this commit
https://github.com/apache/cxf/pull/1645/commits/9f572761f167d24fbf2a6863571f543ab05559a6

rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwe/PbesHmacAesWrapKeyEncryptionAlgorithm.java
ffang: addressed by https://github.com/apache/cxf/pull/1645/commits/9f572761f167d24fbf2a6863571f543ab05559a6

rt/security/src/main/java/org/apache/cxf/rt/security/crypto/CryptoUtils.java
ffang: For this line
```
algorithm = "AESWrap_" + keySize * 8;
```
This is from java security API doc about AESWrap
```
The AES key wrapping algorithm as described in [RFC 3394](https://tools.ietf.org/html/rfc3394).

To use the AESWrap cipher with only one valid key size, use the format AESWrap_<n>, where <n> can be 128, 192, 
or 256.
```
So I think determine the keySize dynamically and specify it for the algorithm is valid.

For this line 
```
secretKey = new SecretKeySpec(secretKey.getEncoded(), 0, keySize, "AES");
```
This is because without it we get execption like
```
Caused by: java.security.InvalidKeyException: Wrong algorithm: AES or Rijndael required
	at java.base/com.sun.crypto.provider.AESCrypt.init(AESCrypt.java:86)
	at java.base/com.sun.crypto.provider.AESKeyWrap.init(AESKeyWrap.java:102)
	at java.base/com.sun.crypto.provider.KeyWrapCipher.implInit(KeyWrapCipher.java:305)
	at java.base/com.sun.crypto.provider.KeyWrapCipher.engineInit(KeyWrapCipher.java:327)
	at java.base/javax.crypto.Cipher.implInit(Cipher.java:867)
	at java.base/javax.crypto.Cipher.chooseProvider(Cipher.java:929)
	at java.base/javax.crypto.Cipher.init(Cipher.java:1299)
	at java.base/javax.crypto.Cipher.init(Cipher.java:1236)
	at org.apache.cxf.rt.security.crypto.CryptoUtils.initCipher(CryptoUtils.java:588)
```

And take a close look at the JDK impl com.sun.crypto.provider.AESCrypt class
```
if (!algorithm.equalsIgnoreCase("AES")
                    && !algorithm.equalsIgnoreCase("Rijndael")) {
            throw new InvalidKeyException
                ("Wrong algorithm: AES or Rijndael required");
        }

```
So AES (AKA Rijndael) is required here

coheigea (approved):

PR #1670, opened 2024-02-02
Add null guards for ManagedRMEndpoint

pingpingy1:
The `endpoint` parameter passed to the constructor of `ManagedRMEndpoint` is immediately saved to a private property, and cannot be changed later with any public method.
As such, passing `null` into the constructor has little use, throws a `NullPointerException` at a wholly unrelated position and time, and causes difficult debugging should this happen.
This commit implements a fail-fast guard against `null`.

### DIFF
--- a/rt/ws/rm/src/main/java/org/apache/cxf/ws/rm/ManagedRMEndpoint.java
+++ b/rt/ws/rm/src/main/java/org/apache/cxf/ws/rm/ManagedRMEndpoint.java
@@ -23,6 +23,7 @@ import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 import javax.management.JMException;
 import javax.management.ObjectName;
@@ -118,7 +119,7 @@ public class ManagedRMEndpoint implements ManagedComponent {
     }
 
     public ManagedRMEndpoint(RMEndpoint endpoint) {
-        this.endpoint = endpoint;
+        this.endpoint = Objects.requireNonNull(endpoint, "endpoint");
     }
 
     /* (non-Javadoc)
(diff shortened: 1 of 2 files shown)


rt/ws/rm/src/main/java/org/apache/cxf/ws/rm/ManagedRMEndpoint.java
```diff
@@ -118,7 +119,7 @@ public class ManagedRMEndpoint implements ManagedComponent {
     }
 
     public ManagedRMEndpoint(RMEndpoint endpoint) {
-        this.endpoint = endpoint;
+        this.endpoint = Objects.requireNonNull(endpoint, "endpoint");
```
reta, line 122: @pingpingy1 could you please share under which circumstances you see `null`passed here?

pingpingy1: @reta Truth be told, I could imagine very few scenarios in the wild where `null` could be passed here as described.
My PR was intended as a nice-to-have guard for debugging possible future NPEs more quickly, though perhaps in highly unlikely scenarios.
Still, one may imagine a situation where one should create a `ManagedRMEndpoint` for an `RMEndpoint` instance that they hadn't explicitly created, for example to automate a part of the management process.
I do believe that a fail-fast guard against `null` values could go a long way in not just debugging but also consciously preparing for such issues, as I have anecdotally experienced with other projects; hence my PR.

reta: Thanks @pingpingy1 , I would prefer to be "sane" with `null` checks  (we have  thousands of places like that in the codebase), unless we have an issue with this code. The class is created and managed by CXF and it has not been susceptible to NPE here.

PR #1674, opened 2024-02-05
CXF-8975: Missing ending CRLF in multipart request

gchauvet:
Please find attached my pull-request regarding the ticket I reported on https://issues.apache.org/jira/browse/CXF-8975 concerning a serializing issue in multipart message.

Please note that I made the fix for the 3.5.x branch, but that it should also be applied to the 3.1.x branch. However, I've been unable to compile the branch on my Windows workstation due to the default encoding of the generated files, as well as several unit tests that don't pass

### DIFF
--- a/core/src/main/java/org/apache/cxf/attachment/AttachmentSerializer.java
+++ b/core/src/main/java/org/apache/cxf/attachment/AttachmentSerializer.java
@@ -327,7 +327,7 @@ public class AttachmentSerializer {
         StringWriter writer = new StringWriter();
         writer.write("\r\n--");
         writer.write(bodyBoundary);
-        writer.write("--");
+        writer.write("--\r\n");
         out.write(writer.getBuffer().toString().getBytes(encoding));
         out.flush();
     }
--- a/core/src/test/java/org/apache/cxf/attachment/AttachmentSerializerTest.java
+++ b/core/src/test/java/org/apache/cxf/attachment/AttachmentSerializerTest.java
@@ -20,10 +20,13 @@ package org.apache.cxf.attachment;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Properties;
 
 import javax.activation.DataHandler;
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/apache/cxf/attachment/AttachmentSerializer.java
```diff
@@ -327,7 +327,7 @@ public void writeAttachments() throws IOException {
         StringWriter writer = new StringWriter();
         writer.write("\r\n--");
         writer.write(bodyBoundary);
-        writer.write("--");
+        writer.write("--\r\n");
```
reta, line 330: @gchauvet this is not correct as per rfc1341 (see please https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html):

```
Overall, the body of a multipart entity may be specified as follows:
multipart-body := preamble 1*encapsulation 
               close-delimiter epilogue 

encapsulation := delimiter CRLF body-part 

delimiter := CRLF "--" boundary   ; taken from  Content-Type 
field. 
                               ;   when   content-type    is 
multipart 
                             ; There must be no space 
                             ; between "--" and boundary. 

close-delimiter := delimiter "--" ; Again, no  space  before 
"--" 

preamble :=  *text                  ;  to  be  ignored  upon 
receipt. 

epilogue :=  *text                  ;  to  be  ignored  upon 
receipt. 

body-part = <"message" as defined in RFC 822, 
         with all header fields optional, and with the 
         specified delimiter not occurring anywhere in 
         the message body, either on a line by itself 
         or as a substring anywhere.  Note that the 
         semantics of a part differ from the semantics 
         of a message, as described in the text.>
```


The element in question is 
```
close-delimiter := delimiter "--" ; Again, no  space  before 
```
"--"

reta (changes requested):

core/src/main/java/org/apache/cxf/attachment/AttachmentSerializer.java
gchauvet: hello @reta, 

Unless I've misinterpreted the quote below, aren't we supposed to have a CRLF?
> Note that the encapsulation boundary must occur at the beginning of a line, i.e., following a CRLF, and that that initial CRLF is considered to be part of the encapsulation boundary rather than part of the preceding part. The boundary must be followed immediately either by another CRLF and the header fields for the next part, or by two CRLFs, in which case there are no header fields for the next part (and it is therefore assumed to be of Content-Type text/plain).

Or the webservice I'm calling follows RFC 2046 (https://www.rfc-editor.org/rfc/rfc2046#page-44):
```
multipart-body := [preamble CRLF]
                  dash-boundary transport-padding CRLF
                  body-part *encapsulation
                  close-delimiter transport-padding
                  [CRLF epilogue]

gchauvet: Okay, this is a trick from the webservice provider for security purpose. We've found a workaround, there's no longer any problem. thank you.

PR #1728, opened 2024-03-06
[CXF-8984] Check for null before closing input stream

thegli:


ffang (approved):

thegli:
Puuh, right on time :smile:

PR #1777, opened 2024-03-29
CXF-8951: Concurrent WebClient usage causes massive thread overhead

reta:
Concurrent `WebClient` usage causes massive thread overhead. The suggested fix is to share the `HttpClient` instance (with reference counting) between many `HTTPConduit`s based on:
 - client policy 
 - TLS client parameters 

The shared `HttpClient` instances (references) are stored in the static cache capped at 100 (I believe the value is reasonably high taking into account that creation of many client `HTTPConduit`s, each with different policy and TLS settings is highly unlikely). The cache is thread safe and uses array list under the hood (switching to `ConcurrentHashMap` or alike was considered but requires more coding efforts and larger heap footprint which for a collection of 100 elements seems to be not beneficial). 

The client per conduit behaviour  could be forced by setting `share.httpclient.http.conduit` contextual property to `false`(the default value is `true`).

### DIFF
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HttpClientHTTPConduit.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HttpClientHTTPConduit.java
@@ -49,11 +49,14 @@ import java.nio.channels.UnresolvedAddressException;
 import java.security.AccessController;
 import java.security.GeneralSecurityException;
 import java.security.Principal;
+import java.security.PrivilegedAction;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
 import java.security.cert.Certificate;
 import java.time.Duration;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -66,6 +69,8 @@ import java.util.concurrent.Flow;
 import java.util.concurrent.Flow.Subscriber;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Supplier;
 
 import javax.net.ssl.SSLContext;
@@ -82,6 +87,7 @@ import org.apache.cxf.io.CacheAndWriteOutputStream;
 import org.apache.cxf.message.Message;
 import org.apache.cxf.message.MessageUtils;
 import org.apache.cxf.service.model.EndpointInfo;
+import org.apache.cxf.transport.http.policy.impl.ClientPolicyCalculator;
 import org.apache.cxf.transport.https.HttpsURLConnectionInfo;
 import org.apache.cxf.transport.https.SSLUtils;
 import org.apache.cxf.transports.http.configuration.HTTPClientPolicy;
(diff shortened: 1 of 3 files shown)


jimma:
@reta Thanks for improving this ! Just some minor comments: 

-  Do we need to limit the cache size number to make sure the bad usage doesn't consumes all memory ? 

- For the cache implementation, does the ConcunrrentHashMap with the client policy+ssl parameter as the key work for this ?  I am not sure if this is better for the performance or can simplify something .

reta:
> @reta Thanks for improving this ! Just some minor comments:

Thanks a lot @jimma for taking a look

>     * Do we need to limit the cache size number to make sure the bad usage doesn't consumes all memory ?
> 

Certainly, I will add reasonable guards to the cache

>     * For the cache implementation, does the ConcunrrentHashMap with the client policy+ssl parameter as the key work for this ?  I am not sure if this is better for the performance or can simplify something .

I was thinking about that.  `ConcurrentHashMap` would need a bit more work and require larger memory footprint, I am not sure it worth it (at the moment). But I will experiment for sure, I am working on test cases right now that would help us estimate the cache lookup overhead and its heap consumption.

reta:
@jimma @coheigea would appreciate folks if you could take a look, I updated the description to outline the design choices for this fix.

jimma:
@reta  I'll have a look tomorrow.

coheigea (approved):
LGTM Andriy. Do you expect this will resolve all the problems we are seeing with the Conduit recently?

reta:
> LGTM Andriy. Do you expect this will resolve all the problems we are seeing with the Conduit recently?

Thanks @coheigea ! With respect to multithreading (original issue) or/and creating tons of new WebClients - yes, definitely. The overhead of creating new HttpClient instance is massive.

jimma:
@reta  This looks good. I'll try to run some performance tests for this change and it looks this change can improve the client performance too next week. I'll update the result here.

reta:
> @reta This looks good. I'll try to run some performance tests for this change and it looks this change can improve the client performance too next week.

Thanks @jimma ! To be fair, the change shouldn't bring improvements (nor regressions_ on the performance profile but the heap / threads footprint - definitely. Do you want me to hold off the merge or we could do these tests along the way? thanks!

reta:
> Isnt client executor the one to share and wire on httpclient? Httpclient itself does not abuse threads but uses the given ones - except for a few tasks but these ones will use common pool whatever you do

Executor is one thing (and we don't support that with any configuration) but the selector thread is created per client anyway.

jimma:
> > @reta This looks good. I'll try to run some performance tests for this change and it looks this change can improve the client performance too next week.
> 
> Thanks @jimma ! To be fair, the change shouldn't bring improvements (nor regressions) on the performance profile but the heap / threads footprint - definitely. Do you want me to hold off the merge or we could do these tests along the way? thanks!

I thought less client creation and less threads would impact performance, but it isn't. I already ran the tests and result as you expected: this change didn't bring improvement. I already finished the tests and you can go ahead to merge this change. 

For the httpclient executor, should it set to some shared new fixed size thread pool like cpu processor count *2 by default 
instead of using JDK's `Executors.newCachedThreadPool()` ?

reta:
Thanks a lot @jimma !

> For the httpclient executor, should it set to some shared new fixed size thread pool like cpu processor count *2 by default
> instead of using JDK's `Executors.newCachedThreadPool()` ?

I think we could have a follow up improvement to allow executor to be configurable, wdyt?

CantosSong:
> > Isnt client executor the one to share and wire on httpclient? Httpclient itself does not abuse threads but uses the given ones - except for a few tasks but these ones will use common pool whatever you do
> 
> Executor is one thing (and we don't support that with any configuration at the moment) but the selector thread is created per client anyway.

@jimma @reta 
When I used version 4.0.4 without any issues, all subsequent versions had problems.

If the address is modified, it will be released multiple times.

image

image

package com.shangrila.hotel.ca.service.web.tools;

import jakarta.jws.WebMethod;
import jakarta.jws.WebParam;
import jakarta.jws.WebResult;
import org.apache.commons.lang3.RandomUtils;
import org.apache.cxf.binding.soap.SoapMessage;
import org.apache.cxf.interceptor.Fault;
import org.apache.cxf.jaxws.JaxWsProxyFactoryBean;
import org.apache.cxf.phase.AbstractPhaseInterceptor;
import org.apache.cxf.phase.Phase;

import java.util.ArrayList;
import java.util.List;

public class TestCxf {

    public static void main(String[] args) throws InterruptedException {
        HTNGOutboundWebService byAddress = getByAddress();
        int i = 0;
        while (i <= 100) {
            Thread.ofVirtual().start(() -> {
                try {
                    byAddress.ratePlanSubmitRequest("");
                } catch (Exception e) {
                    if (!e.getCause().toString().contains("HTTP response")) {
                        e.printStackTrace();
                    }
                }
            });
            i++;
        }
        Thread.sleep(100000);
    }

    public static class HTNGHeaderOutInterceptor1 extends AbstractPhaseInterceptor<SoapMessage> {
        static List<String> address = new ArrayList();

        public HTNGHeaderOutInterceptor1() {
            super(Phase.PREPARE_SEND);
            address.add("https://www.baidu.com");
            address.add("https://www.tencent.com");
        }

        @Override
        public void handleMessage(SoapMessage message) throws Fault {
            int i1 = RandomUtils.secure().randomInt(0, 2);
            message.put(org.apache.cxf.message.Message.ENDPOINT_ADDRESS, address.get(i1));
        }
    }

    public static HTNGOutboundWebService getByAddress() {
        JaxWsProxyFactoryBean jaxWsProxyFactoryBean = new JaxWsProxyFactoryBean();
        jaxWsProxyFactoryBean.setServiceClass(HTNGOutboundWebService.class);
        jaxWsProxyFactoryBean.getOutInterceptors().add(new HTNGHeaderOutInterceptor1());

        HTNGOutboundWebService soap = (HTNGOutboundWebService) jaxWsProxyFactoryBean.create();
        return soap;
    }

    public interface HTNGOutboundWebService {
        @WebMethod(operationName = "RatePlan_SubmitRequest", action = "http://htng.org/PWSWG/2010/12/RatePlan_SubmitRequest")
        @WebResult(name = "OTA_HotelRatePlanNotifRS", targetNamespace = "http://www.opentravel.org/OTA/2003/05", partName = "OTA_HotelRatePlanNotifRS")
        public void ratePlanSubmitRequest(
                @WebParam(partName = "OTA_HotelRatePlanNotifRQ", name = "OTA_HotelRatePlanNotifRQ", targetNamespace = "http://www.opentravel.org/OTA/2003/05")
                String aa
        );
    }

}

PR #1781, opened 2024-04-01
reuse the same WebClient instance and fix CXF-8992 if it's threadsafe

dennis-yemelyanov:
When the `WebClient` is thread safe, it should be fine to reuse it instead of calling `WebClient.fromClient()`. Additionally, this fixes (at least in cases when the `WebClient` is thread safe) an issue due to GC: https://issues.apache.org/jira/browse/CXF-8992

### DIFF
--- a/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/spec/ClientImpl.java
+++ b/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/spec/ClientImpl.java
@@ -232,6 +232,7 @@ public class ClientImpl implements Client {
         private Configurable<WebTarget> configImpl;
         private UriBuilder uriBuilder;
         private WebClient targetClient;
+        private boolean threadSafeTargetClient;
 
 
         public WebTargetImpl(UriBuilder uriBuilder,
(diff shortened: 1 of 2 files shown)


rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/spec/ClientImpl.java
```diff
@@ -470,7 +472,7 @@ public WebTarget resolveTemplatesFromEncoded(Map<String, Object> templatesMap) {
         private WebTarget newWebTarget(UriBuilder newBuilder) {
             WebClient newClient;
             if (targetClient != null) {
-                newClient = WebClient.fromClient(targetClient);
+                newClient = threadSafeTargetClient ? targetClient : WebClient.fromClient(targetClient);
```
reta, line 475: @dennis-yemelyanov the usage of the new client instance is not accidental and is not about thread safety as much as the following the specification JAX-RS spec (please consult section "5.3 Client Targets" of the JSR-339).

dennis-yemelyanov: ah `WebTarget` must be immutable, I missed that part... I might think more if it's possible to separate the externally visible part of it (like the target URL) from internal details, so the closeable (and finalizable) part can still be reused. But I guess it won't be trivial (if possible at all).

Or we need to adjust some logic from this commit, to ensure different instances don't affect each other when finalized https://github.com/apache/cxf/commit/5ffdb906b560458ad8df5a892554b6e806d98d9e#diff-660ab7b7b6a7b75d899f2a1cd889b168aadd5a194d2aa9fbae875c1b56891c6c

paulbors: Or..... you can make a deep copy of the WebClient inside of WebClient.fromClient() instead of a shallow copy and you never have to worry about any of this (since its an immutable copy of it)

gnodet:
Closing — CXF-8992 was already fixed via PR #2281 (merged March 2025) using a reference-counting approach. This PR's approach also violates JAX-RS spec requirements for WebTarget immutability, as acknowledged by the author.

PR #1794, opened 2024-04-07
[CXF-9001] semeru does not have a native Object.hashCode so OWB proxies hashCode…

rmannibucau:
… (request scoped for ex) and hashset triggering a hashCode in CDI extension it can fail since there is no request scope there by default

### DIFF
--- a/integration/cdi/src/main/java/org/apache/cxf/cdi/JAXRSCdiResourceExtension.java
+++ b/integration/cdi/src/main/java/org/apache/cxf/cdi/JAXRSCdiResourceExtension.java
@@ -81,15 +81,16 @@ public class JAXRSCdiResourceExtension implements Extension {
     private boolean hasBus;
     private Bus bus;
 
-    private final Set< Bean< ? > > applicationBeans = new LinkedHashSet< Bean< ? > >();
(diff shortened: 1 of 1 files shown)


integration/cdi/src/main/java/org/apache/cxf/cdi/JAXRSCdiResourceExtension.java
```diff
@@ -473,7 +473,7 @@ private List< Object > loadServices(final BeanManager beanManager, Collection<Cl
     private List< Object > loadBeans(final BeanManager beanManager, Collection<Class<?>> limitedClasses,
                                      Collection<Bean<?>> beans) {
         // Use set to account for singletons and application scoped beans
```
bmarwell, line 475: The comment now does not align anymore. You'd need to filter out duplicates by other means?

rmannibucau: It never had been right technically or at best it was not reliable with this impl will just drop it.

reta:
@rmannibucau what is semuru? do you mean IBM Semeru ? Mind please creating an issue for CXF so we could track it? thank you

rmannibucau:
@reta done with CXF-9001

reta (approved):

PR #1819, opened 2024-04-18
Add check to prevent redundant values in headers when using GZIPOutInterceptor

yvrng:
I use a custom feature to retry a client request if it fails. Unfortunately, on each attempt, the `Content-Encoding` header is modified to add `gzip`, giving `Content-Encoding: gzip,gzip` on the 2nd attempt.

I've added a small check before adding the value to the header, to make sure it's not already defined with the expected value.

### DIFF
--- a/core/src/main/java/org/apache/cxf/transport/common/gzip/GZIPOutInterceptor.java
+++ b/core/src/main/java/org/apache/cxf/transport/common/gzip/GZIPOutInterceptor.java
@@ -337,7 +337,10 @@ public class GZIPOutInterceptor extends AbstractPhaseInterceptor<Message> {
         if (header.isEmpty()) {
             header.add(value);
         } else {
-            header.set(0, header.get(0) + "," + value);
+            String existingValue = header.get(0);
+            if (!existingValue.contains(value)) {
+                header.set(0, existingValue + "," + value);
+            }
         }
     }
     public void setForce(boolean force) {
--- a/core/src/main/java/org/apache/cxf/transport/common/gzip/GZIPOutInterceptor.java
+++ b/core/src/main/java/org/apache/cxf/transport/common/gzip/GZIPOutInterceptor.java
@@ -323,24 +323,14 @@ public class GZIPOutInterceptor extends AbstractPhaseInterceptor<Message> {
      * @param value the value to add
      */
     private static void addHeader(Message message, String name, String value) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/cxf/transport/common/gzip/GZIPOutInterceptor.java
```diff
@@ -337,7 +337,10 @@ private static void addHeader(Message message, String name, String value) {
         if (header.isEmpty()) {
             header.add(value);
         } else {
-            header.set(0, header.get(0) + "," + value);
+            String existingValue = header.get(0);
+            if (!existingValue.contains(value)) {
```
reta, line 341: I am wondering if this is even correct, why not to add values?
I believe we may just use ` header.add`, the multiple values are concatenated together using `,` (at least, should):

```
if (!header.contains(value)) {
    header.add(value);
}
```
```

Comment (yvrng) [reply core/src/main/java/org/apache/cxf/transport/common/gzip/GZIPOutInterceptor.java]: I was thinking the same, and I confirm, I get the same result by adding the item directly to the list: values are concatenated with a comma.

File: core/src/main/java/org/apache/cxf/transport/common/gzip/GZIPOutInterceptor.java [hunk 2]
```diff
@@ -323,21 +323,14 @@ public void thresholdReached() throws IOException {
...
+            message.put(Message.PROTOCOL_HEADERS, headers);
         }
-        List<String> header = CastUtils.cast((List<?>)protocolHeaders.get(name));
-        if (header == null) {
-            header = new ArrayList<>();
-            protocolHeaders.put(name, header);
-        }
-        if (header.isEmpty()) {
+        List<String> header = headers.computeIfAbsent(name, k -> new ArrayList<>());
+        if (header.isEmpty() || !header.contains(value)) {
```
reta, line 332: @yvrng I think we are missing protocol headers update here:

```
protocolHeaders.put(name, header);
```

reta: My bad, the `protocolHeaders` -> `headers`, missed that

reta (approved):

reta:
Thanks @yvrng !

PR #1822, opened 2024-04-19
Exception for getUserPrincipal() in AbstractHTTPDestination is slurped since CXF 4.1.x

rzo1:
Hi all,

I am currently checking the integration of CXF 4.1.0-SNAPSHOT into TomEE 10 and I noticed a change in `AbstractHTTPDestination`.

An additional `try-catch` was added around `getUserPrincipal()` catching any kind of `Exception` and just returning `null`. 

Was this intentional (or required by the spec)? I think, that we shouldn't catch everything here.

I noticed this change by debugging TomEE's Microprofile JWT integration and one specific TCK test was failing:

```
org.eclipse.microprofile.jwt.tck.container.jaxrs.EmptyTokenTest#invalidToken
```

The provided token is invalid, we cannot parse a principal out of it and since it is in an invalid format anyway, our impl will throw an exception.
This exception is now caught and mapped to `null`, which is quite bad (as it will just return http 200 instead of 401). 

wdyt?

### DIFF
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/AbstractHTTPDestination.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/AbstractHTTPDestination.java
@@ -408,11 +408,7 @@ public abstract class AbstractHTTPDestination
 
         SecurityContext httpSecurityContext = new SecurityContext() {
             public Principal getUserPrincipal() {
-                try {
-                    return req.getUserPrincipal();
-                } catch (Exception ex) {
-                    return null;
-                }
+                return req.getUserPrincipal();
             }
             public boolean isUserInRole(String role) {
                 return req.isUserInRole(role);
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/AbstractHTTPDestination.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/AbstractHTTPDestination.java
@@ -408,7 +408,15 @@ public abstract class AbstractHTTPDestination
 
         SecurityContext httpSecurityContext = new SecurityContext() {
             public Principal getUserPrincipal() {
-                return req.getUserPrincipal();
+                try {
+                    return req.getUserPrincipal();
+                } catch (Exception e) {
+                    if (e instanceof NullPointerException) {
+                        //jetty 12 related
+                        return null;
+                    }
+                    throw e;
+                }
             }
             public boolean isUserInRole(String role) {
                 return req.isUserInRole(role);


rzo1:
Ok from what I can read in the commit history, it seems, it was introduced because of jetty 12?

The underlying exception (reason for failing GH actions) is a NPE in Jetty:

```
[INFO] Running org.apache.cxf.systest.ws.rm.CachedOutClientPersistenceTest
java.lang.NullPointerException: Cannot invoke "org.eclipse.jetty.server.Request.getAttribute(String)" because "request" is null
	at org.eclipse.jetty.server.Request.getAuthenticationState(Request.java:949)
	at org.eclipse.jetty.security.AuthenticationState.getAuthenticationState(AuthenticationState.java:45)
	at org.eclipse.jetty.ee10.servlet.ServletApiRequest.getAuthentication(ServletApiRequest.java:254)
	at org.eclipse.jetty.ee10.servlet.ServletApiRequest.getUndeferredAuthentication(ServletApiRequest.java:259)
	at org.eclipse.jetty.ee10.servlet.ServletApiRequest.getUserPrincipal(ServletApiRequest.java:478)
	at org.apache.cxf.transport.http.AbstractHTTPDestination$2.getUserPrincipal(AbstractHTTPDestination.java:412)
	at org.apache.cxf.ext.logging.event.DefaultLogEventMapper.getPrincipal(DefaultLogEventMapper.java:117)
	at org.apache.cxf.ext.logging.event.DefaultLogEventMapper.map(DefaultLogEventMapper.java:104)
	at org.apache.cxf.ext.logging.LoggingInInterceptor.handleMessage(LoggingInInterceptor.java:93)
	at org.apache.cxf.phase.PhaseInterceptorChain.doIntercept(PhaseInterceptorChain.java:307)
	at org.apache.cxf.phase.PhaseInterceptorChain.resume(PhaseInterceptorChain.java:277)
	at org.apache.cxf.ws.addressing.impl.InternalContextUtils$1.run(InternalContextUtils.java:319)
	at org.apache.cxf.workqueue.AutomaticWorkQueueImpl$3.run(AutomaticWorkQueueImpl.java:413)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at org.apache.cxf.workqueue.AutomaticWorkQueueImpl$AWQThreadFactory$1.run(AutomaticWorkQueueImpl.java:346)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

I've adjusted my PR to only check for the NPE, which was causing the failures. Bascially, the introduced catch of a generic exception was hiding this.

ffang (approved):

rt/transports/http/src/main/java/org/apache/cxf/transport/http/AbstractHTTPDestination.java
```diff
@@ -410,8 +410,12 @@ private boolean isWSAddressingReplyToSpecified(Exchange ex) {
             public Principal getUserPrincipal() {
                 try {
                     return req.getUserPrincipal();
-                } catch (Exception ex) {
-                    return null;
+                } catch (Exception e) {
```
reta, line 413: @rzo1 thanks for catching it, @ffang do you have an idea why `request` could be `null` in `ServletApiRequest`? (may be we have problem somewhere with Jetty 12 integration)

```
     at org.eclipse.jetty.server.Request.getAuthenticationState(Request.java:949)
	at org.eclipse.jetty.security.AuthenticationState.getAuthenticationState(AuthenticationState.java:45)
	at org.eclipse.jetty.ee10.servlet.ServletApiRequest.getAuthentication(ServletApiRequest.java:254)
	at org.eclipse.jetty.ee10.servlet.ServletApiRequest.getUndeferredAuthentication(ServletApiRequest.java:259)
	at org.eclipse.jetty.ee10.servlet.ServletApiRequest.getUserPrincipal(ServletApiRequest.java:478)
	at org.apache.cxf.transport.http.AbstractHTTPDestination$2.getUserPrincipal(AbstractHTTPDestination.java:412)
```

rzo1: In CXF 4.0.x where was'nt a a catch block at all. So this PR only cures a symptom and it might be worth to dig deeper on why it's actually null. Don't know enough about CXF internals though 😄

ffang: I'm on it guys.

Freeman

ffang: Hi Guys,

I think I figured out the root cause of the NPE. 

This will happen in case of oneway or ReplyTo is specified when ws-addressing is used, which means we need to switch thread context on the server side. Basically it's pause the current interceptor chain and resume it in another thread, and before the resumed current interceptor chain is completed, the underlying transport might discard any data on the original stream, that's why we need to cache the InputStream in this case and use it later if necessary. And this peace of code is in
org.apache.cxf.ws.addressing.impl.InternalContextUtils.rebaseResponse. Specifically we have
```
             if (ContextUtils.retrieveAsyncPostResponseDispatch(inMessage) && !robust) {
                        //need to suck in all the data from the input stream as
                        //the transport might discard any data on the stream when this
                        //thread unwinds or when the empty response is sent back
                        DelegatingInputStream in = inMessage.getContent(DelegatingInputStream.class);
                        if (in != null) {
                            in.cacheInput();
                        }
                    }
```
Which in turn will call into org.apache.cxf.transport.http.AbstractHTTPDestination.setupMessage
```
       DelegatingInputStream in = new DelegatingInputStream(req.getInputStream()) {
            public void cacheInput() {
                if (!cached && (exchange.isOneWay() || isWSAddressingReplyToSpecified(exchange))) {
                    //For one-ways and WS-Addressing invocations with ReplyTo address,
                    //we need to cache the values of the HttpServletRequest
                    //so they can be queried later for things like paths and schemes
                    //and such like that.
                    //Please note, exchange used to always get the "current" message
                    exchange.getInMessage().put(HTTP_REQUEST, new HttpServletRequestSnapshot(req));
                }
                if (req.getContentLengthLong() != 0) {
                    super.cacheInput();
                }
            }
            private boolean isWSAddressingReplyToSpecified(Exchange ex) {
                AddressingProperties map = ContextUtils.retrieveMAPs(ex.getInMessage(), false, false, false);
                return map != null && !ContextUtils.isGenericAddress(map.getReplyTo());
            }
        };
```

So in this case the HttpServletRequestSnapshot which is a wrapper of original HttpServletRequest is saved in InMessage, and we should always use pre-saved HTTP_REQUEST from InMessage to populate SecurityContext, as pre-saved HTTP_REQUEST from InMessage could be the original HttpServletRequest or the Cached HttpServletRequest when it's necessary. 

We don't see this change is necessary for Jetty 11 and before. 
Just send a PR https://github.com/apache/cxf/pull/1823 to address this for Jetty12.

Thanks and Best Regards
Freeman

rzo1: Nice, I will close my PR and do a test in TomEE with a local build of your PR tmrw 🙃

ffang: Hi @rzo1 ,

PR merged to main branch, so probably the snapshot build will be available from Apache Snapshot maven repo by tomorrow.  

Cheers
Freeman

rzo1: Hi @ffang ,

TomEE is happy with your fix. CI is green again ;-)
Thx for it !

Gruß
Richard

ffang: Awesome! Thanks for the update @rzo1 !

rzo1: @ffang Did some more tests. 4.1.0-SNAPSHOT only produces 3 JAX-RS TCK Failures (3.1.5) - quite promising :)

ffang: Great news! Thanks again for the update @rzo1 ! Please raise jira ticket and append TCK failures there so that we can take a close look.

Best Regards
Freeman

rzo1: @ffang https://issues.apache.org/jira/browse/CXF-9005

PR #1850, opened 2024-05-06
CXF-8951: Concurrent WebClient usage causes massive thread overhead (the case when single instance of WebClient is used from many threads)

reta:
The original issue was describing the case when WebClient instance is created per thread. However, there is another case when single instance of WebClient is used from many thread, that also led to  massive thread overhead.


jgoodyear:
Branch builds on IBM Semeru 17, Maven 3.9.6, Stream 9 OS, on PPC64LE

PR #1861, opened 2024-05-09
CXF-9007: NullPointerException in XMLStreamDataWriter.writeNode

reta:
NullPointerException in XMLStreamDataWriter.writeNode

### DIFF
--- a/core/src/main/java/org/apache/cxf/helpers/DOMUtils.java
+++ b/core/src/main/java/org/apache/cxf/helpers/DOMUtils.java
@@ -790,7 +790,10 @@ public final class DOMUtils {
             //java9 plus hack
             Field f = GET_DOCUMENT_FRAGMENT_FIELDS.get(fragment.getClass());
             if (f != null) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/cxf/helpers/DOMUtils.java
```diff
@@ -790,7 +790,10 @@ public static DocumentFragment getDomDocumentFragment(DocumentFragment fragment)
             //java9 plus hack
             Field f = GET_DOCUMENT_FRAGMENT_FIELDS.get(fragment.getClass());
             if (f != null) {
-                return ReflectionUtil.accessDeclaredField(f, fragment, DocumentFragment.class);
+                final DocumentFragment inner = ReflectionUtil.accessDeclaredField(f, fragment, DocumentFragment.class);
```
reta, line 793: I am not able to reproduce the issue though but it seems like to be the only place where `null` could escape

PR #1867, opened 2024-05-13
CXF-9013: Move performance benchmark to distribution samples folder

jgoodyear:
Verified existing SOAP Test not affected by updates to TestCaseBase.

JAXRS operation performance metric are intended to provide some baseline, future enhancements may include setting payload sizes for POST/PUT calls.


reta:
@jgoodyear thank you for working on this, I honestly not sure what is the end goal: we do have large amount of tests that do very similar (if not the same) things, we could tag them as "performance" and run as the respective test suite. I think we are adding a large chunk of code that no one will be looking after (much like existing benchmarks project).

jgoodyear:
End goal was to have similar JAX-WS and JAX-RS performance harness co-located for ease of use. 

Having the JAX-RS harness would provide some standardized metric to compare runtimes from Version to Version, or between JVM on the same platform:

Example:

Oracle 17, 1 thread, 60 seconds
MacOS 14.4.1, M2 Max
```
=============Overall Test Result============
Overall Throughput: get 8855.006501950586 (invocations/sec)
Overall AVG. response time: 0.11293046479183494 (ms)
531141.0 (invocations), running 59.982 (sec) 
============================================
```

IBM Semeru 17, 1 thread, 60 seconds
MacOS 14.4.1, M2 Max
```
=============Overall Test Result============
Overall Throughput: get 7387.046735830803 (invocations/sec)
Overall AVG. response time: 0.13537209601632935 (ms)
442883.0 (invocations), running 59.954 (sec) 
============================================
```

We can produce similar comparisons with the JAX-WS harness, wanted to know if JAX-RS had similar stats.

reta:
> End goal was to have similar JAX-WS and JAX-RS performance harness co-located for ease of use.

Fair point, AFAIK the `benchmark` projects in general were not run or even integrated into the build. I would imagine, if we want to make it somewhat useful, we would have to:
 - at least have them as part of the build (profile or something)
 - generate some sort of the report (the one you've been suggesting is sufficient as a starting point)
 - integrate benchmark runs into CI (I could take care of this part)

WDYT? Thanks again for working on this.

jgoodyear:
Alternatively, provide it in the samples folder of the distribution.

reta:
> Alternatively, provide it in the samples folder of the distribution.

As an option, yes.

benchmark/performance/basic_type/bin/run_client
reta, line 1: @jgoodyear we probably have to move `bin/*` as well - the project still have `build.xml`!

jgoodyear: Added them back in. Not sure why/how they did not get included.

PR #1901, opened 2024-05-29
[CXF-9023] Increase unit test coverage on org.apache.cxf.bus

jgoodyear:


core/src/test/java/org/apache/cxf/bus/ManagedBusTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.apache.cxf.Bus;
+import org.apache.cxf.BusFactory;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class ManagedBusTest {
```
reta, line 32: @jgoodyear the `ManagedBus` is covered by `ManagedBusTest` (`systests/uncategorized`)

jgoodyear: Feels odd to have classes tested outside of the package, even to have some basic smoke testing.

reta:
@jgoodyear thanks for working on tests, I believe we should spend a bit more time on crafting more or less realistic test cases, may be under `systests/*` even if needed. The ones here clearly use too much mocking (sometimes it is useful) but really shouldn't, plus there is quite large overlap with `systests/*` that I think we should take into account. Hope it makes sense, thank you.

core/src/test/java/org/apache/cxf/bus/ManagedBusTest.java
reta: CXF is quite old and some decisions as well, but we should not add tests for a sake of tests or/and coverage, it adds significant maintenance costs for no benefits (surely, in context of CXF).

jgoodyear: Right on.

Shall CXF-9023 & CXF-9024 be closed as 'will not fix'?

reta: If you don't mind, let's spend some time and find the gaps, fe. I have not find any tests for `ClientLifeCycleManagerImpl` but you added one (thank again for it), there could be others

jgoodyear: Take any portion of the PR you see value in retaining :)

reta:
Relevant code coverage: https://sonarcloud.io/component_measures?metric=coverage&selected=cxf%3Acore%2Fsrc%2Fmain%2Fjava%2Forg%2Fapache%2Fcxf%2Fbus%2Fmanagers&id=cxf

core/src/test/java/org/apache/cxf/bus/ManagedBusTest.java
reta, line 1: image

core/src/test/java/org/apache/cxf/bus/managers/ClientLifeCycleManagerImplTest.java
reta, line 1: image

core/src/test/java/org/apache/cxf/bus/resource/ResourceManagerImplTest.java
reta, line 1: image

reta (approved):

PR #1929, opened 2024-06-18
[CXF-9030] suppress date on '@Generated' when set

bmarwell:
Fixes CXF-9030


rmannibucau (approved):

reta:
Thanks for the fix, @bmarwell , could you please update `CodeGenOptionTest::testSuppressGeneratedDatePresentOption` to assert that the `@Generated` annotation has no date? Thank you.

reta (approved):

bmarwell:
Oh it's probably because of matcher.matches which checks for the whole line to match exactly. I think it should have been .find instead. 

Will check at home later

bmarwell:
@reta I cannot get it to compile on any of my local machines. Can you please run the tests and see what output  java files they create?

tools/wsdlto/test/src/test/java/org/apache/cxf/tools/wsdlto/jaxws/CodeGenOptionTest.java
```diff
@@ -365,10 +368,10 @@ public void testSuppressGeneratedDateNotPresentOption() throws Exception {
         assertTrue(currentDateInGeneratedAnnotationPresent(str));
         str = IOUtils.readStringFromStream(new FileInputStream(new File(types, "SayHi.java")));
         assertFalse(currentDatePresent(str));
-        assertFalse(currentDateInGeneratedAnnotationPresent(str));
+        assertTrue(currentDateInGeneratedAnnotationPresent(str));
         str = IOUtils.readStringFromStream(new FileInputStream(new File(types, "SayHiResponse.java")));
         assertFalse(currentDatePresent(str));
-        assertFalse(currentDateInGeneratedAnnotationPresent(str));
+        assertTrue(currentDateInGeneratedAnnotationPresent(str));
```
bmarwell, line 374: I assumed something like this. I did not see the config... is that really what we want here? Because the test's name is `testSuppressGeneratedDatePresentOption` …

reta: Yeah, I was caught by confusing test names as well, but this is a change for `testSuppressGeneratedDateNotPresentOption`

```
env.put(ToolConstants.CFG_SUPPRESS_GENERATED_DATE, "true");
```

reta: @bmarwell I think it is good to merge, all good from your perspective? thank you!

bmarwell: Yes, thanks! 👍🏻

bmarwell:
Thanks!!

PR #1962, opened 2024-07-14
CXF-9037: NPE when using com.ctc.wstx.returnNullForDefaultNamespace=true

reta:
NPE when using com.ctc.wstx.returnNullForDefaultNamespace=true

### DIFF
--- a/core/src/main/java/org/apache/cxf/staxutils/StaxSource.java
+++ b/core/src/main/java/org/apache/cxf/staxutils/StaxSource.java
@@ -142,7 +142,10 @@ public class StaxSource extends SAXSource implements XMLReader {
                         if (nsUri == null) {
                             nsUri = "";
                         }
-                        contentHandler.startPrefixMapping(nsPrefix, nsUri);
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/cxf/staxutils/StaxSource.java
```diff
@@ -142,7 +142,10 @@ protected void parse() throws SAXException {
                         if (nsUri == null) {
                             nsUri = "";
                         }
-                        contentHandler.startPrefixMapping(nsPrefix, nsUri);
+                        // see please "com.ctc.wstx.returnNullForDefaultNamespace" property
+                        if (nsPrefix != null) {
```
ud1, line 146: I'm not entirely sure what the purpose of this startPrefixMapping() call is, but I figure it's better to always make the call but use an empty string instead of zero. Just like nsUri above.

reta: Thanks @ud1 , I think it makes perfect sense

PR #1966, opened 2024-07-15
Updating Derby to fix CVE

coheigea:
Fix for https://github.com/apache/cxf/security/dependabot/32


reta:
Yeah, 10.16.1.2 does not exist and 10.17.x needs JDK-21 :(

PR #1972, opened 2024-07-17
Allow JDK23 compatibility for Subject#getSubject

ilgrosso:
As reported by [JDK 23 release notes](https://jdk.java.net/23/release-notes#JDK-8296244):

> The terminally deprecated method Subject.getSubject(AccessControlContext) has been re-specified to throw UnsupportedOperationException if invoked when a Security Manager is not allowed.

The proposed change allows to run flawlessly under JDK 23.

### DIFF
--- a/rt/features/logging/src/main/java/org/apache/cxf/ext/logging/event/DefaultLogEventMapper.java
+++ b/rt/features/logging/src/main/java/org/apache/cxf/ext/logging/event/DefaultLogEventMapper.java
@@ -141,9 +141,16 @@ public class DefaultLogEventMapper {
     }
 
     private Iterator<? extends Object> getJAASPrincipals() {
-        Subject subject = Subject.getSubject(AccessController.getContext());
-        return subject != null && subject.getPrincipals() != null
-            ? subject.getPrincipals().iterator() : Collections.emptyIterator();
+        try {
+            Subject subject = Subject.getSubject(AccessController.getContext());
+            return subject != null && subject.getPrincipals() != null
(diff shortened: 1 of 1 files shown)


reta (approved):

ppalaga:
@reta are there any plans to increase the Java baseline to 18 or newer? If so, we could start using `Subject.current()` instead of `Subject.getSubject(AccessController.getContext())` and catching `UnsupportedOperationException` which may harm the performance. 

Even before upgrading the baseline, we could perhaps avoid calling `Subject.getSubject(AccessController.getContext())` if the current Java Runtime is 18+ and `System.getSecurityManager()` returns null. That might be faster. WDYT?

reta:
> @reta are there any plans to increase the Java baseline to 18 or newer?

For 4.1.0, we are settled on JDK-17 at the moment, I think we could go with at least with JDK-21 for 4.2.0, but it is largely dictated by our dependencies (Spring, etc) and the projects that depend on us (fe Microprofile Rest client, etc), thank @ppalaga

reta:
> Even before upgrading the baseline, we could perhaps avoid calling `Subject.getSubject(AccessController.getContext())` if the current Java Runtime is 18+ and `System.getSecurityManager()` returns null. That might be faster. WDYT?

My apologies @ppalaga , took me a bit more time to look into it. AFAIK, there is no relation between  `System.getSecurityManager()` returning `null` and `Subject.getSubject(AccessController.getContext())` . The only setting that matters is `-Djava.security.manager=allow` which is sadly is hidden inside `System::allowSecurityManager()` class, having access to it would have been helpful to not call `Subject.getSubject()` at all.

PR #1978, opened 2024-07-23
[CXF-9042] Fixed non-idempotent unit tests

kaiyaok2:
Fixes https://issues.apache.org/jira/browse/CXF-9042

# The Problem

Some unit tests are non-idempotent, as they pass in the first run but fail in the second run in the same environment. A fix is necessary since unit tests shall be self-contained, ensuring that the state of the system under test is consistent at the beginning of each test. In practice, fixing non-idempotent tests can help proactively avoid state pollution that results in test order dependency (which could cause problems under test selection, prioritization or parallelization).

# Reproduce

Using the `NIOInspector` plugin that supports rerunning JUnit tests in the same environment. Use `org.apache.cxf.management.InstrumentationManagerTest#testWorkQueueInstrumentation` as an example:
```
cd rt/management
mvn edu.illinois:NIOInspector:rerun -Dtest=org.apache.cxf.management.InstrumentationManagerTest#testWorkQueueInstrumentation

```

# 3 Non-Idempotent Tests & Proposed Fix
## org.apache.cxf.management.InstrumentationManagerTest#testWorkQueueInstrumentation
Reason: The test creates and registers `MBean`s in the `MBeanServer` every time it runs, but it does not properly clean up those `MBean`s after the test execution. As a result, when the test is repeatedly run in the same environment, the number of `MBean`s registered with the `MBeanServer` accumulates, causing `s.size()` to increase.

Error message of the test in the repeated run:
```
java.lang.AssertionError: expected:<2> but was:<4>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.junit.Assert.assertEquals(Assert.java:633)
	at org.apache.cxf.management.InstrumentationManagerTest.testWorkQueueInstrumentation(InstrumentationManagerTest.java:91)
```
Fix: Unregister `MBean`s at the end of the test.

## org.apache.cxf.sts.operation.ValidateJWTTransformationTest#testJWTToSAMLTransformationRealm
Reason: The line `wrapper = (Element) doc.getDocumentElement().appendChild(wrapper);` is not idempotent. This is because each time this line is executed, it attempts to append the wrapper element to the document's root element. If the wrapper element has already been appended to the root (as done in a previous test run), attempting to append it again will violate the document's tree structure, resulting in `org.w3c.dom.DOMException`.

Error message in the repeated run:
```
org.w3c.dom.DOMException: HIERARCHY_REQUEST_ERR: An attempt was made to insert a node where it is not permitted.
	at java.xml/com.sun.org.apache.xerces.internal.dom.CoreDocumentImpl.insertBefore(CoreDocumentImpl.java:439)
	at java.xml/com.sun.org.apache.xerces.internal.dom.NodeImpl.appendChild(NodeImpl.java:230)
	at org.apache.cxf.sts.operation.ValidateJWTTransformationTest.testJWTToSAMLTransformationRealm(ValidateJWTTransformationTest.java:209)
```
Fix: Add a checker to avoid appending the `wrapper` if it is already part of the document.

## org.apache.cxf.systest.dispatch.DispatchClientServerTest#testJAXBObjectPAYLOADWithFeature
Reason: The static counters in `TestDispatchFeature` and `TestInInterceptor` gets incremented repeatedly when `testJAXBObjectPAYLOADWithFeature()` is repeatedly run. The counters will be 1 in the first test run, but will be 2 in the second run and so on.

Error message  in the repeated run:
```
java.lang.AssertionError: Feature should be applied expected:<1> but was:<2>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.apache.cxf.systest.dispatch.DispatchClientServerTest.testJAXBObjectPAYLOADWithFeature(DispatchClientServerTest.java:691)
```
Fix:  Save the initial counts of the `TestDispatchFeature` before creating the bus, and then comparing the increments in the assertion checks.

# Verifying this change
After the patch, running the tests repeatedly in the same environment will not lead to failures.

### DIFF
--- a/services/sts/sts-core/src/test/java/org/apache/cxf/sts/operation/ValidateJWTTransformationTest.java
+++ b/services/sts/sts-core/src/test/java/org/apache/cxf/sts/operation/ValidateJWTTransformationTest.java
@@ -206,9 +206,7 @@ public class ValidateJWTTransformationTest {
         TokenProviderResponse providerResponse = createJWT();
         Element wrapper = createTokenWrapper((String)providerResponse.getToken());
         Document doc = wrapper.getOwnerDocument();
-        if (doc.getDocumentElement() == null) {
-            wrapper = (Element)doc.appendChild(wrapper);
-        }
+        wrapper = (Element)doc.appendChild(wrapper);
 
         ValidateTargetType validateTarget = new ValidateTargetType();
         validateTarget.setAny(wrapper);
@@ -340,7 +338,7 @@ public class ValidateJWTTransformationTest {
     }
 
     private Element createTokenWrapper(String token) {
-        Document doc = DOMUtils.getEmptyDocument();
+        Document doc = DOMUtils.createDocument();
         Element tokenWrapper = doc.createElementNS(null, "TokenWrapper");
         tokenWrapper.setTextContent(token);
         return tokenWrapper;
--- a/services/sts/sts-core/src/test/java/org/apache/cxf/sts/operation/ValidateJWTTransformationTest.java
+++ b/services/sts/sts-core/src/test/java/org/apache/cxf/sts/operation/ValidateJWTTransformationTest.java
@@ -206,7 +206,9 @@ public class ValidateJWTTransformationTest {
         TokenProviderResponse providerResponse = createJWT();
         Element wrapper = createTokenWrapper((String)providerResponse.getToken());
         Document doc = wrapper.getOwnerDocument();
-        wrapper = (Element)doc.appendChild(wrapper);
+        if (doc.getDocumentElement() == null) {
+            wrapper = (Element)doc.appendChild(wrapper);
+        }
 
         ValidateTargetType validateTarget = new ValidateTargetType();
         validateTarget.setAny(wrapper);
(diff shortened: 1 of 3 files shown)


services/sts/sts-core/src/test/java/org/apache/cxf/sts/operation/ValidateJWTTransformationTest.java
```diff
@@ -206,7 +206,9 @@ public void testJWTToSAMLTransformationRealm() throws Exception {
         TokenProviderResponse providerResponse = createJWT();
         Element wrapper = createTokenWrapper((String)providerResponse.getToken());
         Document doc = wrapper.getOwnerDocument();
-        wrapper = (Element)doc.appendChild(wrapper);
+        if (doc.getDocumentElement() == null) {
```
reta, line 209: @kaiyaok2 this change is a bit suspicious, why the document element will be null here?

reta: @kaiyaok2 I think this test should be refactored to create a new document (vs manipulating empty one which is static):

```
    private Element createTokenWrapper(String token) {
        Document doc = DOMUtils.createDocument();
        Element tokenWrapper = doc.createElementNS(null, "TokenWrapper");
        tokenWrapper.setTextContent(token);
        return tokenWrapper;
    }

```

Could you please make a change? Thank you.

reta (approved):

reta:
Thank you @kaiyaok2 !

PR #1989, opened 2024-08-03
CXF-9041: CXF will only accept org.glassfish.jaxb.runtime.v2.ContextFactory

reta:
CXF will only accept org.glassfish.jaxb.runtime.v2.ContextFactory. Switching the discovery mechanism to ServiceLoader instead.

### DIFF
--- a/core/src/main/java/org/apache/cxf/common/jaxb/JAXBUtils.java
+++ b/core/src/main/java/org/apache/cxf/common/jaxb/JAXBUtils.java
@@ -49,6 +49,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.StringTokenizer;
 import java.util.logging.Level;
@@ -69,6 +70,7 @@ import org.xml.sax.EntityResolver;
 import org.xml.sax.InputSource;
 
 import jakarta.xml.bind.JAXBContext;
+import jakarta.xml.bind.JAXBContextFactory;
 import jakarta.xml.bind.JAXBElement;
 import jakarta.xml.bind.JAXBException;
 import jakarta.xml.bind.Marshaller;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/cxf/common/jaxb/JAXBUtils.java
```diff
@@ -1176,18 +1179,48 @@ private static Object createEscapeHandler(Class<?> cls, String simpleClassName)
...
         try {
             ctx = AccessController.doPrivileged(new PrivilegedExceptionAction<JAXBContext>() {
                 public JAXBContext run() throws Exception {
-                    //This is a workaround for CXF-8675
-                    Class<?> factoryClass = ClassLoaderUtils.loadClass("org.glassfish.jaxb.runtime.v2.ContextFactory",
-                            JAXBContextCache.class);
-                    Method m = factoryClass.getMethod("createContext", Class[].class, Map.class);
-                    Object context = m.invoke(null, classes.toArray(new Class<?>[0]), map);
-                    return (JAXBContext) context;
+                    final String factoryClassName = System.getProperty(JAXBContext.JAXB_CONTEXT_FACTORY);
```
reta, line 1190: @rmannibucau would appreciate if you validate the approach here, I think that would solve the underlying issue and properly instantiate the JAXB provider of choice, thanks in advance!

rmannibucau: Why not using first JAXBContext factory methods, now it is on jakarta they are there so reflection is not needed and all the new logic looks like the factoryfinder of the jaxb so can be worth avoiding to duplicate for cases it will be different (thinking to OSGi quite loud but I have other cases in mind like some tweaks in EE servers).

reta: Very valid point, it does look like factoryfinder, but something like should do it:

```
return JAXBContext.newInstance(classes.toArray(new Class<?>[0]), map);
```

Is that what you had in mind? thanks!

PR #2018, opened 2024-08-19
CXF-8931: HttpClientHTTPConduit can't disable the http chunk mode

reta:
HttpClientHTTPConduit can't disable the http chunk mode. Drafting for now and working on test cases.


reta:
@jimma would really appreciate if you could validate my approach here, thank you!

PR #2056, opened 2024-09-09
CXF-9054 do not use Date and SimpleDateFormat when parsing LocalDate,…

vershnik:
… use the appropriate DateTimeFormatter for LocalDate

### DIFF
--- a/rt/rs/extensions/search/src/main/java/org/apache/cxf/jaxrs/ext/search/AbstractSearchConditionParser.java
+++ b/rt/rs/extensions/search/src/main/java/org/apache/cxf/jaxrs/ext/search/AbstractSearchConditionParser.java
@@ -343,7 +343,7 @@ public abstract class AbstractSearchConditionParser<T> implements SearchConditio
             if (LocalTime.class.isAssignableFrom(valueType)) {
                 return LocalTime.parse(value);
             } else if (LocalDate.class.isAssignableFrom(valueType)) {
-                return LocalDate.from(convertToDefaultDate(value).toInstant().atZone(ZoneId.systemDefault()));
+                return LocalDate.parse(value, SearchUtils.getLocalDateFormat(contextProperties));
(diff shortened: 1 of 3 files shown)


reta:
Thanks @vershnik , as you pointed out, this is very likely JDK related issue [1] that is not related to CXF (and impact all other applications out there).

[1] https://bugs.openjdk.org/browse/JDK-8294357

vershnik:
Even though it may turn out to be JDK issue, I think using DateTimeFormatter for parsing LocalDate is more appropriate and performance efficient (no need to create instances of SimpleDateFormat, Date, Instant, ZonedDateTime)

reta:
> Even though it may turn out to be JDK issue, I think using DateTimeFormatter for parsing LocalDate is more appropriate and performance efficient (no need to create instances of SimpleDateFormat, Date, Instant, ZonedDateTime)

The `DateTimeFormatter` is certainly never API than `SimpleDateFormat` but to my knowledge,  `DateTimeFormatter` and `SimpleDateFormat` are not compatible [1], [2] (please correct me if I am wrong here). We could bring the changes than risk to break existing applications but those won't be backported. 

To your point, it would make sense to migrate to `java.time` APIs but that should not be limited to `LocalDate` only (for example,  the `SearchUtils` still use  `SimpleDateFormat`).

[1] https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html
[2] https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html

vershnik:
yes, looks like there is incompatibility between `DateTimeFormatter`  and `SimpleDateFormat`  
For example `SimpleDateFormat`  for pattern 'yyyy-MM-dd' can parse even dates that contain time: 
`parse("2019-07-07 23:59:59")`

while  `DateTimeFormatter`  is more strict here. So this change cannot be backported.

PR #2069, opened 2024-09-17
[CXF-9059] HttpConduit should flush output stream 

jgoodyear:
HttpConduit should flush output stream ahead of closing connections when HttpClient is autocloseable.

When running WebClient in a load test (single host) we'll observe low throughput.

When we adjust the HttpConduit to flush output stream ahead of connection close, we observe significant throughput improvement. 

Given this appears to affect just autocloseable HttpClient, we select to flush the output stead only when the JVM provides the autocloseable HttpClient variety.

Tested on:
Ubuntu 22.04 LTS with Java 17 and 21.
CentOS Stream 9 (PPC64LE) with Java 17 and 21.
Windows 11 Pro (x64) with Java 17 and 21.
MacOS 14.6.1 with Java 17 and 21.

Notes: 
Azul Zulu and Eclipse Adoptium used for OpenJDK builds.
IBM Semeru used for OpenJ9 coverage.

### DIFF
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
@@ -32,7 +32,6 @@ import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.net.http.HttpClient;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
@@ -715,11 +714,6 @@ public abstract class HTTPConduit
                 }
             }
         } finally {
-            OutputStream os = msg.getContent(OutputStream.class);
-            // Java 21 may hang on close, we flush stream to help close them out.
-            if (os != null && AutoCloseable.class.isAssignableFrom(HttpClient.class)) {
-                os.flush();
-            }
             super.close(msg);
         }
     }
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
@@ -32,6 +32,7 @@ import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.http.HttpClient;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


rt/transports/http/src/main/java/org/apache/cxf/transport/http/HTTPConduit.java
```diff
@@ -714,6 +715,11 @@ public void close(Message msg) throws IOException {
                 }
             }
         } finally {
+            OutputStream os = msg.getContent(OutputStream.class);
```
reta, line 718: This is very strange (I will try to find out what is happening) but this change if needed belongs to `HttpClientHTTPConduit`

jgoodyear: Ok - i'll update the branch :)

jgoodyear: Branch is updated.

Running some tests to ensure the ephemeral port issue remains tamed.

jgoodyear: Appears good to me after moving the code.

rt/transports/http/src/main/java/org/apache/cxf/transport/http/HttpClientHTTPConduit.java
```diff
@@ -243,13 +243,18 @@ private boolean isSslTargetDifferent(URI lastURL, URI url) {
                 || !lastURL.getHost().equals(url.getHost())
                 || lastURL.getPort() != url.getPort();
     }
-    
+
     @Override
     public void close(Message msg) throws IOException {
+        OutputStream os = msg.getContent(OutputStream.class);
+        // Java 21 may hang on close, we flush stream to help close them out.
+        if (os != null && AutoCloseable.class.isAssignableFrom(HttpClient.class)) {
```
reta, line 251: Think it though, the `os.flush()`  my end up with IOException, should be safer to:

```

        try (OutputStream os = msg.getContent(OutputStream.class)) {
            // Java 21 may hang on close, we flush stream to help close them out.
            if (os != null && AutoCloseable.class.isAssignableFrom(HttpClient.class)) 
                 os.flush();
            } 
        }  finally {
           super.close(msg);
           msg.remove(HttpClient.class);
        }
```

Wdyt?

jgoodyear: Running a build with the above 
Will let you know how it tests

jgoodyear: NettyHttpConduit does not like it.
Trying a different impl.

reta: NettyHttpConduit  should not be impacted, right?

reta (approved):

PR #2075, opened 2024-09-19
CXF-9060 Only import java.net.URI when necessary to avoid unused import warnings

instagibb:
## Changes
 - Modified `service.vm` template  to only import `java.net.URI` when necessary to avoid unused import warnings


reta:
Thank you @instagibb

reta (approved):

PR #2111, opened 2024-10-17
CXF-7396: CachedOutputStream doesn't delete temp files. Add forced cleanup on Bus shutdown

reta:
Add forced cleanup on Bus shutdown


core/src/main/java/org/apache/cxf/io/DelayedCachedOutputStreamCleaner.java
```diff
@@ -226,11 +235,19 @@ public void initComplete() {
     
     @Override
     public void postShutdown() {
+        // If cleanup on shutdown is asked, force cleaning all cached output streams
+        if (cleanupOnShutdown) {
+            forceClean();
```
reta, line 240: @coheigea minor improvement to forceably cleanup all cached output streams upon `Bus` shutdown

coheigea (approved):

PR #2137, opened 2024-11-06
CXF-9076: Exception message is not unmarshalled with JDK17+

reta:
Exception message is not unmarshalled with JDK17+


huhmare:
I am little bit late for commenting the PR, but it seems to me that this aproach might prevent usage of custom subbed Exception, that contain custom member variables. The cloning is done with certain constructor, so the custom subclass will not be initialized correctly. I apologize, that I can not try this out just now.
						
Is this something that you might consider a problem? Then again, maybe the messageDetail is more important, if we have to choose.

Here is the subclass with a member to make the explanation clear:			
```
		class ExampleException extends Exception {
			public int errorCode;
			
			public ExampleException(String msg, Throwable cause, int errorCode) {
				this(msg cause)
				this.errorCode = errorCode;
			}
			public ExampleException(String message, Throwable ex) {
				super(msg, cause)
			}
		}
```

PR #2144, opened 2024-11-11
Fix possible IllegalStateException when bundle is stopped

tjwatson:


rt/features/logging/src/main/java/org/apache/cxf/ext/logging/osgi/Activator.java
```diff
@@ -121,6 +121,16 @@ public void updated(Dictionary config) throws ConfigurationException {
             }
         }
 
+        private ServiceRegistration<?> safeRegister(String name, Object service,
```
reta, line 124: Thanks @tjwatson , I am wondering if we could implement more robust solution by relying on:

```
    @Override
    public void stop(BundleContext context) throws Exception {

    }
```
?

reta (approved):

PR #2148, opened 2024-11-12
CXF-9076: Exception message is not unmarshalled with JDK17+ (limit the fix to JDK standard library exceptions)

reta:
Exception message is not unmarshalled with JDK17+ (limit the fix to JDK standard library exceptions)

### DIFF
--- a/core/src/main/java/org/apache/cxf/interceptor/ClientFaultConverter.java
+++ b/core/src/main/java/org/apache/cxf/interceptor/ClientFaultConverter.java
@@ -200,26 +200,44 @@ public class ClientFaultConverter extends AbstractInDatabindingInterceptor {
                     ReflectionUtil.setAccessible(f);
                     f.set(ex, fault.getMessage());
                 } catch (Exception e1) {
-                    try {
-                        // Fallback, try to clone the exception instead of accessing the detailMessage over reflection
-                        final Constructor<? extends Object> constructor = ReflectionUtil.getConstructor(ex.getClass(),
-                            String.class, Throwable.class); /* String message, Throwable cause */
-
-                        if (constructor != null) {
-                            final Exception clone = (Exception) constructor.newInstance(message, ex.getCause());
-                            clone.setStackTrace(ex.getStackTrace());
-                            if (ex.getSuppressed().length > 0) {
-                                Arrays.stream(ex.getSuppressed()).forEach(clone::addSuppressed);
-                            }
-                            ex = clone;
-                        }
-                    } catch (Exception e2) {
-                        /* nothing to do */
+                    if (isJdkException(ex.getClass().getPackageName())) {
+                        ex = cloneJdkException(ex, message);
                     }
                 }
             }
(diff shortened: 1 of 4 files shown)


reta:
@huhmare the fix should address your concerns, thank you

huhmare:
Thank you.

The CXF tests are run with "--add-opens java.base/java.lang=ALL-UNNAMED" by default. The test ClientServerExceptionTest.exceptionCustomExceptionMessageIsPreserved fails without it. The test is bit misleading, as the CXF-9076 is about using jdk17+ without add-opens.

As I understand, using this PR with JDK17+ without add-opens:
- java.lang.*Exceptions will have detailMessages, if it has a constructor(String, Throwable).
- Any other (Sub)Exception will not have it, so that we can have member variables in subclasses.

Difference to before CXF-9076 is that we will have detailMessage in some java.lang.*Exceptions. The necessary constructor is found in 11 / 29 Exceptions (openJDK17+35). For example SecurityException has it, and IllegalAccessException does not.

There are 3 exceptions (ClassNotFoundException, EnumConstantNotPresentException, TypeNotPresentException) in java.lang that have their own member variables. Only TypeNotPresent will be affected by these CXF-9076 changes. Not a big problem. 


All in all, this and the previous PR are small step to right direction, but are not a complete solution to the general problem.

In the code base I am working with, we still probably have to use either:
* add-opens OR
* rewrite the code to not rely on detailMessage OR
* rewrite subExceptions to hide Throwable.detailMessage with a our own detailMessage and get/set methods in the subclass. 

These PRs might help us in case we have a java.lang.*Exception thrown out of the SOAP-interface. But it will not cover for example java.sql.SQLException.


Unfortunately I do not have a suggestion for a correct way of doing this. Dynamic Proxy works only with interfaces(?), so a way to create instance with getMessage returning correct value would propably need bytecode manipulation stuff.

reta:
> The CXF tests are run with "--add-opens java.base/java.lang=ALL-UNNAMED" by default. The test ClientServerExceptionTest.exceptionCustomExceptionMessageIsPreserved fails without it. The test is bit misleading, as the CXF-9076 is about using jdk17+ without add-opens.

That is right, although I've tested it manually, the tests have to be run without `--add-opens`, I will try to isolate those since it is not only related to this specific test case

> As I understand, using this PR with JDK17+ without add-opens:

    java.lang.*Exceptions will have detailMessages, if it has a constructor(String, Throwable).
    Any other (Sub)Exception will not have it, so that we can have member variables in subclasses.

Right

> All in all, this and the previous PR are small step to right direction, but are not a complete solution to the general problem.

Right, this is a best effort. 

> Unfortunately I do not have a suggestion for a correct way of doing this. Dynamic Proxy works only with interfaces(?), so a way to create instance with getMessage returning correct value would propably need bytecode manipulation stuff.

I think it all runs down to the edge cases (and how often they appear in the wild), if we cover 80%+ with this fix, it is already a win I believe. Thanks @huhmare

PR #2150, opened 2024-11-14
CXF-9082: Allow to modify sensitive header list

dmytro-sylaiev:
* Fix any sensitive information is printed to the logs

https://issues.apache.org/jira/browse/CXF-9082

### DIFF
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/Headers.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/Headers.java
@@ -111,7 +111,7 @@ public class Headers {
         Map<String, List<Object>> filteredHeaders = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
         filteredHeaders.putAll(headers);
         if (!logSensitiveHeaders) {
-            for (String filteredKey : SENSITIVE_HEADERS) {
+            for (String filteredKey : getSensitiveHeaderList()) {
                 filteredHeaders.put(filteredKey, SENSITIVE_HEADER_MARKER);
             }
         }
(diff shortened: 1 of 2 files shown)


rt/transports/http/src/main/java/org/apache/cxf/transport/http/Headers.java
```diff
@@ -580,4 +580,8 @@ public static String toHttpDate(Date date) {
     public static String toHttpLanguage(Locale locale) {
         return locale.toString().replace('_', '-');
     }
+
+    public static List<String> getSensitiveHeaderList() {
```
reta, line 584: Please convert `SENSITIVE_HEADERS` to the contextual property:

```
private static final String SENSITIVE_HEADERS = "org.apache.http.sensitive.headers";


List<String> getSensitiveHeaders() {
    return MessageUtils.getContextualStrings(message, SENSITIVE_HEADERS, Set.of("Authorization", "Proxy-Authorization");
}
```

reta (changes requested):

rt/transports/http/src/main/java/org/apache/cxf/transport/http/Headers.java
dmytro-sylaiev: Hello @reta 
Both `toString()` and `logProtocolHeaders` methods (who are the only users of `SENSITIVE_HEADERS` list) are static and don't use `message` to print headers.

reta: Hello @dmytro-sylaiev , that is right, those are internal APIs (package protected) and has to be changed to instance methods (or accept message as an argument), thanks.

dmytro-sylaiev:
Hello @reta 
I've changed the PR to make sensitive headers a contextual property as you asked, but now for me is not clear how to pass the value when use `WebClient:invoke`, seems the end-user can not modify the result Message properties

reta:
> I've changed the PR to make sensitive headers a contextual property as you asked, but now for me is not clear how to pass the value when use `WebClient:invoke`, seems the end-user can not modify the result Message properties

Thanks @dmytro-sylaiev , there a multiple ways to provide these properties, to list just a few:
 - using `Bus` instance
 - using `WebClient.getConfig(wc).getRequestContext().put(...)` context

reta (approved):

PR #2184, opened 2024-12-06
Update Jackson to 2.18.2

reta:
Update Jackson to 2.18.2


systests/rs-security/src/test/java/org/apache/cxf/systest/jaxrs/security/jose/jwejws/JAXRSJweJwsTest.java
```diff
@@ -85,8 +82,7 @@ public void testJweJwkPlainTextRSA() throws Exception {
     @Test
     public void testJweJwkBookBeanRSA() throws Exception {
         String address = "https://localhost:" + PORT + "/jwejwkrsa";
-        BookStore bs = createJweBookStore(address,
-                                       Collections.singletonList(new JacksonJsonProvider()));
+        BookStore bs = createJweBookStore(address, List.of());
```
reta, line 85: It turns out, Jackson for never used in serde process :)) With 2.18.x, it kicks in but conflicts with CXF's JSONProvider

PR #2219, opened 2025-01-08
CXF-9095 fix connection leak

suxuan-king1:


ffang (changes requested):
Hi @suxuan-king1 ,

How about we put all potential resources check and close in one final block, something like this
```
-            super.close();
-            if (pout != null) {
-                pout.close();
-                pout = null;
-            }
-            if (publisher != null) {
-                publisher.close();
-                publisher = null;
+            try {
+                super.close();
+            } finally {
+                if (pout != null) {
+                    try {
+                        pout.close();
+                    } catch (IOException e) {
+                        logStackTrace(e);
+                    }
+                    pout = null;
+                }
+                if (publisher != null) {
+                    try {
+                        publisher.close();
+                    } catch (IOException e) {
+                        logStackTrace(e);
+                    }
+                    publisher = null;
+                }
+                request = null;
+                subscribers = null;
             }
-            request = null;
-            subscribers = null;
+            
+            

```

Freeman

ffang:
Hi @suxuan-king1 ,

Could you please revise your PR accordingly as we discussed?

Thanks!
Freeman

utpalch:
Any info on when the codebase will be merged ?

PR #2234, opened 2025-01-25
CXF-9057: Chunked Stream is closed regularly when Exception is thrown (MTOM)

reta:
Chunked Stream is closed regularly when Exception is thrown (MTOM)

### DIFF
--- a/core/src/main/java/org/apache/cxf/phase/PhaseInterceptorChain.java
+++ b/core/src/main/java/org/apache/cxf/phase/PhaseInterceptorChain.java
@@ -325,6 +325,9 @@ public class PhaseInterceptorChain implements InterceptorChain {
                     }
                     pause();
                     throw ex;
+                } catch (AbortedInvocationException ex) {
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/apache/cxf/phase/PhaseInterceptorChain.java
```diff
@@ -325,6 +325,9 @@ public synchronized boolean doIntercept(Message message) {
                     }
                     pause();
                     throw ex;
+                } catch (AbortedInvocationException ex) {
+                    abort();
```
reta, line 329: @coheigea @ffang folks would appreciate if you could validate my implementation here. Introduction of `AbortedInvocationException` is a best effort attempt to terminate chunking transfer that may fail in the middle. Sadly, there are no standard means to deal with that (AFAIK).

ffang (approved):
LGTM, thanks @reta !

PR #2237, opened 2025-01-27
CXF-9003: Name clash when two SEIs have a same name method in the same Java package

ppalaga:
https://issues.apache.org/jira/browse/CXF-9003

### DIFF
--- a/rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/spi/WrapperClassNamingConvention.java
+++ b/rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/spi/WrapperClassNamingConvention.java
@@ -1,0 +1,148 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.cxf.jaxws.spi;
+
+import java.util.Locale;
+import java.util.regex.Pattern;
+
+import org.apache.cxf.common.util.PackageUtils;
+/**
+ * Provides names for storing generated wrapper classes.
+ *
+ * @since 4.1.1
+ */
+public interface WrapperClassNamingConvention {
+
+    /**
+     * Returns a package name unique for the given {@code sei} and {@code anonymous}
+     * parameters suitable for storing generated wrapper classes.
+     *
+     * @param sei the service endpoint interface for which the package name should be created
+     * @param anonymous whether the generated wrapper types are anonymous
+     * @return a valid Java package name
+     */
+    String getWrapperClassPackageName(Class<?> sei, boolean anonymous);
+
+    /**
+     * Default naming scheme since CXF 4.2.0.
+     * <p>
+     * The package name returned by {@link #getWrapperClassPackageName(Class, boolean)} are unique
+     * per given {@code sei}.
+     * <p>
+     * Examples:
+     * <table>
+     * <tr>
+     * <th>SEI</th><th>anonymous</th>
+     * <th>{@code getWrapperClassPackageName()} return value</th>
+     * </tr>
+     * <tr>
+     * <td>{@code org.example.Service}</td>
+     * <td>{@code false}</td>
+     * <td>{@code org.example.jaxws_asm.service}</td>
+     * </tr>
+     * <tr>
+     * <td>{@code org.example.OuterClass$Service}</td>
+     * <td>{@code false}</td>
+     * <td>{@code org.example.jaxws_asm.outerclass_service}</td>
+     * </tr>
+     * <tr>
+     * <td>{@code org.example.Service}</td>
+     * <td>{@code true}</td>
+     * <td>{@code org.example.jaxws_asm_an.service}</td>
+     * </tr>
+     * </table>
+     *
+     * @since 4.1.1
+     */
+    class DefaultWrapperClassNamingConvention implements WrapperClassNamingConvention {
+        public static final String DEFAULT_PACKAGE_NAME = "defaultnamespace";
+        private static final Pattern JAVA_PACKAGE_NAME_SANITIZER_PATTERN = Pattern.compile("[^a-zA-Z0-9_]");
+
+        @Override
+        public String getWrapperClassPackageName(Class<?> sei, boolean anonymous) {
+            final String className = sei.getName();
+            final int start = className.startsWith("[L") ? 2 : 0;
+            final int end = className.lastIndexOf('.');
+            final String pkg;
+            final String cl;
+            if (end >= 0) {
+                pkg = className.substring(start, end);
+                cl = className.substring(end + 1);
+            } else {
+                pkg = DefaultWrapperClassNamingConvention.DEFAULT_PACKAGE_NAME;
+                cl = className;
+            }
+
+            return pkg
+                    + (anonymous ? ".jaxws_asm_an." : ".jaxws_asm.")
+                    + JAVA_PACKAGE_NAME_SANITIZER_PATTERN.matcher(cl).replaceAll("_").toLowerCase(Locale.ROOT);
+        }
+
+    }
+
+    /**
+     * An implementation restoring the behavior of CXF before version 4.2.0.
+     * <p>
+     * Unlike with {@link DefaultWrapperClassNamingConvention}, this implementation's
+     * {@link #getWrapperClassPackageName(Class, boolean)} takes only package name
+     * of the given {@code sei} into account.
+     * Therefore naming clashes may occur if two SEIs are in the same package
+     * and both of them have a method with the same name but possibly different signature.
+     * <p>
+     * Examples:
+     * <table>
+     * <tr>
+     * <th>SEI</th>
+     * <th>anonymous</th>
+     * <th>{@code getWrapperClassPackageName()} return value</th>
+     * </tr>
+     * <tr>
+     * <td>{@code org.example.Service}</td>
+     * <td>{@code false}</td>
+     * <td>{@code org.example.jaxws_asm}</td>
+     * </tr>
+     * <tr>
+     * <td>{@code org.example.OuterClass$Service}</td>
+     * <td>{@code false}</td>
+     * <td>{@code org.example.jaxws_asm}</td>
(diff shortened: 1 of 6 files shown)


ppalaga:
Could you please review @reta or anybody else?

rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/spi/WrapperClassUtils.java
```diff
@@ -0,0 +1,60 @@
...
+     * Returns a package name unique for the declaring class of the given {@code method} and {@code anonymous}
+     * parameters suitable for generated classes.
+     *
+     * @param method the web method for which the package name should be created
+     * @param anonymous whether the wrapper are anonymous
+     * @return a valid Java package name
+     */
+    public static String getPackageName(Method method, boolean anonymous) {
+        String className = method.getDeclaringClass().getName();
+        if (className.startsWith("[L")) {
```
reta, line 42: Should we generalize a little bit array handling? (not only Object[]):

```suggestion
        if (className.startsWith("[")) {
```

reta: Oh, it is coming from PackageUtils::getPackageName, should be good :-)

rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/spi/WrapperClassUtils.java
```diff
@@ -0,0 +1,60 @@
...
+public final class WrapperClassUtils {
+    public static final String DEFAULT_PACKAGE_NAME = "defaultnamespace";
+    private static final Pattern JAVA_PACKAGE_NAME_SANITIZER_PATTERN = Pattern.compile("[^a-zA-Z0-9_]");
+
+    private WrapperClassUtils() {
+    }
+
+    /**
+     * Returns a package name unique for the declaring class of the given {@code method} and {@code anonymous}
+     * parameters suitable for generated classes.
```
reta, line 34: @ppalaga thanks for javadoc, could you please document a few examples how package name is generated? For example (from tests):

```
org.apache.cxf.systest.jaxws.CXF9003Test$Hello1Service -> org.apache.cxf.systest.jaxws.jaxws_asm.cxf9003test_hello1service
```

Thank you.

reta:
Thanks a lot @ppalaga !

ppalaga:
> Thanks a lot @ppalaga, but I believe this change could be considered a breaking one (please correct me if I am wrong): the previously generated classes would end up under different package. This could be disruptive,

Yes, you are right, the generated classes now end up in a different package. The classification of the change depends a bit from how CXF defines its public API. While I'd personally tend to see this as an implementation detail not exposed to CXF users, you may know of use cases where users rely on the package name. I'd let you decide because I do not think I am experienced enough with CXF. 

Another aspect is whether this is actually a bug in plain CXF. When writing the test, I found out that the issue is not happening in stock CXF where only `WrapperClassGenerator` is active. That one is using per service class loaders and it is thus able to provide multiple wrapper classes with the same FQ name. The issue only happens when there is a `GeneratedClassClassLoaderCapture` storing the classes per class name and a `WrapperClassLoader` using them and loading in flat class loader. That's the way we do it in Quarkus CXF. But still, the issue is reproducible on plain CXF using solely CXF classes in a way they were (I believe) designed for. So as long as this issue could be seen as a CXF bug, we perhaps may apply measures "breaking" some internal details.

> what do you think if we introduce the package generation extension (SimplePackageNameGenerator, the default / PerSeiPackageNameGenerator) so it would be picked from the Bus?

Thanks for the idea! That would certainly solve the problem. I'd be fine going that way, although, in the long term having less configuration knobs is better for maintenance and testing. Wouldn't suffice to introduce the extension only if there comes somebody really needing the old behavior?

ppalaga:
> I found out that the issue is not happening in stock CXF where only `WrapperClassGenerator` is active. That one is using per service class loaders and it is thus able to provide multiple wrapper classes with the same FQ name. The issue only happens when there is a `GeneratedClassClassLoaderCapture` storing the classes per class name and a `WrapperClassLoader` using them and loading in flat class loader.

Hm... when thinking of it more, another possible solution strategy could be to pass the service name to `GeneratedClassClassLoaderCapture`, let it store per service and make `WrapperClassLoader` also load the classes through per-service class loaders... I am not quite sure we should go that way, because it would be more complicated than the current solution.

ffang:
Thanks for the contribution @ppalaga!

And about the backward compatibility concerns, how about we maintain a cache in WrapperClassGenerator(or wherever more suitable), so if there's no conflict for the wrapper class names, we use the previous way to generate the package name, if we find the conflict to generate the same wrapper class in the same package(which didn't work previously), we use the new way to add SEI classname to generate the package name. This way, the scenarios which have been working are the same package name and same behaviour as before, but we also make it working for the scenarios that haven't been working(the cases described in CXF-9003). 

WDYT?

Freeman

reta:
Thanks @ppalaga 

> Wouldn't suffice to introduce the extension only if there comes somebody really needing the old behavior?

I think we could indeed pick the new strategy as default but provide the fallback to the old one if users run into issues.

> Hm... when thinking of it more, another possible solution strategy could be to pass the service name to GeneratedClassClassLoaderCapture, let it store per service and make WrapperClassLoader also load the classes through per-service class loaders...

I think that would be more disruptive change since `GeneratedClassClassLoaderCapture` should be changed (yes, we could make it non-breaking but make it bloated).

> And about the backward compatibility concerns, how about we maintain a cache in WrapperClassGenerator(or wherever more suitable), so if there's no conflict for the wrapper class names, we use the previous way to generate the package name, if we find the conflict to generate the same wrapper class in the same package(which didn't work previously), we use the new way to add SEI classname to generate the package name

I think this would also work, thank you @ffang

rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/spi/WrapperClassUtils.java
ppalaga: Addressed in 6582f43

ppalaga:
6582f43:

* Improved the JavaDoc as suggested by @reta 
* Introduced a new extension, called `WrapperClassNamingConvention` for the naming. 
* Provided two implementations of the above: one new solving the current issue (default) and one minicking the behavior before this fix.
* Both implementations are covered in `WrapperClassLoaderTest`

rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/spi/WrapperClassNamingConvention.java
```diff
@@ -0,0 +1,148 @@
...
+     * <td>{@code org.example.jaxws_asm}</td>
+     * </tr>
+     * <tr>
+     * <td>{@code org.example.Service}</td>
+     * <td>{@code true}</td>
+     * <td>{@code org.example.jaxws_asm_an}</td>
+     * </tr>
+     * </table>
+     *
+     * @since 4.2.0
```
reta, line 132: Super nit, I think we could target it to 4.1.1
```suggestion
     * @since 4.1.1
```

rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/spi/WrapperClassNamingConvention.java
```diff
@@ -0,0 +1,148 @@
...
+     * <td>{@code org.example.jaxws_asm.outerclass_service}</td>
+     * </tr>
+     * <tr>
+     * <td>{@code org.example.Service}</td>
+     * <td>{@code true}</td>
+     * <td>{@code org.example.jaxws_asm_an.service}</td>
+     * </tr>
+     * </table>
+     *
+     * @since 4.2.0
```
reta, line 71:
```suggestion
     * @since 4.1.1
```

rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/spi/WrapperClassNamingConvention.java
```diff
@@ -0,0 +1,148 @@
...
+package org.apache.cxf.jaxws.spi;
+
+import java.util.Locale;
+import java.util.regex.Pattern;
+
+import org.apache.cxf.common.util.PackageUtils;
+/**
+ * Provides names for storing generated wrapper classes.
+ *
+ * @since 4.2.0
```
reta, line 28:
```suggestion
 * @since 4.1.1
```

rt/frontend/jaxws/src/test/java/org/apache/cxf/jaxws/spi/WrapperClassLoaderTest.java
```diff
@@ -115,8 +119,31 @@ public void testWrapperClassLoaderWhenNoWrappedOperations() throws Exception {
...
+        final List<String> loadedClassNames = testWrapperClassLoaderWithNamingConvention(
+                new WrapperClassNamingConvention.LegacyWrapperClassNamingConvention());
+        assertEquals(
+                List.of(
+                        "org.apache.cxf.jaxws.service.jaxws_asm.SayHi",
+                        "org.apache.cxf.jaxws.service.jaxws_asm.SayHiResponse"),
+                loadedClassNames);
+    }
+
+    public List<String> testWrapperClassLoaderWithNamingConvention(WrapperClassNamingConvention convention)
```
reta, line 143:
```suggestion
    private List<String> testWrapperClassLoaderWithNamingConvention(WrapperClassNamingConvention convention)
```

reta:
Love it @ppalaga , thank you, a few nits but LGTM otherwise, @ffang wdyt? Thanks!

ffang:
> Love it @ppalaga , thank you, a few nits but LGTM otherwise, @ffang wdyt? Thanks!

LGTM, thanks @ppalaga , @reta !

rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/spi/WrapperClassNamingConvention.java
ppalaga: Addressed in adc8f6f

rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/spi/WrapperClassNamingConvention.java
ppalaga: Addressed in adc8f6f

rt/frontend/jaxws/src/main/java/org/apache/cxf/jaxws/spi/WrapperClassNamingConvention.java
ppalaga: Addressed in adc8f6f

rt/frontend/jaxws/src/test/java/org/apache/cxf/jaxws/spi/WrapperClassLoaderTest.java
ppalaga: Done in da140fe

reta (approved):

ffang (approved):

reta:
> Thanks for the review and suggestions, gentlemen!

Thanks a lot @ppalaga !

PR #2245, opened 2025-01-31
[CXF-9105] Synchronize iteration over JAAS principals Set to prevent ConcurrentModificationException

sstremler:
https://issues.apache.org/jira/browse/CXF-9105

### DIFF
--- a/rt/features/logging/src/main/java/org/apache/cxf/ext/logging/event/DefaultLogEventMapper.java
+++ b/rt/features/logging/src/main/java/org/apache/cxf/ext/logging/event/DefaultLogEventMapper.java
@@ -19,6 +19,7 @@
 package org.apache.cxf.ext.logging.event;
 
 import java.security.AccessController;
+import java.security.Principal;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
@@ -109,7 +110,7 @@ public class DefaultLogEventMapper {
     }
 
     private String getPrincipal(Message message) {
-        String principal = getJAASPrincipal();
+        String principal = getConcatenatedJAASPrincipals();
         if (principal != null) {
             return principal;
         }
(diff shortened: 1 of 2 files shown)


rt/features/logging/src/main/java/org/apache/cxf/ext/logging/event/DefaultLogEventMapper.java
```diff
@@ -125,31 +126,36 @@ private String getPrincipal(Message message) {
...
     }
 
-    private String getJAASPrincipal() {
-        StringBuilder principals = new StringBuilder();
-        Iterator<? extends Object> principalIt = getJAASPrincipals();
-        while (principalIt.hasNext()) {
-            principals.append(principalIt.next());
-            if (principalIt.hasNext()) {
-                principals.append(',');
+    public String getConcatenatedJAASPrincipals() {
```
reta, line 129:
```suggestion
    private String getConcatenatedJAASPrincipals() {
```

rt/features/logging/src/main/java/org/apache/cxf/ext/logging/event/DefaultLogEventMapper.java
```diff
@@ -125,31 +126,36 @@ private String getPrincipal(Message message) {
...
+        StringBuilder principalsStringBuilder = new StringBuilder();
+        Set<Principal> principals = getJAASPrincipals();
```
reta, line 131: I believe we could do:

```
if (principals.isEmpty()){
    return null;
}
```

since `getJAASPrincipals()` may return the empty set which does not need synchronization (moreover, it is backed by static instance).

reta:
Thank you @sstremler , a few comments but LGTM otherwise!

sstremler:
> Thank you @sstremler , a few comments but LGTM otherwise!

Thank you, I added the changes you suggested.

reta (approved):

PR #2253, opened 2025-02-03
CXF-9107 Clean up coverity resource leaks and a javascript method typo

cunningt:
https://issues.apache.org/jira/browse/CXF-9107

Clean up coverity resource leaks (inputstream and resultset) and a javascript method typo

### DIFF
--- a/integration/jca/src/main/java/org/apache/cxf/jca/core/classloader/PlugInClassLoader.java
+++ b/integration/jca/src/main/java/org/apache/cxf/jca/core/classloader/PlugInClassLoader.java
@@ -95,7 +95,7 @@ public class PlugInClassLoader extends SecureClassLoader {
 
         props.load(in);
         LOG.fine("Contents: " + propsFileName + props);
-
+        in.close();
         return props;
     }
 
--- a/rt/rs/extensions/providers/src/main/java/org/apache/cxf/jaxrs/provider/json/JSONProvider.java
+++ b/rt/rs/extensions/providers/src/main/java/org/apache/cxf/jaxrs/provider/json/JSONProvider.java
@@ -239,8 +239,8 @@ public class JSONProvider<T> extends AbstractJAXBProvider<T>  {
             } else {
                 response = checkAdapter(response, type, anns, false);
             }
+            realStream.close();
             return type.cast(response);
-
         } catch (JAXBException e) {
             handleJAXBException(e, true);
         } catch (XMLStreamException e) {
--- a/rt/ws/rm/src/main/java/org/apache/cxf/ws/rm/persistence/jdbc/RMTxStore.java
+++ b/rt/ws/rm/src/main/java/org/apache/cxf/ws/rm/persistence/jdbc/RMTxStore.java
@@ -886,6 +886,8 @@ public class RMTxStore implements RMStore {
             while (rs.next()) {
                 dbCols.add(rs.getString(4));
             }
+            rs.close();
+
             for (String[] col : tableCols) {
                 if (!dbCols.contains(col[0])) {
                     newCols.add(col);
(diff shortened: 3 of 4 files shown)


integration/jca/src/main/java/org/apache/cxf/jca/core/classloader/PlugInClassLoader.java
```diff
@@ -95,7 +95,7 @@ private static Properties getProperties(ClassLoader parent, String propsFileName
 
         props.load(in);
         LOG.fine("Contents: " + propsFileName + props);
-
+        in.close();
```
reta, line 98: Could we wrap it in `try(InputStream in = parent.getResourceAsStream(propsFileName)) {...}` please?

rt/rs/extensions/providers/src/main/java/org/apache/cxf/jaxrs/provider/json/JSONProvider.java
```diff
@@ -239,8 +239,8 @@ public T readFrom(Class<T> type, Type genericType, Annotation[] anns, MediaType
             } else {
                 response = checkAdapter(response, type, anns, false);
             }
+            realStream.close();
```
reta, line 242: So this may not work: the stream is not created here and should not be closed here

rt/ws/rm/src/main/java/org/apache/cxf/ws/rm/persistence/jdbc/RMTxStore.java
```diff
@@ -886,6 +886,8 @@ protected void verifyTable(Connection con, String tableName, String[][] tableCol
             while (rs.next()) {
                 dbCols.add(rs.getString(4));
             }
+            rs.close();
```
reta, line 889: Same, could we please wrap in it a `try` block? thank you

integration/jca/src/main/java/org/apache/cxf/jca/core/classloader/PlugInClassLoader.java
cunningt: Needed to rewrite a bit of this in order to do this, as try-with-resources does not allow a stream that is set in try-with-resources to be reset later within the try.      Split it out into two try-with-resources statements.

rt/ws/rm/src/main/java/org/apache/cxf/ws/rm/persistence/jdbc/RMTxStore.java
cunningt: Wrapped in try

rt/rs/extensions/providers/src/main/java/org/apache/cxf/jaxrs/provider/json/JSONProvider.java
cunningt: Can we do this within a try-with-resources?

reta: I think it is being done this way (higher in the stack), when the request / response body is being processed

reta: @cunningt could you please revert this one change and we are good to go, as per spec:

> The caller is responsible for ensuring that the input stream ends when the entity has been consumed. The implementation should not close the input stream.

gnodet:
Superseded by #2943 which cherry-picks the two valid fixes (PlugInClassLoader and RMTxStore). The JavaScript typo fix was already merged via #2781, and the JSONProvider change was not applied per JAX-RS spec (the caller is responsible for closing the input stream).

PR #2259, opened 2025-02-06
[CXF-9108] Remove leading slash in resource name when calling ClassLoader's getResourceAsStream method

AmshikaH:
## Purpose

Addresses https://issues.apache.org/jira/browse/CXF-9108.

### DIFF
--- a/core/src/main/java/org/apache/cxf/version/Version.java
+++ b/core/src/main/java/org/apache/cxf/version/Version.java
@@ -29,7 +29,7 @@ public final class Version {
     private static String fullVersion;
     private static String buildNumber;
 
-    private static final String VERSION_BASE = "/org/apache/cxf/version/";
+    private static final String VERSION_BASE = "org/apache/cxf/version/";
 
     private Version() {
         // utility class - never constructed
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/cxf/version/Version.java
```diff
@@ -29,19 +29,15 @@ public final class Version {
     private static String fullVersion;
     private static String buildNumber;
 
-    private static final String VERSION_BASE = "/org/apache/cxf/version/";
+    private static final String VERSION_BASE = "org/apache/cxf/version/";
```
reta, line 32: @AmshikaH thank you for looking into this, but I believe this change has high risks of breaking things. The CXF deals with many containers (notably, OSGi / Apache Karaf / Apache Felix) that have own classloaders and are sensitive to resource naming.

reta (changes requested):

core/src/main/java/org/apache/cxf/version/Version.java
AmshikaH: @reta, noted, thank you for the feedback.

PR #2260, opened 2025-02-07
rename badly named parameters

Enrice:
just fix the parameter names


gnodet:
Superseded by #2942 which applies the same fix on current main, also covering the new attribute methods and fixing the typo in `inSensitiveAttirbuteNames`.

PR #2262, opened 2025-02-07
CXF-9091: Cache payload for oneway

mash-sap:


mash-sap:
The message is already cache here:
https://github.com/apache/cxf/blob/3.6.x-fixes/rt/ws/addr/src/main/java/org/apache/cxf/ws/addressing/impl/InternalContextUtils.java#L276

Maybe one would rather set here the property for cache?
inMessage.put("cxf.io.cacheinput", Boolean.TRUE)

ffang:
> The message is cache here wihtout setting "cxf.io.cacheinput": https://github.com/apache/cxf/blob/3.6.x-fixes/rt/ws/addr/src/main/java/org/apache/cxf/ws/addressing/impl/InternalContextUtils.java#L276
> 
> Maybe one would set "cxf.io.cacheinput" in InternalContextUtils instead of in MAPAggregatorImpl? inMessage.put("cxf.io.cacheinput", Boolean.TRUE)

Hi @mash-sap ,

Yes, I think this is better.

Could you please revise the PR accordingly?

Thanks!
Freeman

mash-sap:
Hi @ffang , thanks for looking into this. I updated the PR.
Best regards,
Manuel

ffang (approved):

mash-sap:
Thanks for the approval, @ffang. What else is required for this change to be merged?

mash-sap:
Thanks @ffang

PR #2275, opened 2025-02-18
CXF-9110: Fix extensive memory usage with enabled LoggingFeature

yurkom:


core/src/main/java/org/apache/cxf/io/DelayedCachedOutputStreamCleaner.java
```diff
@@ -167,7 +167,8 @@ public boolean equals(Object obj) {
             }
             
             final DelayedCloseable other = (DelayedCloseable) obj;
-            return Objects.equals(closeable, other.closeable);
+            //  because of the broken Liskov Substitution Principle, check in two ways to find equal streams
+            return Objects.equals(other.closeable, closeable) || Objects.equals(closeable, other.closeable);
```
reta, line 171: Thank you @yurkom , this is clearly violates the `equals` contract, I believe we need to find out why and fix the root cause of it.

yurkom: Hello,
The violation is in [CachedOutputStream](https://github.com/apache/cxf/blob/3efe2deba4086aae0b9ebd6f7ee16253decc7c82/core/src/main/java/org/apache/cxf/io/CachedOutputStream.java#L243). When obj compares with not wrapped instance:
```
        return currentStream.equals(obj);
```
I have not found more elegant solution to fix the issue than above, because of replacing `currentStream.equals(obj);` with `false` breaks other tests and does not solve the issue with memory consumption.

reta: Thanks @yurkom , I will take a look as well

reta: @yurkom do you mind if I push some changes to your pull request? I think I found out the problem, thank you

yurkom: Hi @reta 
Of course you can.

reta: Sorry @yurkom , I cannot push to your branch (may be because you pushed into `main`), but I've cherrypicked you changes over here https://github.com/apache/cxf/pull/2278, thank you

yurkom:
Done in #2278

PR #2370, opened 2025-04-16
swaref attachement cid association failed

qq1038765585:
Or am I using it improperly?

### DIFF
--- a/rt/databinding/jaxb/src/main/java/org/apache/cxf/jaxb/attachment/JAXBAttachmentMarshaller.java
+++ b/rt/databinding/jaxb/src/main/java/org/apache/cxf/jaxb/attachment/JAXBAttachmentMarshaller.java
@@ -80,7 +80,7 @@ public class JAXBAttachmentMarshaller extends AttachmentMarshaller {
         AttachmentImpl att = new AttachmentImpl(id, handler);
         att.setXOP(false);
         atts.add(att);
-        return id;
+        return "cid:" + id;
     }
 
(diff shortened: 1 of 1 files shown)


rt/databinding/jaxb/src/main/java/org/apache/cxf/jaxb/attachment/JAXBAttachmentMarshaller.java
```diff
@@ -80,7 +80,7 @@ public String addSwaRefAttachment(DataHandler handler) {
         AttachmentImpl att = new AttachmentImpl(id, handler);
         att.setXOP(false);
         atts.add(att);
-        return id;
+        return "cid:" + id;
```
reta, line 83: It looks like it, 

> 3. swaRef
[WS-I Attachment Profile 1.0](http://www.ws-i.org/Profiles/AttachmentsProfile-1.0-2004-08-24.html) defines mechanism to reference MIME attachment parts using [swaRef](http://www.ws-i.org/Profiles/AttachmentsProfile-1.0-2004-08-24.html#Referencing_Attachments_from_the_SOAP_Envelope). In this mechanism the content of XML element of type [wsi:swaRef](http://ws-i.org/profiles/basic/1.1/xsd/) is sent as MIME attachment and the element inside SOAP Body holds the reference to this attachment in the CID URI scheme as defined by [RFC 2111](http://www.faqs.org/rfcs/rfc2111.html).

PR #2386, opened 2025-04-29
CXF-9132 fiexd

CantosSong:
I don't think we should care about the http url&client here anymore

https://issues.apache.org/jira/projects/CXF/issues/CXF-9132?filter=allopenissues&orderby=priority+DESC%2C+updated+DESC


rt/transports/http/src/main/java/org/apache/cxf/transport/http/HttpClientHTTPConduit.java
```diff
@@ -349,13 +348,6 @@ protected void setupConnection(Message message, Address address, HTTPClientPolic
             return;
         }
 
-        if (sslURL != null && isSslTargetDifferent(sslURL, uri)) {
```
reta, line 352: @CantosSong thanks for the fix, but I lack some context here why this reset on HTTPS URL was even introduced. I would suggest we play safe here and introduce the contextual property that would do the reset if needed:

``
final boolean resetClientNeeded = MessageUtils.getContextualBoolean(message, RESET_CLIENT_ON_HTTPS_URL_CHANGE);
....
```

PR #2392, opened 2025-05-06
[CXF-9134] Do not wrap Attachment into another Attachment

andreibastun:
This supports interfaces generated by openapi jaxrs-cxf generator


reta:
Thank you @andreibastun !

PR #2395, opened 2025-05-08
CXF-9132 When the domain name changes, not execute 'release'

CantosSong:
When the domain name changes, 'acquire' is not executed and 'release' is executed, resulting in inconsistent count.


rt/transports/http/src/main/java/org/apache/cxf/transport/http/HttpClientHTTPConduit.java
```diff
@@ -352,7 +352,6 @@ protected void setupConnection(Message message, Address address, HTTPClientPolic
         if (sslURL != null && isSslTargetDifferent(sslURL, uri)) {
             sslURL = null;
             if (clientRef != null) {
-                clientRef.release();
```
reta, line 355: This is a client leak, not an option

reta (changes requested):

PR #2397, opened 2025-05-09
CXF-9132 fixed-one  remove unuse sslUrl

CantosSong:


CantosSong:
> @CantosSong we do have regression with this change, I've submitted a bit more elaborate fix here #2398, please take a look, thank you

OK.

PR #2398, opened 2025-05-10
CXF-9132: HttpClientHTTPConduit releases the client while there are connection(s) still using it

reta:
HttpClientHTTPConduit releases the client while there are connection(s) still using it


CantosSong:
@reta  I think there's no problem

CantosSong:
@reta I have a question that comes to mind. Domain switching is random, and will deferredClientRefs continue to increase, leading to memory leaks? Perhaps defining a Map using a Host would be a better choice

PR #2411, opened 2025-05-19
try to fix incompatibility with http/2 with many http server...

mcarbonneaux:
try to make a fix with http server only.
https://issues.apache.org/jira/browse/CXF-8896

### DIFF
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HttpClientHTTPConduit.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HttpClientHTTPConduit.java
@@ -425,6 +425,9 @@ public class HttpClientHTTPConduit extends URLConnectionHTTPConduit {
                 }
             }
             String verc = (String)message.getContextualProperty(FORCE_HTTP_VERSION);
+            if (verc == null && "http".equals(uri.getScheme())) {
+                verc="1.1";  
+            }
             if (verc == null) {
                 verc = csPolicy.getVersion();
             }


rt/transports/http/src/main/java/org/apache/cxf/transport/http/HttpClientHTTPConduit.java
```diff
@@ -425,6 +425,9 @@ protected void setupConnection(Message message, Address address, HTTPClientPolic
                 }
             }
             String verc = (String)message.getContextualProperty(FORCE_HTTP_VERSION);
+            if (verc == null && "http".equals(uri.getScheme())) {
```
reta, line 428: @mcarbonneaux the Apache CXF client conduit uses JDK `HttpClient` under the hood and relies on its  behavior with respect to preferred protocol selection - which is HTTP/2 by default (https://docs.oracle.com/en/java/javase/21/docs/api/java.net.http/java/net/http/HttpClient.html#newHttpClient())

mcarbonneaux: But in real life http/2 on http is not very safe to use by default ...
In all my program I enforce by default to http/1.1 for http usage...

mcarbonneaux: Is not juste openjdk problem is http/2 problem... At minimum more documenting in cfx documentation about FORCE_HTTP_VERSION because each time I need to explain to my developer why there code are not working with http....

reta: Sure, I will update documentation shortly to explain the HTTP protocol selection preferences, thank you

mcarbonneaux: but with using of http 1.1 by default on http backend you have no chance to have problem... is why i've proposed to enforce http 1.1 on http...

reta: @mcarbonneaux as I explained: a) we follow the HttpClient defaults b) if the defaults are not right for you - please use the property (FORCE_HTTP_VERSION )

reta: As I also mentioned, if you disagree with HttpClient  - please bring it to OpenJDK community or use another HTTP client conduit (Apache HttpClient 5 fe).

mcarbonneaux: httpclient leave you the posibility to choose... and cxf conduit can choose...

gnodet:
Closing — the maintainer feedback indicates this conflicts with CXF's design philosophy of following JDK HttpClient defaults. Users can use the FORCE_HTTP_VERSION property to force HTTP/1.1 for cleartext connections.

PR #2425, opened 2025-05-26
Masking fails when empty tag is repeated with content later on

bedaka:


reta:
> I opened a ticket here: https://issues.apache.org/jira/browse/CXF-9141

Thank you, @bedaka !

reta (approved):

PR #2427, opened 2025-05-28
CXF-9142 Change async tracing tests to be order-agnostic

cunningt:
https://issues.apache.org/jira/browse/CXF-9142

Change the async tracing tests so that the order of the responses no longer matters.


ffang (approved):

reta:
@cunningt @ffang my apologies for coming late. I will revert this commit folks although acknowledging we may have a problem / race with reporting. The fix I am suggesting to try out is to sort all spans by start / end - the order of spans is super important, asserting that span is just there is not sufficient (see please https://github.com/apache/cxf/commit/4a540e5e1afeab52a35a02c75979d47d850829a1)

ffang:
Thanks for the clarification and suggesting fix @reta !

PR #2431, opened 2025-05-31
CXF-9143: Remove Jakarta Mail Dependency

dnl50:
The Jakarta Mail API is only used in Test Sources and example projects, but is a compile time dependency of `cxf-rt-frontend-simple`. The dependencies on a Mail API Implementation (Apache Angus Mail) are replaced with the Mail API when no actual implementation is required.

### DIFF
--- a/distribution/src/main/release/samples/mtom/pom.xml
+++ b/distribution/src/main/release/samples/mtom/pom.xml
@@ -133,8 +133,8 @@
             <artifactId>angus-activation</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.eclipse.angus</groupId>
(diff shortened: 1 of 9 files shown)


dnl50:
After debugging through the failed tests, I found out that Angus Mail is required since CXF (or at least the failing test) depends on the `DataContentHandler`s which are implicitly registered by the [META-INF/mailcap file in the Angus Mail Jar](https://github.com/eclipse-ee4j/angus-mail/blob/2.0.3/core/src/main/resources/META-INF/mailcap). Since I am not familiar enough with the code of CXF to determine whether it is really needed at runtime, I prefer not to make any changes to the dependencies.

PR #2455, opened 2025-06-11
Fix some potential resource leaks

AROP-REPLICATION:
looks like some resources were not handled properly which may lead to resource leak, I used `try-with` stmt to close them safely

### DIFF
--- a/core/src/main/java/org/apache/cxf/configuration/jsse/TLSParameterJaxBUtils.java
+++ b/core/src/main/java/org/apache/cxf/configuration/jsse/TLSParameterJaxBUtils.java
@@ -160,8 +160,8 @@ public final class TLSParameterJaxBUtils {
                     LOG.severe(msg);
                     throw new IOException(msg);
                 }
+                keyStore.load(is, password);
             }
-            keyStore.load(is, password);
         } else if (kst.isSetUrl()) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/cxf/configuration/jsse/TLSParameterJaxBUtils.java
```diff
@@ -153,12 +153,13 @@ public static KeyStore getKeyStore(KeyStoreType kst, boolean trustStore)
...
-                throw new IOException(msg);
+            try (InputStream is = getResourceAsStream(kst.getResource())) {
+                if (is == null) {
+                    final String msg =
+                            "Could not load keystore resource " + kst.getResource();
+                    LOG.severe(msg);
+                    throw new IOException(msg);
+                }
             }
             keyStore.load(is, password);
```
reta: This is not correct, `keyStore.load(is, password);` would be operating over closed stream

reta (changes requested):

core/src/main/java/org/apache/cxf/configuration/jsse/TLSParameterJaxBUtils.java
AROP-REPLICATION: Thanks, I've moved it inside

PR #2461, opened 2025-06-13
CXF-9146 ThreadLocalClientState internally now uses a ThreadLocal

eric6iese:
…instead of a WeakHashMap when timeToKeepState. This fixes a memory leak which occurs when the Thread is referenced within the ClientState to which the Thread points in the WeakHashMap. This scenario happens often, something with the new HttpClientConduit which internally might use a PipedInputStream referencing said thrad.

I also added test to both check that the new implementation does not change the obversable behavior and a ruther blunt test which causes an OOME with the current implementation, but not with the new one. The tests are far from perfect because they use passive sleep statements instead of something like awaitility, but I did not want to write custom wait loops or integrate external libraries.

### DIFF
--- a/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/ThreadLocalClientState.java
+++ b/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/ThreadLocalClientState.java
@@ -31,13 +31,19 @@ import jakarta.ws.rs.core.UriBuilder;
 /**
  * Keeps the client state such as the baseURI, currentURI, requestHeaders, current response
  * in a thread local storage
- *
  */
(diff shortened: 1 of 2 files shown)


eric6iese:
Sadly, the solution does not work as expected: While it fixes the MemoryLeak with dead threads, it opens up a completely different memory leak when creating many short-lived instances from the same thread. 

This, and the fact that there are even more WeakHashMaps of Threads in the SOAP-Client convinces me that the fix should instead be within the HttpClientHTTPConduit by removing the PipedStreams as soon as they are closed.

eric6iese:
Closed, I will create a simpler PR with a more appropiate solution.

PR #2467, opened 2025-06-16
CXF-9146 Fix MemoryLeak in HttpClientHttpConduit

eric6iese:
If the wrappedStream corresponds to pout then it should be nulled as well. this fixes a memoryleak caused when pout is a PipedOutputStream, as the pipe holds a strong reference to the client thread which must be nulled to avoid MemoryLeaks in class like ThreadLocalClientState, which utilizes the same thread as a key for weakhashmap which in turn points to this very outputstream. this means that the thread is both the key and within the value of an entry of the weakhashamp, and as per the javadoc of said class, the GC cannot detect these cycles. to break this cycle, not only pout but also the wrappedStream needs to be nulled.

An alternative, simpler solution would be to always null the wrappedStream, but I did not debug the code enough to be sure if this is safe in all usecases like the retransmission. I am also not sure if it would also be necessary to null the cachedOutputStream, as it could point to the PipedOutputStream as well.

I've tried to keep the fix simpler this time, if you need additional tests, then let me now.

### DIFF
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HttpClientHTTPConduit.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HttpClientHTTPConduit.java
@@ -844,9 +844,6 @@ public class HttpClientHTTPConduit extends URLConnectionHTTPConduit {
                     } catch (IOException e) {
                         logStackTrace(e);
                     }
-                    if (wrappedStream == pout){
-                        wrappedStream = null;
-                    }
                     pout = null;
                 }
                 if (publisher != null) {
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HttpClientHTTPConduit.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/HttpClientHTTPConduit.java
@@ -844,6 +844,9 @@ public class HttpClientHTTPConduit extends URLConnectionHTTPConduit {
                     } catch (IOException e) {
                         logStackTrace(e);
                     }
+                    if (wrappedStream == pout){
+                        wrappedStream = null;
+                    }
                     pout = null;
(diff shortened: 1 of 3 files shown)


rt/transports/http/src/main/java/org/apache/cxf/transport/http/HttpClientHTTPConduit.java
```diff
@@ -844,6 +844,9 @@ public void close() throws IOException {
                     } catch (IOException e) {
                         logStackTrace(e);
                     }
+                    if (wrappedStream == pout){
```
reta: It is very likely that `AbstractWrappedOutputStream::close()` should be doing it (the call chain goes from `super.close()`), just set to `null` right when the stream is closed.

I am not sure why it has to be equal to `pout` to be reset to `null`?

eric6iese: The reason I put the check in there was to ensure that I do not clear a reference which might still be in use. But you are right, clearing the wrappedOutputStream after closing it should be safe and is much simpler. I did the same to the cachedOutputStream, since it could be holding a reference to the PipedOutputStream as well.

The only negative side effect these changes could have is that they trade an IOException (closed stream) for a NullpointerException if any of these streams is deferenced after they were closed.

reta: Thank  you

> The only negative side effect these changes could have is that they trade an IOException (closed stream) for a NullpointerException if any of these streams is deferenced after they were closed.

I think this should be non-issue, usage of the closed stream should never happen

reta (approved):

reta:
Thank you @eric6iese !

PR #2473, opened 2025-06-21
CXF-8828: Support Jakarta EE 11

reta:
CXF-8828: Support Jakarta EE 11


reta:
Ah, a few more certificates have expired ...

coheigea (approved):

PR #2539, opened 2025-08-05
CXF-9152: Adding @Timed annotation to JaxWS WebService class breaks Micrometer integration

reta:
Adding @Timed annotation to JaxWS WebService class breaks Micrometer integration


systests/spring-boot/src/test/java/org/apache/cxf/systest/jaxws/spring/boot/SpringJaxwsTimedTest.java
```diff
@@ -0,0 +1,271 @@
...
+
+        assertThat(timedTags)
+            .containsOnly(
+                entry("exception", "None"),
+                entry("faultCode", "None"),
+                entry("method", "POST"),
+                entry("operation", "sayHello"),
+                entry("uri", "/Service/" + HELLO_SERVICE_NAME_V1),
+                entry("outcome", "SUCCESS"),
+                entry("status", "200"));
```
knoobie, line 178: Thanks for the fix! 

General question, I'm not sure myself what would be best / most expexted: Should `sayHello.requests` contain all those "cxf related" tags or just the tags provided by the user e.g. with `@Timed(extraTags="...")`?

reta: > General question, I'm not sure myself what would be best / most expexted: Should sayHello.requests contain all those "cxf related" tags or just the tags provided by the user e.g. with @Timed(extraTags="...")?

I think having a union of tags is fine, those could be always filtered out (by scrapers fe) if not needed. Also, the CXF tags are indeed per service scope (applied to all metrics).

PR #2557, opened 2025-08-19
CXF-9158: Make CXF buildable offline with -Doffline

ppalaga:
https://issues.apache.org/jira/browse/CXF-9158

This PR does the following:

* Download the suffix file to rt/transports/http/src/main/resources/mozilla/public-suffix-list.txt instead of target
* Put the plugin into a new profile called `online` that is active by default unless the build is run with `-Doffline`
* Add a check to the CI script to fail if there are uncommitted changes in rt/transports/http/src/main/resources/mozilla/public-suffix-list.txt so that we are notified if the list is changed and so that we are forced to have all the changes in git for the offline build.
* This also helps to make this aspect of the build reproducible (when building with -Doffline).


parent/pom.xml
```diff
@@ -585,7 +586,19 @@
                         <dependency>
                             <groupId>org.apache.groovy</groupId>
                             <artifactId>groovy</artifactId>
-                            <version>4.0.27</version>
+                            <version>${cxf.groovy.version}</version>
+                        </dependency>
+                    </dependencies>
+                </plugin>
+                <plugin>
+                    <groupId>org.codehaus.gmaven</groupId>
```
reta, line 594: @ppalaga why do we need `groovy-maven-plugin` plugin? AFAIK it is not supported and will be problematic (potentially) for Maven 4 support.

ppalaga: > why do we need groovy-maven-plugin plugin?

The first solution I could figure out that does not change the timestamp on the local file when there is no content change. Those timestamps might be important for incremental builds. 
Ant plugin could perhaps be used to achieve the same with a bit more code - should I try?

> AFAIK it is not supported

In which sense not supported?

reta: > In which sense not supported?

It was not updated since 2019 (https://groovy.github.io/gmaven/groovy-maven-plugin/) , probably better wording would be "not maintained".

ppalaga: I see. gmavenplus has some recent releases: https://mvnrepository.com/artifact/org.codehaus.gmavenplus/gmavenplus-plugin
Would that one be OK?

ppalaga: Ouch, I just found out, how many changes the suffix file is receiving https://github.com/publicsuffix/list/commits/main/public_suffix_list.dat
It is nearly every day.
I am afraid, the approach proposed in this PR would not be sustainable.

It would actually be enough to update the list shortly before releasing CXF. @ffang I wonder whether there is any release automation in place for CXF? Is there a CI job or a script? If so, the suffix list update could be added there.

ffang: Hi @ppalaga ,

No such automation before release where we can put the [public_suffix_list.dat](https://github.com/publicsuffix/list/commits/main/public_suffix_list.dat) update in.

But as we need to do manual steps before each release anyway, like update release notes in distribution/src/main/release/release_notes.txt, so we can also update the offline public_suffix_list.dat before each release.

Cheers
Freeman

ppalaga: Thanks for the explanation, @ffang! Is there at least some document somewhere listing the release steps where I could add it?

ffang: Hi @ppalaga ,

We have wiki based page here
https://cxf.apache.org/release-management.html
This should be the page we add that step. When we edit the wiki page, an apache buildbot will sync the change to apache infra svn here
https://svn.apache.org/repos/infra/websites/production/cxf/content/release-management.html

I don't think you have the permission to edit the wiki directly. The simplest way is that you send an old days svn based patch(not a github pull request) to CXF-9158 so we can change the website accordingly once this PR get in.

Best Regards
Freeman

gnodet:
Closing — as @ppalaga noted in the review thread, the public suffix list changes nearly daily, making the approach of committing it to the repo unsustainable. The alternative of updating it as a pre-release step was discussed but not pursued. This PR also has merge conflicts. A different approach would be needed to address CXF-9158.

PR #2566, opened 2025-08-22
[CXF-9160] Include JBoss remoting protocols in the whitelist of JNDI protocols

ropalka:
https://issues.apache.org/jira/browse/CXF-9160

### DIFF
--- a/rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/util/JndiHelper.java
+++ b/rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/util/JndiHelper.java
@@ -30,7 +30,9 @@ import javax.naming.NamingException;
 public class JndiHelper {
 
     private static final List<String> ALLOWED_PROTOCOLS = Arrays.asList(
-        "vm://", "tcp://", "nio://", "ssl://", "http://", "https://", "ws://", "wss://");
+        "vm://", "tcp://", "nio://", "ssl://", "http://", "https://", "ws://", "wss://",
+        "remote://", "remote+tls://", "remote+http://", "remote+https://" // JBoss Remoting protocols
+    );
     private Properties environment;
 
     /**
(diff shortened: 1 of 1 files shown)


rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/util/JndiHelper.java
```diff
@@ -30,7 +30,9 @@
 public class JndiHelper {
 
     private static final List<String> ALLOWED_PROTOCOLS = Arrays.asList(
-        "vm://", "tcp://", "nio://", "ssl://", "http://", "https://", "ws://", "wss://");
+        "vm://", "tcp://", "nio://", "ssl://", "http://", "https://", "ws://", "wss://",
```
reta: Thank you @ropalka , @coheigea any concerns with relaxing the defaults? I think alternatively we could make the allowed protocols configurable and keep defaults intact.

coheigea: @ropalka I think as these are unlikely to be relevant to most CXF users, it makes more sense to make the default configurable so that they can be added by JBoss?

rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/util/JndiHelper.java
```diff
@@ -18,21 +18,48 @@
...
-    private static final List<String> ALLOWED_PROTOCOLS = Arrays.asList(
-        "vm://", "tcp://", "nio://", "ssl://", "http://", "https://", "ws://", "wss://");
+    /**
+     * JVM/System property name holding allowed jms protocols.
+     */
+    private static final String CONFIGURED_JMS_PROTOCOLS = "jms.protocols";
+    /**
+     * JVM/System property name holding default allowed jms protocols.
+     */
+    private static final String DEFAULT_JMS_PROTOCOLS = "vm,tcp,nio,ssl,http,https,ws,wss";
```
asoldano, line 45: The default protocol list in the new system property ("vm,tcp,...") omits the :// suffix that was present in the original hardcoded list ("vm://", "tcp://", ...). Perhaps it's not a major issue, but this would make the check less strict/secure.

asoldano (changes requested):

rt/transports/jms/src/main/java/org/apache/cxf/transport/jms/util/JndiHelper.java
ropalka: Fixed @asoldano

PR #2575, opened 2025-08-29
CXF-9007: NullPointerException in XMLStreamDataWriter.writeNode

radulacatusu:
The same fix was applied here: https://github.com/apache/cxf/pull/1861

### DIFF
--- a/core/src/main/java/org/apache/cxf/helpers/DOMUtils.java
+++ b/core/src/main/java/org/apache/cxf/helpers/DOMUtils.java
@@ -790,7 +790,10 @@ public final class DOMUtils {
             //java9 plus hack
             Field f = GET_DOCUMENT_FRAGMENT_FIELDS.get(fragment.getClass());
             if (f != null) {
-                return ReflectionUtil.accessDeclaredField(f, fragment, DocumentFragment.class);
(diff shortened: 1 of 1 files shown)


reta:
@radulacatusu the 3.5.x is not supported / maintained, there won't be any releases cut

radulacatusu:
@reta Thank you for your reply. Our project is using 3.5.11 and we wanted to add this fix and expected a 3.5.12. What version do you recommend to use, with no major updates but that would contain [CXF-9118](https://issues.apache.org/jira/browse/CXF-9118) and [CXF-9007](https://issues.apache.org/jira/browse/CXF-9007)?

reta:
> @reta Thank you for your reply. Our project is using 3.5.11 and we wanted to add this fix and expected a 3.5.12. What version do you recommend to use, with no major updates but that would contain [CXF-9118](https://issues.apache.org/jira/browse/CXF-9118) and [CXF-9007](https://issues.apache.org/jira/browse/CXF-9007)?

Thanks @radulacatusu , please use latest 3.6.x release, [CXF-9007](https://issues.apache.org/jira/browse/CXF-9007) seemed to slip and not backported , I am taking care of it right away. Thank you.

radulacatusu:
@reta thank you. I will use the next version that you will generate for 3.6.x.

PR #2582, opened 2025-09-04
[CXF-9165] Removing unnecessary JDK8 hacks

ropalka:
https://issues.apache.org/jira/browse/CXF-9165

### DIFF
--- a/core/src/main/java/org/apache/cxf/common/logging/JDKBugHacks.java
+++ b/core/src/main/java/org/apache/cxf/common/logging/JDKBugHacks.java
@@ -20,17 +20,14 @@
 package org.apache.cxf.common.logging;
 
 import java.io.BufferedReader;
-import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
-import java.lang.reflect.Method;
-import java.net.URL;
 import java.net.URLConnection;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
+import java.security.SecureRandom;
 
 import javax.imageio.ImageIO;
-import javax.xml.parsers.DocumentBuilderFactory;
 
 import org.apache.cxf.common.classloader.ClassLoaderUtils;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/cxf/common/logging/JDKBugHacks.java
```diff
@@ -99,91 +96,61 @@ public static void doHacks() {
...
-                        Method method = policyClass.getMethod("getPolicy");
-                        method.invoke(null);
-                    }
-                } catch (Throwable e) {
-                    // ignore
-                }
-                try {
-                    // Initializing javax.security.auth.login.Configuration retains a static reference
-                    // to the context class loader.
-                    if (!skipHack("org.apache.cxf.JDKBugHacks.authConfiguration")) {
```
reta: `javax.security.auth.login.Configuration` seems to be present in JDK-21, could you please keep it?

core/src/main/java/org/apache/cxf/common/logging/JDKBugHacks.java
```diff
@@ -99,91 +96,61 @@ public static void doHacks() {
...
+                        URLConnection.setDefaultUseCaches("JAR", false);
                     }
-                } catch (Throwable e) {
-                    // Ignore
-                }
-                // Creating a MessageDigest during web application startup
-                // initializes the Java Cryptography Architecture. Under certain
-                // conditions this starts a Token poller thread with TCCL equal
-                // to the web application class loader.
-                if (!skipHack("org.apache.cxf.JDKBugHacks.securityProviders")) {
```
reta: This one is also seems to be in effect (`java.security.Security.getProviders()`), right? Could we keep it?

reta:
Thanks @ropalka , I understand that Tomcat removed some of them, but CXF is not bound to Tomcat, I am unsure of any other unintended consequences of removing the existing hacks. If those are noop - we could keep them, derisking possible regressions.

ropalka:
OK I restored them @reta

PR #2588, opened 2025-09-09
Fix expired certificates

reta:
Fix expired certificates


ffang:
Thanks @reta for fixing this up quickly!

Freeman

PR #2607, opened 2025-09-16
[CXF-9159]CXF client to support spring boot SSL bundles

ffang:
The CXF client can wire Spring Boot SSL bundle configuration automatically if Spring Boot SSL bundle available in the configuration and if cxf.client.ssl.enabled property is true(default is false so keep the CXF client behaviour as is).  We can select the bundle name per application or per server address match.

### DIFF
--- a/integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/ssl/CxfClientSslProperties.java
+++ b/integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/ssl/CxfClientSslProperties.java
@@ -117,11 +117,11 @@ public class CxfClientSslProperties {
         this.disableCnCheck = disableCnCheck;
     }
 
-    public List<CxfClientSslBundle> getCxfClientSslBundle() {
+    public List<CxfClientSslBundle> getCxfClientSslBundles() {
         return cxfClientSslBundles;
     }
 
-    public void setCxfClientSslBundle(List<CxfClientSslBundle> clientSslBundles) {
+    public void setCxfClientSslBundles(List<CxfClientSslBundle> clientSslBundles) {
         this.cxfClientSslBundles = clientSslBundles;
     }
 }
--- a/integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/ssl/CxfClientSslProperties.java
+++ b/integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/ssl/CxfClientSslProperties.java
@@ -31,7 +31,7 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 public class CxfClientSslProperties {
     private boolean enabled;
     //Name of the Spring SSL bundle to use for outbound CXF clients. */
-    private String bundle = "cxf-client";
+    private String defaultBundle = "cxf-client";
     // Convenience flag for tests
     private Boolean disableCnCheck = Boolean.FALSE;
     private List<CxfClientSslBundle> cxfClientSslBundles = new ArrayList<>();
@@ -101,12 +101,12 @@ public class CxfClientSslProperties {
         this.enabled = enabled;
     }
 
-    public String getBundle() {
-        return bundle;
+    public String getDefaultBundle() {
+        return defaultBundle;
     }
 
-    public void setBundle(String bundle) {
-        this.bundle = bundle;
+    public void setDefaultBundle(String bundle) {
+        this.defaultBundle = bundle;
     }
     
     public Boolean getDisableCnCheck() {
--- a/integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/ssl/CxfClientSslProperties.java
+++ b/integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/ssl/CxfClientSslProperties.java
@@ -1,0 +1,109 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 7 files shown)


integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/ssl/CxfClientSslProperties.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.List;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Client-side SSL integration properties for CXF when Spring Boot SslBundles are present.
+ * bundles are applied once when an HTTPConduit is configured.
+ */
+@ConfigurationProperties("cxf.client.ssl")
+public class CxfClientSslProperties {
```
reta, line 31: Not particularly better option, but may be we could mirror SSL bundle with list of named `CxfClientSslBundle` instances? Fe:

```
@ConfigurationProperties("cxf.client.ssl")
public class CxfClientSslProperties {
    List<CxfClientSslBundle> bundles;
    private boolean enabled;

    class CxfClientSslBundle {
        private String name; <--- same as the name in `HTTPConduitConfigurer`
        private Boolean disableCnCheck = Boolean.FALSE;
        private String bundle;
        private String protocol;
        private List<String> cipherSuites;
    }
}
```

To me the combination of props + rules does make it confusing somewhat

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/ssl/CxfClientSslProperties.java
```diff
@@ -0,0 +1,127 @@
...
+    
+    public Boolean getDisableCnCheck() {
+        return disableCnCheck;
+    }
+
+    public void setDisableCnCheck(Boolean disableCnCheck) {
+        this.disableCnCheck = disableCnCheck;
+    }
+
+    public List<CxfClientSslBundle> getCxfClientSslBundle() {
```
reta: Very minor:

```suggestion
    public List<CxfClientSslBundle> getCxfClientSslBundles() {
```

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/ssl/CxfClientSslProperties.java
```diff
@@ -0,0 +1,127 @@
...
+        return cxfClientSslBundles;
+    }
+
+    public void setCxfClientSslBundle(List<CxfClientSslBundle> clientSslBundles) {
```
reta:
```suggestion
    public void setCxfClientSslBundles(List<CxfClientSslBundle> clientSslBundles) {
```

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/ssl/CxfClientSslProperties.java
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Client-side SSL integration properties for CXF when Spring Boot SslBundles are present.
+ * bundles are applied once when an HTTPConduit is configured.
+ */
+@ConfigurationProperties("cxf.client.ssl")
+public class CxfClientSslProperties {
+    private boolean enabled;
+    //Name of the Spring SSL bundle to use for outbound CXF clients. */
+    private String bundle = "cxf-client";
```
reta: Do we need this property or `CxfClientSslBundle::bundle` is sufficient? or the idea is to have a default one for clients?

ffang: Hi @reta ,

Yes, the idea is to have default one for clients, so that in most cases, the List<CxfClientSslBundle> isn't necessary or could be too much configuration, so
```
cxf:
    client:
    ssl:
      enabled: true
      bundle: cxf-client
```
is sufficient.

Freeman

reta: In your opinion, do you think it is better to call it `defaultBundle` (disambiguate with `CxfClientSslBundle::bundle`)?

ffang: Yes, revised accordingly

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/ssl/CxfClientSslProperties.java
```diff
@@ -0,0 +1,127 @@
...
+
+        public String getAddress() {
+            return address;
+        }
+
+        public void setAddress(String address) {
+            this.address = address;
+        }
+
+        public String getBundle() {
```
reta, line 55: Also getters / setters probably:

```suggestion
        public String getDefaultBundle() {
```

integration/spring-boot/autoconfigure/src/main/java/org/apache/cxf/spring/boot/autoconfigure/ssl/CxfClientSslProperties.java
```diff
@@ -0,0 +1,127 @@
...
+            return bundle;
+        }
+
+        public void setBundle(String bundle) {
```
reta, line 59:
```suggestion
        public void setDefaultBundle(String bundle) {
```

reta:
LGTM, thanks a lot @ffang !

ffang:
Thanks @reta !

PR #2948, opened 2026-03-11
Fix RFC 9068 compliance: set typ header to at+jwt for JWT access tokens

gnodet:
## Summary

CXF supports JWT-formatted access tokens via `useJwtFormatForAccessTokens`, but the produced tokens are missing the `typ: at+jwt` header required by [RFC 9068 Section 2.1](https://datatracker.ietf.org/doc/html/rfc9068#section-2.1):

> The JWT MUST contain a typ header parameter with the value at+jwt

Without this header, any RFC 9068-compliant resource server would reject CXF-issued JWT access tokens. This is a spec compliance bug.

### Changes

- **`JoseConstants`**: Added `TYPE_AT_JWT = "at+jwt"` constant
- **`JoseType`**: Added `AT_JWT` enum member with proper `getType()` lookup (since `at+jwt` can't be a Java identifier, it needs explicit handling like `JOSE_JSON`)
- **`AbstractOAuthDataProvider.processJwtAccessToken()`**: Sets the `at+jwt` type on **both** JWS and JWE headers, handling both the sign-only and sign+encrypt cases per RFC 9068 Section 2.1 and Section 4

This is a proper implementation incorporating @reta's review feedback from #990 (using the type system properly and handling both JWS and JWE).

Closes #990

## Test plan
- [x] `JwsCompactReaderWriterTest` — passes
- [x] `OAuth2JwtFiltersTest` — 5 tests pass
- [x] `JAXRSOAuth2TlsTest` — 5 tests pass

### DIFF
--- a/rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/common/JoseConstants.java
+++ b/rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/common/JoseConstants.java
@@ -43,6 +43,7 @@ public final class JoseConstants extends RSSecurityConstants {
     public static final String JWS_HEADER_B64_STATUS_HEADER = "b64";
 
     public static final String TYPE_JWT = "JWT";
+    public static final String TYPE_AT_JWT = "at+jwt";
     public static final String TYPE_JOSE = "JOSE";
     public static final String TYPE_JOSE_JSON = "JOSE+JSON";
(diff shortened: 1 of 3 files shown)


reta:
@coheigea could you please help me to validate this pull request? I will also take a look, thank you

coheigea (approved):
@reta It looks OK to me, I verified manually the type is added to the headers in the systests.

PR #2953, opened 2026-03-12
Fix ModelEncryptionSupport serialization/deserialization bugs

gnodet:
## Summary

Fixes two bugs in `ModelEncryptionSupport` that made `DefaultEncryptingOAuthDataProvider` practically unusable:

1. **Null values serialized as spaces**: `tokenizeString(null)` returned `" "` (space), so `new Client("clientId", null, false)` would deserialize with `clientSecret = " "` instead of `null`
2. **Collections serialized with brackets and spaces**: Using `List.toString()` produced `[openid, email]`, and `parseSimpleList` would strip the brackets but leave spaces, resulting in `" email"` instead of `"email"`

### Changes

- `tokenizeString`: returns `""` instead of `" "` for null
- `getStringPart`: checks for `""` instead of `" "` 
- New `tokenizeCollection`/`tokenizeMap` helpers replace direct `.toString()` calls
- `parseSimpleList`/`parseSimpleMap`: simplified to work with comma-separated values without brackets
- All `split()` calls use `-1` limit to preserve trailing empty strings
- `parseSimpleMap`: added empty check to avoid NPE on empty strings
- Applied `getStringPart` to client id/secret in `recreateClientInternal`

Supersedes #816. A JIRA ticket should be created for this fix.

## Test plan

- [x] Added `testPublicClient` test — verifies null secret round-trips correctly
- [x] All existing `CryptoUtilsTest` tests pass
- [ ] CI verification

### DIFF
--- a/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/utils/crypto/ModelEncryptionSupport.java
+++ b/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/utils/crypto/ModelEncryptionSupport.java
@@ -21,11 +21,14 @@ package org.apache.cxf.rs.security.oauth2.utils.crypto;
 
 import java.security.Key;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


gnodet:
CI failures are unrelated flaky OAuth2 systest failures (AuthorizationGrantTest, PublicClientTest — port/server startup issues). Re-triggering.

PR #2954, opened 2026-03-12
Reduce equal method candidates log from WARNING to FINE

gnodet:
## Summary

- Downgrades the "Both X and Y are equal candidates for handling the current request" log message from `WARNING` to `FINE` (debug) level in `OperationResourceInfoComparatorBase`
- This message fires during normal CORS preflight handling where `CrossOriginResourceSharingFilter.findPreflightMethod()` uses `MediaType.WILDCARD`, causing all methods on the same path to match equally
- The behavior is deterministic (stable sort order) so the warning is misleading — it says "unpredictable results" but the results are actually predictable
- Users debugging ambiguous routing can still see this message by enabling FINE logging

Supersedes #862 with a simpler fix — rather than changing the CORS filter to pass real media types, we simply reduce the noise level of a message that fires in expected scenarios.

## Test plan

- [x] Compiles cleanly
- [ ] CI verification


ffang (approved):
Thanks @gnodet !

PR #607, opened 2019-12-03
cxf-core: cleanup ConfigurerImpl

amarkevich:

PR #2956, opened 2026-03-12
Clean up ConfigurerImpl: replace MatcherHolder with immutable PatternHolder

gnodet:
## Summary
- Replaces `MatcherHolder` (mutable `Matcher` with `synchronized` reuse) with `PatternHolder` (immutable `Pattern`, creates new `Matcher` per call) — simpler and inherently thread-safe without explicit synchronization
- Replaces `TreeMap` with `HashMap` — the `Comparable` implementation on `MatcherHolder` was dead code (lists were appended to, never sorted)
- Uses `computeIfAbsent` for cleaner wildcard map population
- Makes `isWildcardBeanName` static

Supersedes #607 (which also targeted the now-removed Blueprint `ConfigurerImpl`).

## Test plan
- [x] `ConfigurerImplTest` passes
- [ ] CI pipeline

### DIFF
--- a/core/src/main/java/org/apache/cxf/configuration/spring/ConfigurerImpl.java
+++ b/core/src/main/java/org/apache/cxf/configuration/spring/ConfigurerImpl.java
@@ -21,15 +21,14 @@ package org.apache.cxf.configuration.spring;
 
 import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.TreeMap;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


gnodet:
Looking at this more carefully, the change from `TreeMap` to `HashMap` is correct and the `Comparable` on `MatcherHolder` was effectively dead code. Here's why:

The `TreeMap<String, List<MatcherHolder>>` is keyed by **class name** — the `TreeMap` sorts by that key (alphabetical class name order), which has no semantic meaning for the matching logic.

The `MatcherHolder.compareTo()` method was designed to prioritize patterns with more literal characters, but it was never actually used: the holders are stored in an `ArrayList<MatcherHolder>`, which doesn't use `Comparable` to maintain order. Items are always matched in insertion order regardless.

So both the `TreeMap` ordering and `MatcherHolder.compareTo()` were dead code — the `HashMap` + `PatternHolder` replacement is a correct simplification.

The other changes (reusing `Pattern` instead of `Matcher`, dropping `synchronized`, using `computeIfAbsent`) are also improvements — `Pattern` is thread-safe and immutable, while `Matcher` is not, which is why the old code needed `synchronized (m.matcher)`.

PR #2963, opened 2026-03-12
Fix ExecutorService leak in WS-RM oneway tests

gnodet:
## Summary
- `DeliveryAssuranceOnewayTest` and `MessageCallbackOnewayTest` create `ExecutorService` instances via `Executors.newSingleThreadExecutor()` in their `*AsyncExecutor` test methods but never shut them down
- The leaked non-daemon threads prevent clean test completion, causing intermittent timeouts in CI (~95-100s for DeliveryAssurance, ~75s for MessageCallback)
- Fix: track the `ExecutorService` in an instance field and shut it down in `tearDown()` with a 10-second grace period before forceful termination

## Test plan
- [ ] CI passes without WS-RM oneway test timeouts
- [ ] Both `DeliveryAssuranceOnewayTest` and `MessageCallbackOnewayTest` complete within their 60s timeout


coheigea:
I tried this locally and it doesn't seem to solve the flakiness of the tests

coheigea (approved):
It seems to help somewhat, even though it doesn't fix all the issues, so let's just merge.

PR #2966, opened 2026-03-12
Fix flaky Grizzly EndpointAPITest: use graceful server shutdown

gnodet:
## Summary

- Replace `shutdownNow()` with graceful `shutdown().get(10, TimeUnit.SECONDS)` in `EndpointAPITest.tearDown()`
- This ensures the Grizzly server fully releases ports and resources before the next test's `setUp()` tries to bind a new port
- Falls back to `shutdownNow()` if graceful shutdown times out

## Test plan

- [ ] CI passes: Grizzly EndpointAPITest should be more stable
- [ ] No other tests affected


coheigea:
@gnodet @reta is this causing build problems? I don't see any problems running this test repeatedly on my machine

reta:
> @gnodet @reta is this causing build problems? I don't see any problems running this test repeatedly on my machine

Same, I have never seen an issue with those tests

PR #2972, opened 2026-03-13
Fix AttachmentMtomChunkingTest port collision with AttachmentChunkingTest

gnodet:
## Summary

- Both `AttachmentChunkingTest` and `AttachmentMtomChunkingTest` have inner classes named `DownloadServer`
- Both use `allocatePort(DownloadServer.class)` which resolves to the same port via `TestUtil` (stores under the simple class name `DownloadServer`)
- When both tests run in the same JVM, the second test fails with `ListenerRegistrationException: Soap 1.1 endpoint already registered on address`
- Fix: use `allocatePort(AttachmentMtomChunkingTest.class)` to get a distinct port

## Test plan

- [ ] Both `AttachmentChunkingTest` and `AttachmentMtomChunkingTest` pass in the same surefire fork

### DIFF
--- a/systests/jaxws/src/test/java/org/apache/cxf/systest/jaxws/AttachmentMtomChunkingTest.java
+++ b/systests/jaxws/src/test/java/org/apache/cxf/systest/jaxws/AttachmentMtomChunkingTest.java
@@ -44,7 +44,7 @@ import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
 
 public class AttachmentMtomChunkingTest extends AbstractAttachmentChunkingTest {
-    private static final String PORT = allocatePort(DownloadServer.class);
+    private static final String PORT = allocatePort(AttachmentMtomChunkingTest.class);
     private static final Logger LOG = LogUtils.getLogger(AttachmentMtomChunkingTest.class);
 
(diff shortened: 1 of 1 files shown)


systests/jaxws/src/test/java/org/apache/cxf/systest/jaxws/AttachmentMtomChunkingTest.java
```diff
@@ -44,7 +44,7 @@
 import static org.junit.Assert.assertTrue;
 
 public class AttachmentMtomChunkingTest extends AbstractAttachmentChunkingTest {
-    private static final String PORT = allocatePort(DownloadServer.class);
+    private static final String PORT = allocatePort(AttachmentMtomChunkingTest.class);
```
reta, line 47: This is great finding @gnodet , however I think we would have to address the cause:
 - `allocatePort()` sets two properties, with full class name, and simple class name
 - the conflict is apparently over simple class name

It would be beneficial to improve `TestUtil` to catch those conflicts (it can do it, properties are static), so the test suite fails to run and would need to be fixed. Wdyt?

reta:
Should be fixed by https://github.com/apache/cxf/pull/2973

PR #3026, opened 2026-04-10
CXF-9210 Read JAX-RS requests with CacheControlFeature enabled and no Cache-Control header

gopackgo90:
Fixes https://issues.apache.org/jira/browse/CXF-9210


ffang (approved):
LGTM, thanks for the PR @gopackgo90 !

PR #3082, opened 2026-05-07
Update Jetty to 12.1.10

reta:
Update Jetty to 12.1.10


reta:
Holding off with update, regression on Windows confirmed and observed as well: https://github.com/jetty/jetty.project/issues/15021

PR #3091, opened 2026-05-08
Restrict valid URL protocols in TLSParameterJaxBUtils and URIResolver

coheigea:


core/src/main/java/org/apache/cxf/resource/URIResolver.java
```diff
@@ -56,6 +61,10 @@
  */
 public class URIResolver implements AutoCloseable {
     private static final Logger LOG = LogUtils.getLogger(URIResolver.class);
+    private static final String ALLOWED_URL_SCHEMES_PROPERTY =
+        "org.apache.cxf.resource.uriresolver.allowedSchemes";
+    private static final Set<String> DEFAULT_ALLOWED_URL_SCHEMES =
+        Collections.unmodifiableSet(new HashSet<>(Arrays.asList("file", "http", "https", "jar", "zip", "wsjar")));
```
reta: We also probably need "local" for LocalTransport?

```suggestion
        Collections.unmodifiableSet(new HashSet<>(Arrays.asList("file", "http", "https", "jar", "zip", "wsjar", "local")));
```

On the similar note, we added AbstractWrapperWSDLLocator::ALLOWED_SCHEMES recently for very similar purpose, do you think we could consolidate those or it does not make much sense? 

Thanks!

reta (approved):

core/src/main/java/org/apache/cxf/resource/URIResolver.java
coheigea: Thanks @reta , I added "local" and consolidated the older code in AbstractWrapperWSDLLocator

PR #3092, opened 2026-05-08
Remove SQLPrinterVisitor

coheigea:


rt/rs/extensions/search/src/main/java/org/apache/cxf/jaxrs/ext/search/sql/SQLPrinterVisitor.java
```diff
@@ -32,6 +33,10 @@
 
 public class SQLPrinterVisitor<T> extends AbstractUntypedSearchConditionVisitor<T, String> {
 
+    private static final Pattern SQL_IDENTIFIER_PATTERN = Pattern.compile("[A-Za-z_][A-Za-z0-9_]*");
```
reta: I am wondering if that is exhaustive / accurate? Fe Postgres let you use quotes in name: `SELECT * FROM users AS "users_alias";`, may be we shouldn't fail hard on those? Or there is a particular issue you run into?

coheigea:
@reta On review I doubt anyone is using this code and I propose just to remove it for 4.2 + 4.1 as it has security issues, and the regex fixes are messy.

reta:
> @reta On review I doubt anyone is using this code and I propose just to remove it for 4.2 + 4.1 as it has security issues, and the regex fixes are messy.

@coheigea +1 to remove it, could we mark it as `@Deprecated(forRemoval = true)` and remove in 4.2? (since 4.1.x is already out and change may hit by surprise) Thanks!

coheigea:
Thanks @reta that makes sense!

PR #3093, opened 2026-05-08
[CXF-9147]To accommodate the behaviour that several schema namespaces…

ffang:
… are explicitly mapped to the same Java package

### DIFF
--- a/tools/wsdlto/databinding/jaxb/src/main/java/org/apache/cxf/tools/wsdlto/databinding/jaxb/JAXBDataBinding.java
+++ b/tools/wsdlto/databinding/jaxb/src/main/java/org/apache/cxf/tools/wsdlto/databinding/jaxb/JAXBDataBinding.java
@@ -72,6 +72,7 @@ import org.xml.sax.helpers.XMLFilterImpl;
 
 import com.sun.codemodel.ClassType;
 import com.sun.codemodel.CodeWriter;
+import com.sun.codemodel.JAnnotationUse;
 import com.sun.codemodel.JClass;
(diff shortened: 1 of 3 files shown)


reta (commented):
Thanks @ffang

tools/wsdlto/databinding/jaxb/src/main/java/org/apache/cxf/tools/wsdlto/databinding/jaxb/JAXBDataBinding.java
```diff
@@ -851,6 +855,90 @@ public void generate(ToolContext c) throws ToolException {
...
+        }
+
+        for (Map.Entry<String, String> entry : classToNamespace.entrySet()) {
+            JDefinedClass cls = codeModel._getClass(entry.getKey());
+            if (cls == null) {
+                continue;
+            }
+            JAnnotationUse xmlType = getAnnotation(cls, "jakarta.xml.bind.annotation.XmlType");
+            if (xmlType == null) {
+                xmlType = getAnnotation(cls, "javax.xml.bind.annotation.XmlType");
```
reta: I think we should not see `"javax.* within 4.x release line, or I am mistaken?

ffang: Hi @reta ,

You are right, just a fallback which never will be reached for CXF 4.x, I will remove it.

Thanks!
Freeman

reta (approved):

PR #3124, opened 2026-05-19
CXF-9214 WSDLs and XSDs cannot be loaded from class path on GraalVM

ppalaga:
https://issues.apache.org/jira/browse/CXF-9214

### DIFF
--- a/core/src/main/java/org/apache/cxf/resource/URIResolver.java
+++ b/core/src/main/java/org/apache/cxf/resource/URIResolver.java
@@ -65,7 +65,8 @@ public class URIResolver implements AutoCloseable {
         "org.apache.cxf.resource.uriresolver.allowedSchemes";
     private static final Set<String> DEFAULT_ALLOWED_URL_SCHEMES =
         Collections.unmodifiableSet(
-            new HashSet<>(Arrays.asList("file", "http", "https", "jar", "zip", "wsjar", "local", "classpath", "vfs")));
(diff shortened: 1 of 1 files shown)


reta (approved):

ffang (approved):

ffang:
Hi @coheigea ,

Could you please take a look if this is also good for you?
And do you think we should re-cut the release to include this fix?

Thanks!
Freeman

ppalaga:
.oO( There is so many nicer things to do in life then figure out why checkstyle failed. There is a few good code formatters for Java, that do not require any attention. )

ppalaga:
A workaround in Camel Quarkus is possible https://github.com/quarkiverse/quarkus-cxf/pull/2144 
I do not think the CXF release train needs to get stopped.

coheigea (approved):

coheigea:
@ffang As we have a workaround for Quarkus, I'd say let's just go ahead with the current votes. We'll likely be releasing quite often over the next while with all the AI related findings.

ppalaga:
Should I send separate PRs to maintenance branches (which)?

coheigea:
No thanks, I already backported to 4.1.x-fixes https://github.com/apache/cxf/commit/760945f99ee5d5e1011c52619c08d9fdd10ec313

ffang:
> @ffang As we have a workaround for Quarkus, I'd say let's just go ahead with the current votes. We'll likely be releasing quite often over the next while with all the AI related findings.

Thanks @coheigea !

PR #3128, opened 2026-05-20
Removing OAuthServiceExceptionMapper which isn't used anywhere

coheigea:
@reta WDYT about backporting this to 4.1.x as well? I doubt anyone is using it, as it's not referenced anywhere in the codebase


reta:
> @reta WDYT about backporting this to 4.1.x as well? I doubt anyone is using it, as it's not referenced anywhere in the codebase

@coheigea I am somewhat hesitant, this is public class that may be used by applications / services, even if we are not aware of it, deprecation for removal would be a good way to approach it. Thank you

reta (approved):

coheigea:
> > @reta WDYT about backporting this to 4.1.x as well? I doubt anyone is using it, as it's not referenced anywhere in the codebase
> 
> @coheigea I am somewhat hesitant, this is public class that may be used by applications / services, even if we are not aware of it, deprecation for removal would be a good way to approach it. Thank you

Thanks @reta, I'll merge it for main only and mark the class as deprecated for removal on 4.1.x

PR #3132, opened 2026-05-21
CXF-9216 - Switch default OAuth2 code verifier to Digest

coheigea:


rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/grants/code/AuthorizationCodeGrantHandler.java
```diff
@@ -173,9 +173,9 @@ private boolean compareCodeVerifierWithChallenge(Client c, String clientCodeVeri
                     return false;
                 }
             }
-            // Fall back to plain
+            // Fall back to digest
             if (codeVerifierTransformer == null) {
-                codeVerifierTransformer = new PlainCodeVerifier();
+                codeVerifierTransformer = new DigestCodeVerifier();
```
reta: Thanks @coheigea , two questions please: 
 - do you think we could make this default configurable? (fe on a class level field, not a system property)
 - seems like this is the only place where `PlainCodeVerifier` is used, if we won't provide the option to supply default, it could be probably removed

WDYT? Thanks!

coheigea: @reta I've made the default code verifier configurable on the class. PlainCodeVerifier can still be plugged in to the class, so I think we can keep it for now.

reta (approved):

PR #3133, opened 2026-05-21
Synchronize access token refresh when refresh tokens aren't recycled

coheigea:


rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/provider/AbstractOAuthDataProvider.java
```diff
@@ -225,6 +225,16 @@ protected ServerAccessToken createNewAccessToken(Client client, UserSubject user
     @Override
     public ServerAccessToken refreshAccessToken(Client client, String refreshTokenKey,
                                                 List<String> restrictedScopes) throws OAuthServiceException {
+        if (!recycleRefreshTokens) {
+            synchronized (refreshTokenLock) {
+                return doRefreshAccessToken(client, refreshTokenKey, restrictedScopes);
```
reta, line 230: Thanks @coheigea , the solution looks legit (since synchronized is reentrant), cleaning up the `doRefreshAccessToken` probably is too risky

reta (approved):

karlvr:
@coheigea Do you remember the reason for this change or do you know the reason for synchronisation in this code path at all? We had already worked around the synchronisation elsewhere for non-recycled refresh tokens and now we've had to work around this one (slightly harder due to the private method it calls). It doesn't _appear_ to us that there's any need to synchronise, there's no shared state? only the objects passed in are mutated. This style of global synchronisation block causes large problems for us in this highly concurrent point in the stack.

coheigea:
@karlvr I think it was the fix for https://cxf.apache.org/security-advisories.data/CVE-2026-50631.txt?version=2&modificationDate=1786101161000&api=v2. If you have a better proposal please raise a new PR for it.

PR #3137, opened 2026-05-22
Handle NaN or infinite times in JSON/JWT

coheigea:


rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwt/JwtUtils.java
```diff
@@ -50,9 +50,14 @@ public static void validateJwtExpiry(JwtClaims claims, int clockOffset, boolean
             return;
         }
         Instant now = Instant.now();
-        Instant expires = Instant.ofEpochMilli(expiryTime * 1000L);
-        if (clockOffset != 0) {
-            expires = expires.plusSeconds(clockOffset);
+        Instant expires;
+        try {
```
reta, line 54: @coheigea apologies, what changed here? `Instant.ofEpochMilli(expiryTime * 1000L)` is equivalent to `Instant.ofEpochSecond(expiryTime)` unless we are talking about overflow ...

reta (approved):

rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwt/JwtUtils.java
coheigea: @reta 1000L can overflow before creating the Instant, producing a wrong timestamp.

PR #3143, opened 2026-05-25
[CXF-9213]make RetryStrategy stateless

ffang:
### Problem

`RetryStrategy` carries a retry counter (`count`) as an instance field. When the same
instance is shared across invocations (e.g. configured as a Spring/Blueprint singleton
bean), concurrent or sequential call chains corrupt each other's counter, causing retries
to be skipped or exhausted prematurely.

### Approach

Introduce a `PerInvocationFailoverStrategy` interface that marks `FailoverStrategy`
implementations carrying per-invocation mutable state:

```java
public interface PerInvocationFailoverStrategy {
    FailoverStrategy newStrategy();
}
```

`FailoverTargetSelector` detects this interface at the start of each top-level invocation
(`prepare()`) and stores a fresh instance — obtained via `newStrategy()` — in a new
`inProgressStrategies` map keyed by invocation identity, parallel to the existing
`inProgress` map. The new `getStrategy(String key)` overload returns the per-invocation
instance when present, falling back to the shared strategy otherwise.

`RetryStrategy` implements `PerInvocationFailoverStrategy`. Its `newStrategy()` returns a
fresh `RetryStrategy` pre-configured with the same `maxNumberOfRetries`,
`alternateAddresses`, and `delayBetweenRetries` but with `count` reset to zero.

### Changes

- `PerInvocationFailoverStrategy.java` — new interface
- `RetryStrategy.java` — implements interface; simple instance `count`; `newStrategy()` copies full config
- `FailoverTargetSelector.java` — `inProgressStrategies` map; `getStrategy(String key)`; lifecycle hooks in `prepare`/`complete`
- `RetryStrategyTest.java` — unit tests for `newStrategy()` contract and per-invocation isolation
- `AbstractFailoverTest.java` — `CustomRetryStrategy.newStrategy()` delegates tracking to master instance

### DIFF
--- a/rt/features/clustering/src/main/java/org/apache/cxf/clustering/FailoverTargetSelector.java
+++ b/rt/features/clustering/src/main/java/org/apache/cxf/clustering/FailoverTargetSelector.java
@@ -53,6 +53,8 @@ public class FailoverTargetSelector extends AbstractConduitSelector {
 
     protected FailoverStrategy failoverStrategy;
     private ConcurrentHashMap<String, InvocationContext> inProgress = new ConcurrentHashMap<>();
+    // CXF-9213: per-invocation strategy instances for PerInvocationFailoverStrategy implementations.
+    private ConcurrentHashMap<String, FailoverStrategy> inProgressStrategies = new ConcurrentHashMap<>();
     private boolean supportNotAvailableErrorsOnly = true;
     private String clientBootstrapAddress;
 
@@ -113,6 +115,11 @@ public class FailoverTargetSelector extends AbstractConduitSelector {
                                       params,
                                       context);
             inProgress.putIfAbsent(key, invocation);
+            // CXF-9213: create a fresh per-invocation instance for strategies that carry state.
+            if (getStrategy() instanceof PerInvocationFailoverStrategy) {
+                inProgressStrategies.putIfAbsent(key,
+                    ((PerInvocationFailoverStrategy) getStrategy()).newStrategy());
+            }
         }
     }
 
@@ -179,6 +186,7 @@ public class FailoverTargetSelector extends AbstractConduitSelector {
 
         if (!failover) {
             inProgress.remove(key);
+            inProgressStrategies.remove(key);
             doComplete(exchange);
         }
     }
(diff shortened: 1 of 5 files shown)


rt/features/clustering/src/main/java/org/apache/cxf/clustering/FailoverTargetSelector.java
```diff
@@ -113,6 +113,13 @@ && getClientBootstrapAddress().equals(message.get(Message.ENDPOINT_ADDRESS))) {
                                       params,
                                       context);
             inProgress.putIfAbsent(key, invocation);
+
+            // CXF-9213: reset the per-exchange retry counter at the start of
+            // each new invocation so a singleton RetryStrategy starts from zero.
+            if (getStrategy() instanceof RetryStrategy) {
```
reta: @ffang thanks for looking into this, I think it would be better to not tangle RetryStrategy with exchanges and thread locals, what I have in mind (but need some time to flash out) is something along this lines: 
 - introduce new interface `PerInvocationFailoverStrategy` - that would indicate that the strategy is stateful
  ```java
   interface PerInvocationFailoverStrategy {
     FailoverStrategy newStrategy(InvocationContext  context, Exchange exchange);
   }
   ```
- store it in map the like `inProgress` does: 
  ```
   private ConcurrentHashMap<String, FailoverStrategy> inProgressStrategies = new ConcurrentHashMap<>();
   ```
- introduce `getStrategy(String key)` method to actually retrieve `FailoverStrategy` per invocation, that will fallback to `getStrategy()` if is not per invocation
- the `RetryStrategy` would implement `PerInvocationFailoverStrategy`  and return new fresh instance for each invocation


I think it should work,  may be I am missing some details, but I believe it will more maintainable solution, wdyt? Thanks again!

ffang: Hi @reta ,

Thanks for the feedback, let me think about the PerInvocationFailoverStrategy way.

Freeman

reta (approved):
Thanks @ffang !

ffang:
Thanks @reta !

PR #3154, opened 2026-05-28
[CXF-9218]Respect configured ProxySelector instead of hard-wiring sys…

ffang:
…tem proxy properties

## Summary

`ProxyFactory` was reading `http.proxyHost`, `http.proxyPort`, and `http.nonProxyHosts`
directly from system properties at construction time, bypassing Java's `ProxySelector`
mechanism entirely. This prevented applications from using custom `ProxySelector`
implementations — for example, wrapping `sun.net.spi.DefaultProxySelector` to selectively
bypass the proxy for specific hosts.

The fix removes the system-property reading from `ProxyFactory` and delegates to
`ProxySelector.getDefault()` instead. When no custom `ProxySelector` is installed the
behaviour is identical to before (`DefaultProxySelector` reads the same system properties
internally). When a custom `ProxySelector` is installed it is now consulted as expected.

The explicit `HTTPClientPolicy.setProxyServer(...)` path is unaffected.

## Behaviour notes

- **Default behaviour unchanged** — `DefaultProxySelector` reads `http.proxyHost` /
  `http.proxyPort` / `http.nonProxyHosts` exactly as before.
- Both `HttpClientHTTPConduit` and `URLConnectionHTTPConduit` benefit from this fix.

### DIFF
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/http/ProxyFactory.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/http/ProxyFactory.java
@@ -20,75 +20,15 @@ package org.apache.cxf.transport.http;
 
 import java.net.InetSocketAddress;
 import java.net.Proxy;
+import java.net.ProxySelector;
 import java.net.URI;
+import java.util.List;
 import java.util.regex.Pattern;
 
 import org.apache.cxf.common.util.StringUtils;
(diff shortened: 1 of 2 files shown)


reta (commented):
Much simpler, thanks @ffang !

reta (approved):

ffang:
Thanks @reta !

PR #3165, opened 2026-06-01
compare oauth2 secret tokens with constant-time MessageDigest.isEqual

dxbjavid:
The registration access-token checks in DynamicRegistrationService, the session authenticity token in RedirectionBasedGrantService, and the state token in MemoryClientCodeStateManager all authenticate a request-supplied secret with String.equals, which returns on the first differing byte.
Route them through a new OAuthUtils.compareTokens that wraps MessageDigest.isEqual, the same constant-time idiom already used by compareCertificateThumbprints and the JOSE verifiers.

### DIFF
--- a/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/utils/OAuthUtils.java
+++ b/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/utils/OAuthUtils.java
@@ -88,6 +88,13 @@ public final class OAuthUtils {
         }
     }
 
+    public static boolean compareTokens(String token1, String token2) {
+        if (token1 == null || token2 == null) {
+            return false;
+        }
+        return MessageDigest.isEqual(StringUtils.toBytesUTF8(token1), StringUtils.toBytesUTF8(token2));
+    }
+
     public static boolean compareCertificateThumbprints(X509Certificate cert, String encodedThumbprint) {
         try {
             byte[] thumbprint = createCertificateThumbprint(cert);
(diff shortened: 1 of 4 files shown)


rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/utils/OAuthUtils.java
```diff
@@ -88,6 +88,13 @@ public static void setCertificateThumbprintConfirmation(MessageContext mc, X509C
         }
     }
 
+    public static boolean compareTokens(String token1, String token2) {
+        if (token1 == null || token2 == null) {
+            return false;
+        }
+        return MessageDigest.isEqual(StringUtils.toBytesUTF8(token1), StringUtils.toBytesUTF8(token2));
```
reta, line 95: Thanks @dxbjavid , it looks somewhat non-intuitive to rely on `MessageDigest.isEqual` here (since tokens are not really message digests) but it does not look harmful. @coheigea wdyt?

reta (approved):

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/utils/OAuthUtils.java
coheigea: @reta It's a standard way of ensuring constant time comparison to avoid leaking information about which character the comparison fails at. I think we didn't do it before here as we're comparing tokens that should have high entropy, but it doesn't hurt to use it I guess.

coheigea (approved):

PR #3166, opened 2026-06-01
require azp when oidc id token has multiple audiences

dxbjavid:
Noticed OidcClaimsValidator only checks the azp claim when it is present. A multi-audience ID token that omits azp is accepted as long as the aud array contains the client id, so a token minted for a different relying party can be replayed here. OIDC Core 3.1.3.7 requires azp to be present once a token lists more than one audience; this rejects that case while leaving single-audience tokens unchanged.


coheigea:
I don't see the claim in https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation that "requires azp to be present once a token lists more than one audience" - can you point me to the exact text please?

dxbjavid:
You're right, that's on me. The wording I had in mind, "If the ID Token contains multiple audiences, the Client SHOULD verify that an azp Claim is present," is step 4 in the original Core 1.0 and in errata1, but errata set 2 (the version you linked) dropped it and replaced it with the weaker "validate the azp value as specified by those extensions" text. So the current spec doesn't require it the way I framed here. I'll close this one. Thanks for checking.

PR #3167, opened 2026-06-02
escape raw quotes after an escaped backslash in escapeJson

dxbjavid:
Found while fuzzing toJson/fromJson round-trips. escapeJson looked only at the single char before a " or \ to decide it was already escaped, so a value holding a complete escaped-backslash pair (\\) followed by a quote left that quote raw and broke out of its JSON string, letting a string value inject sibling members. Consume a backslash together with the char it escapes in one pass, so any leftover raw quote or backslash is always escaped while existing sequences are kept. Same single-char-lookback flaw as the reader fix in #3140.


coheigea (approved):

dxbjavid:

PR #3168, opened 2026-06-02
Properly handle nested arrays in json parsing

coheigea:
The issue was a bug in the JSON parser when it read arrays inside arrays.

Before the fix:

It knew how to handle objects inside a list, like [[{"a":1}]](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html).
But it did not handle nested lists, like [[]] or [[1,2]].
So it tried to read the inner list as if it were a number/text value.
That caused an unexpected NumberFormatException crash.

### DIFF
--- a/rt/rs/extensions/json-basic/src/main/java/org/apache/cxf/jaxrs/json/basic/JsonMapObjectReaderWriter.java
+++ b/rt/rs/extensions/json-basic/src/main/java/org/apache/cxf/jaxrs/json/basic/JsonMapObjectReaderWriter.java
@@ -285,6 +285,10 @@ public class JsonMapObjectReaderWriter {
                 readJsonObjectAsSettable(nextMap, json.substring(i + 1, closingIndex), depth + 1);
                 values.add(nextMap.map);
                 i = closingIndex + 1;
+            } else if (json.charAt(i) == ARRAY_START) {
+                int closingIndex = getClosingIndex(json, ARRAY_START, ARRAY_END, i);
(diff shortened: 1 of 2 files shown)


rt/rs/extensions/json-basic/src/main/java/org/apache/cxf/jaxrs/json/basic/JsonMapObjectReaderWriter.java
```diff
@@ -285,6 +285,10 @@ private List<Object> internalFromJsonAsList(String name, String json, int depth)
                 readJsonObjectAsSettable(nextMap, json.substring(i + 1, closingIndex), depth + 1);
                 values.add(nextMap.map);
                 i = closingIndex + 1;
+            } else if (json.charAt(i) == ARRAY_START) {
+                int closingIndex = getClosingIndex(json, ARRAY_START, ARRAY_END, i);
+                values.add(internalFromJsonAsList(name, json.substring(i + 1, closingIndex), depth + 1));
```
reta, line 294: Just wondering if we could be dealing with invalid json by being defensive here:
```suggestion
               if (closingIndex == -1) {
                   throw new UncheckedIOException(new IOException("Error in parsing json"));
               }
```

coheigea: @reta Good catch! I've applied that suggestion in a few places.

reta (approved):

PR #3182, opened 2026-06-04
compare username token password and digest in constant time

dxbjavid:
reading through STSStaxTokenValidator i noticed the local UsernameToken path checks the client-supplied password and password digest with String.equals, which bails out at the first mismatching character. that timing difference looks like enough to recover the expected digest (for a chosen nonce/created) or the plaintext password byte by byte and then authenticate. routed both comparisons through MessageDigest.isEqual over utf-8 bytes, which is the constant-time idiom already used elsewhere in cxf (OAuthUtils.compareTokens, HmacJwsSignatureVerifier). also made it null-safe so a missing password value just fails authentication instead of throwing.


coheigea (approved):

dxbjavid:

PR #3188, opened 2026-06-06
Add sensible default value to attachment-max-size property

reta:
Add sensible (50Mb) default value to `attachment-max-size` property

### DIFF
--- a/core/src/main/java/org/apache/cxf/attachment/AttachmentUtil.java
+++ b/core/src/main/java/org/apache/cxf/attachment/AttachmentUtil.java
@@ -222,7 +222,7 @@ public final class AttachmentUtil {
             if (size >= 0) {
                 bos.setMaxSize(size);
             } else {
-                LOG.warning("Max size value overflowed long. Do not set max size!");
+                LOG.warning("The max size value is set to unlimited.");
             }
         } else if (maxSize instanceof String) {
             try {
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/cxf/attachment/AttachmentUtil.java
```diff
@@ -214,24 +214,25 @@ public static void setStreamedAttachmentProperties(Message message, CachedOutput
...
+            maxSize = AttachmentDeserializer.DEFAULT_ATTACHMENT_MAX_SIZE;
+        }
+        if (maxSize instanceof Number) {
+            long size = ((Number) maxSize).longValue();
+            if (size >= 0) {
+                bos.setMaxSize(size);
             } else {
-                throw new IOException("The value set as " + AttachmentDeserializer.ATTACHMENT_MAX_SIZE
-                        + " should be either an instance of Number or String");
+                LOG.warning("Max size value overflowed long. Do not set max size!");
```
coheigea: Maybe we could change this warning message - I'm not sure what "Do not set max size!" means exactly? Maybe we should print the maximum allowed value?

coheigea (changes requested):

core/src/main/java/org/apache/cxf/attachment/AttachmentUtil.java
reta: Didn't even check old comments, but definitely could be better framed, thanks @coheigea !

```
LOG.warning("The max size value is set to unlimited.");
```

coheigea (approved):

PR #3212, opened 2026-06-12
strip cr/lf from attachment part headers in writeHeaders

dxbjavid:
AttachmentSerializer.writeHeaders writes each part header name and value straight into the multipart stream, so an attachment whose Content-Disposition carries a filename with a bare CR or LF (for instance a filename taken from an uploaded part) ends the header line early and injects extra part headers into the serialised MTOM or multipart body. It seemed safest to handle this where the bytes are actually written rather than at each place a header gets built, so this strips CR and LF from the header name and values in writeHeaders itself. There is a test that serialises an attachment with an embedded line break and checks the injected header does not appear.


reta (approved):
Thanks @dxbjavid , LGTM! @coheigea could you please cross check here? thanks!

coheigea (approved):

dxbjavid:
Thank you for the review and merge. Much appreciated

PR #3223, opened 2026-06-16
validate hex digits when decoding quoted-printable attachments

dxbjavid:
an inbound multipart attachment whose Content-Transfer-Encoding is quoted-printable is decoded by QuotedPrintableDecoderStream, which looks each of the two bytes following an '=' up in a 128 entry table without checking they are hex digits. a byte of 0x80 or above arrives as a negative java byte and throws ArrayIndexOutOfBoundsException, an invalid ascii pair like =GG is silently turned into 0x00, and a lowercase pair like =e2 decodes to the wrong byte because the table only holds the uppercase forms. this decodes each nibble with Character.digit over the unsigned byte and refuses anything that isn't valid hex, so malformed input is rejected rather than mis-decoded or crashing the parser.

### DIFF
--- a/core/src/test/java/org/apache/cxf/attachment/QuotedPrintableDecoderStreamTest.java
+++ b/core/src/test/java/org/apache/cxf/attachment/QuotedPrintableDecoderStreamTest.java
@@ -1,0 +1,66 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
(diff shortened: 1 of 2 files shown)


reta:
Thanks @dxbjavid , our baseline for 4.2/4.1 release line is JDK-17, could we please replace this custom hex coding  / decoding logic with https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/HexFormat.html ? Thank you.

dxbjavid:
good call, switched both nibbles over to HexFormat.fromHexDigit and wrapped the NumberFormatException as the IOException. the high-bit byte and lowercase cases still behave the same, and the custom table is gone now.

core/src/test/java/org/apache/cxf/attachment/QuotedPrintableDecoderStreamTest.java
```diff
@@ -0,0 +1,66 @@
...
+        int b;
+        while ((b = in.read()) != -1) {
+            out.write(b);
+        }
+        return out.toByteArray();
+    }
+
+    @Test
+    public void testDecodesHexPair() throws Exception {
+        assertArrayEquals("hello world".getBytes("US-ASCII"),
```
reta:
```suggestion
        assertArrayEquals("hello world".getBytes(StandardCharsets.US_ASCII),
```

To address PMD violations, there a few other places in this test with the same issue

dxbjavid: done, swapped all four getBytes calls in the test over to StandardCharsets.US_ASCII.

reta (approved):

dxbjavid:
Thank you for the review and merge. Much appreciated

PR #3235, opened 2026-06-18
Respect the AsyncRequestProducer contract by properly implementing available() method

reta:
Respect the `AsyncRequestProducer` contract by properly implementing available() method. Full credit goes to @rschmitt and @ok2c for pointing out at the issue (see please https://github.com/apache/httpcomponents-core/pull/674).

### DIFF
--- a/rt/transports/http-hc5/src/main/java/org/apache/cxf/transport/http/asyncclient/hc5/CXFHttpAsyncRequestProducer.java
+++ b/rt/transports/http-hc5/src/main/java/org/apache/cxf/transport/http/asyncclient/hc5/CXFHttpAsyncRequestProducer.java
@@ -94,24 +94,7 @@ public class CXFHttpAsyncRequestProducer implements AsyncRequestProducer {
 
     @Override
     public int available() {
-        if (content != null) {
-            final ByteBuffer b = buffer;
-            if (b != null && b.hasRemaining()) {
-                return b.remaining();
-            } 
-            final InputStream in = fis;
-            if (in != null) {
-                try {
-                    return in.available();
-                } catch (final IOException ex) {
-                    /* Nothing to do */
-                }
-            }
-        } else if (buf != null && buf.hasData()) {
-            return buf.length(); /* returns remaining */
-        }
-
-        return 0; /* no more data */
+        return Integer.MAX_VALUE;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


ok2c:
@reta Damn. One needs to be extremely careful here. 

1. Classic `InputSteam` / `OutputStream` abstraction is inherently blocking. It just does not mix well with an event driven i/o model. 

2. If at any point the data producer signals zero data availability, the protocol handler can choose to disable output interest for the channel and let it go to sleep. That means no more output for that channel until the producer signals its readiness to produce more through `DataStreamChannel#requestOutput`. The trouble is the actual data stream producer is hidden behind that `InputStream` and it basically has no way of pushing data through `DataStreamChannel` or requesting output. That means the channel will remain stuck in sleep mode (no output interest).

Sadly, i do not see elegant way of solving this issue. One either must use an extra thread to read data from the blocking `InputSteam` or always return **non-zero** data availability to ensure the channel cannot go to sleep. This, of course, basically means the i/o reactor event loop will be running hot.

Can you represent your data as anything other than `InputSteam`?

reta:
> @reta Damn. One needs to be extremely careful here.

Thanks a lot for the review, @ok2c , really appreciate it

> The trouble is the actual data stream producer is hidden behind that InputStream and it basically has no way of pushing data through DataStreamChannel or requesting output. That means the channel will remain stuck in sleep mode (no output interest).

Agree, I also found no better way but rely on `InputStream::available()` (which may return 0), the only good news here -  by design (javadoc at least), this method should not block.

> Can you represent your data as anything other than InputSteam?

We sadly support dual (buffer / `InputStream`) for a legacy reasons, replacing it would require large scale changes, I will look into it, thank you.

rschmitt:
> This, of course, basically means the i/o reactor event loop will be running hot.

It also means the event loop thread will block indefinitely while reading from `fis`, which will stall all IO on that thread. In the general case, this whole arrangement is fundamentally unfixable without an additional thread (which on Java 25 could be a virtual thread). The fact that this works at all leads me to wonder whether `fis` is reliably backed by some sort of non-blocking data source like a file or an in-memory buffer.

reta:
> The fact that this works at all leads me to wonder whether `fis` is reliably backed by some sort of non-blocking data source like a file or an in-memory buffer.

So the `fis` is backed by `CachedOutputStream` which could be backed by in-memory buffer (to my knowledge, this is the true most of the time).

rschmitt (approved):
Although I don't think this PR fixes the backpressure issues in the general case, it can't hurt. It's a strict improvement over the current `return 0` implementation of `available()`.

ok2c:
I think it would be safer to just all ways return Integer.MAX_VALUE from this method given there is nothing that can trigger`DataStreamChannel#requestOutput` in case it ever returns 0.
```
    @Override
    public int available() {
        return Integer.MAX_VALUE;
    }
```

PR #3240, opened 2026-06-20
CXF-9221: JCache providers use inverted isExpired() logic causing expired tokens/codes to never be evicted

reta:
JCache providers use inverted isExpired() logic causing expired tokens/codes to never be evicted


reta:
@coheigea could you please take a look when time permits? thank you

coheigea (approved):

PR #3241, opened 2026-06-21
validate redirect origin in oidc rp sign-in completion

dxbjavid:
when the rp sign-in flow completes, OidcRpAuthenticationService redirects the browser to the state value, which OidcRpAuthenticationFilter copies straight from the current request parameters, so a request such as /rp/complete?state=https://evil.example against an authenticated session returns a 303 to an arbitrary external host, an open redirect. the legitimate value is always the application's own request uri, so completeAuthentication now only honours a location that is relative or shares the same scheme and authority as the base path, otherwise it falls back to the configured default location. the added test covers the cross-origin, protocol-relative and userinfo-host variants.


rt/rs/security/sso/oidc/src/main/java/org/apache/cxf/rs/security/oidc/rp/OidcRpAuthenticationService.java
```diff
@@ -57,18 +57,42 @@ public Response completeAuthentication(@Context OidcClientTokenContext oidcConte
         URI redirectUri = null;
         MultivaluedMap<String, String> state = oidcContext.getState();
         String location = state != null ? state.getFirst("state") : null;
-        if (location == null && defaultLocation != null) {
+        if (location != null) {
```
reta: @coheigea need your advice please, the usage of `state` for redirect URL looks very suspicious, I haven't found any references in OIDC RP specs of such designation, is it correct or I am missing something? thank you

dxbjavid: you're right that it isn't an OIDC concept, the spec treats state as opaque. this is cxf's own reuse of the parameter name rather than anything from the RP specs.

it comes from OidcRpAuthenticationFilter: when addRequestUriAsRedirectQuery is set, the filter writes the original request uri into a query parameter it happens to call state (redirectBuilder.queryParam("state", rc.getUriInfo().getRequestUri().toString())), and completeAuthentication later reads that same state back to send the browser to where it was originally heading. so in normal use the value is always the application's own request uri, which is why the legitimate case is same-origin.

the problem is that toRequestState copies every current request parameter into the context state, so /rp/complete?state=https://evil.example against an already-authenticated session is read back verbatim and returned as a 303 Location, an open redirect. the origin check just keeps that redirect within the application's own scheme and authority. if you'd rather the validation lived in the filter where the value is first written, i'm happy to move it there.

reta: > . if you'd rather the validation lived in the filter where the value is first written, i'm happy to move it there.

Oh thanks for clarifying @dxbjavid , I think hardening filter would make sense, also making sure the `state` is not colliding with the request state (may be filtering here will help), thanks again

dxbjavid: moved it into the filter. toRequestState now drops the state value when it isn't same-origin, so the caller-supplied parameter can't collide with the redirect query the filter writes and completeAuthentication goes back to its original form. relative and same scheme/authority values are kept, cross-origin, protocol-relative and userinfo-host ones are dropped and it falls back to defaultLocation. the test moved to OidcRpAuthenticationFilterTest to cover toRequestState directly.

rt/rs/security/sso/oidc/src/main/java/org/apache/cxf/rs/security/oidc/rp/OidcRpAuthenticationFilter.java
```diff
@@ -115,8 +115,37 @@ private MultivaluedMap<String, String> toRequestState(ContainerRequestContext rc
...
+        try {
+            uri = URI.create(location);
+        } catch (IllegalArgumentException ex) {
+            return false;
+        }
+        if (uri.getScheme() == null && uri.getAuthority() == null) {
+            // a path-only reference is resolved by the browser against the current request
+            return true;
+        }
+        String basePath = (String)mc.get("http.base.path");
```
reta: Wondering if we could we use `rc.getUriInfo().getAbsolutePath()` here to verify `scheme` and `authority` ?

dxbjavid: good shout, switched it over. isSameOrigin now takes the request context and compares the location scheme and authority against rc.getUriInfo().getAbsolutePath(), so it no longer reaches for http.base.path off the message context. the filter test mocks getAbsolutePath instead and the five cases still pass.

rt/rs/security/sso/oidc/src/test/java/org/apache/cxf/rs/security/oidc/rp/OidcRpAuthenticationFilterTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+    private static final URI ABSOLUTE_PATH = URI.create("https://app.example.com:8080/services/rp/complete");
+
+    @Test
+    public void testDropsCrossOriginState() {
+        MultivaluedMap<String, String> state = requestState("https://evil.example.com/phish");
+        assertFalse(state.containsKey("state"));
+    }
+
+    @Test
```
reta, line 47: Super nit, could you please add test case without authority part:

```suggestion
    @Test
    public void testDropsNoAuthority() {
        MultivaluedMap<String, String> state = requestState("http:/");
        assertFalse(state.containsKey("state"));
    }
```

reta (approved):
@coheigea could you please take a look? thank you!

rt/rs/security/sso/oidc/src/test/java/org/apache/cxf/rs/security/oidc/rp/OidcRpAuthenticationFilterTest.java
dxbjavid: added it, testDropsNoAuthority covers the scheme-only http:/ case and it gets dropped since there's no authority to match. all six pass.

coheigea (approved):

dxbjavid:
Thank you both for the thorough review. I appreciate your time.

PR #3255, opened 2026-06-26
CXF-9226: AsyncHTTPConduit throws ISE instead of HTTPException(407) o…

ffang:
…n proxy auth failure

When the HC5 async conduit receives a 407 Proxy Authentication Required response with wrong credentials and MaxRetransmits>0 or AutoRedirect=true (as in the JIRA), the request entity is marked repeatable so HC5 retries after the first 407. A concurrent write-failure on the Connection: close socket races with the retry's execute() call, causing InternalHttpAsyncExecRuntime.ensureValid() to find a null endpointRef and throw IllegalStateException. Without MaxRetransmits/AutoRedirect the entity is not repeatable and HC5 intercepts the 407 internally, calling asyncExecCallback.completed() without signaling CXF's future and leaving getHttpResponse() blocked indefinitely on wait().

Fix:
- Promote HttpClientContext ctx to a field on AsyncWrappedOutputStream so it is accessible after HC5 calls back.
- Replace wait() with wait(receiveTimeout) to bound the wait.
- In both the timeout path and the RuntimeException (ISE) path, check ctx.getResponse() for a 407 status and throw HTTPException(407) instead of re-throwing the raw exception or blocking forever.

Test:
- ProxyAuthIllegalStateTest: reproduces the ISE via a raw ServerSocket proxy returning 407+Connection:close, with MaxRetransmits=5 and AutoRedirect=true matching the exact JIRA configuration.

### DIFF
--- a/rt/transports/http-hc5/src/main/java/org/apache/cxf/transport/http/asyncclient/hc5/AsyncHTTPConduit.java
+++ b/rt/transports/http-hc5/src/main/java/org/apache/cxf/transport/http/asyncclient/hc5/AsyncHTTPConduit.java
@@ -72,14 +72,13 @@ import org.apache.cxf.transports.http.configuration.HTTPClientPolicy;
 import org.apache.cxf.version.Version;
 import org.apache.cxf.ws.addressing.EndpointReferenceType;
 import org.apache.hc.client5.http.async.HttpAsyncClient;
-import org.apache.hc.client5.http.auth.AuthCache;
-import org.apache.hc.client5.http.auth.AuthScheme;
 import org.apache.hc.client5.http.auth.AuthSchemeFactory;
 import org.apache.hc.client5.http.auth.AuthScope;
 import org.apache.hc.client5.http.auth.Credentials;
 import org.apache.hc.client5.http.auth.UsernamePasswordCredentials;
 import org.apache.hc.client5.http.config.RequestConfig;
 import org.apache.hc.client5.http.impl.async.CloseableHttpAsyncClient;
+import org.apache.hc.client5.http.impl.auth.BasicAuthCache;
 import org.apache.hc.client5.http.impl.auth.BasicCredentialsProvider;
 import org.apache.hc.client5.http.impl.auth.BasicScheme;
 import org.apache.hc.client5.http.protocol.HttpClientContext;
@@ -557,24 +556,9 @@ public class AsyncHTTPConduit extends HttpClientHTTPConduit {
 
             ctx.setCredentialsProvider(credsProvider);
             if (proxyAuthorizationPolicy != null && proxyAuthorizationPolicy.getUserName() != null) {
-                // We only support BasicAuth
-                ctx.setAuthCache(new AuthCache() {
-                    @Override
-                    public void remove(HttpHost host) {
-                    }
-
-                    @Override
-                    public void put(HttpHost host, AuthScheme authScheme) {
-                    }
-                    @Override
-                    public AuthScheme get(HttpHost host) {
-                        return new BasicScheme();
-                    }
-
-                    @Override
-                    public void clear() {
-                    }
-                });
+                final BasicAuthCache cache = new BasicAuthCache();
+                cache.put(entity.getConfig().getProxy(), new BasicScheme());
+                ctx.setAuthCache(cache);
             }
 
             TlsStrategy tlsStrategy = null;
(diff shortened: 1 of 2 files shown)


rt/transports/http-hc5/src/main/java/org/apache/cxf/transport/http/asyncclient/hc5/AsyncHTTPConduit.java
```diff
@@ -673,7 +675,12 @@ protected synchronized HttpResponse getHttpResponse() throws IOException {
             while (httpResponse == null) {
                 if (exception == null) { //already have an exception, skip waiting
                     try {
-                        wait();
+                        long timeout = csPolicy.getReceiveTimeout();
+                        if (timeout > 0) {
+                            wait(timeout);
```
reta: I think this changes semantics: to my knowledge, the receiveTimeout is measured between bytes received (so fe slow server could send one byte somewhere around `receiveTimeout`), but it is not for complete response, the caller could set timeout on the future instead

rt/transports/http-hc5/src/main/java/org/apache/cxf/transport/http/asyncclient/hc5/AsyncHTTPConduit.java
```diff
@@ -686,12 +693,34 @@ protected synchronized HttpResponse getHttpResponse() throws IOException {
                         if (exception instanceof IOException) {
                             throw (IOException)exception;
                         } else if (exception instanceof RuntimeException) {
-                            throw (RuntimeException)exception;
+                            // HC5 async calls failed() with RuntimeException
```
reta: @ffang I think I figured out the underlying problem, that is related to pre-populating the  `AuthCache` properly (which is the reason we see the `IllegalStateException` in the first place. I think I should be able to hae a fix shortly, could share the snippet (or push one commit to you branch), whichever is the best for you, thanks!

ffang: Hi @reta ,

You can push your change on my branch.

Thanks for looking at it!
Freeman

rt/transports/http-hc5/src/main/java/org/apache/cxf/transport/http/asyncclient/hc5/AsyncHTTPConduit.java
```diff
@@ -553,6 +556,26 @@ public Credentials getCredentials(final AuthScope authscope, HttpContext context
             };
 
             ctx.setCredentialsProvider(credsProvider);
+            if (proxyAuthorizationPolicy != null && proxyAuthorizationPolicy.getUserName() != null) {
+                // We only support BasicAuth
+                ctx.setAuthCache(new AuthCache() {
```
reta: @ffang the authentication handler looks for  `AuthScheme` in `AuthCache` in case of proxy auth, I did a "
"dummy" implementation for a sake of showcasing the idea, but I think we could do better job by using BasicAuthCache and proxy host/port, looking into it.

ffang: Hi @reta ,

This is promising, if  I read it correctly, the ISE ("AuthScheme is null") occurs because when HC5 processes a 407, it transitions the auth exchange to CHALLENGED state but leaves authScheme null (auth scheme selection hasn't happened yet). When a second 407 arrives, it finds authScheme == null and asserts.

If the AuthCache is pre-populated with the proxy's auth scheme (e.g., BasicScheme for Basic realm), HC5 knows upfront which scheme to use for proxy auth. The auth exchange initializes with a non-null authScheme, so the Asserts.notNull(authScheme, "AuthScheme") check in AuthenticationHandler.handleResponse() never fires. The ISE is avoided entirely.

Thanks for it!
Freeman

reta: Exactly, I had some difficulties with properly passing pathPrefix, but I think we will make it work, the proxy host / port is available, I we try to wrap it up today, thanks!

reta: OK, I think we should be all set, thanks @ffang !

reta (approved):

rt/transports/http-hc5/src/main/java/org/apache/cxf/transport/http/asyncclient/hc5/AsyncHTTPConduit.java
ffang: Thanks a lot @reta !

PR #3274, opened 2026-07-02
 (3.6.x) [CXF-9227] - Fix SecurityManager permission regressions

fabiobrz:
Based on the changes in #3256.

ProxyFactory.java is not in scope since it does not rely on ProxySelector in 3.6.x.

@coheigea, @ffang WDYT?


ffang (approved):
LGTM, thanks @fabiobrz !

PR #3279, opened 2026-07-03
Disable decoupled destinations by default for WS-Addressing

coheigea:
(Non-anonymous) decoupled destinations represent an SSRF risk and are not needed for most WS-Addressing scenarios. This PR disables it by default and adds a system property to turn them on again. The flag only applies for WS-Addressing, as WS-RM relies on decoupled addresses we allow it there.

### DIFF
--- a/core/src/main/java/org/apache/cxf/ws/addressing/ContextUtils.java
+++ b/core/src/main/java/org/apache/cxf/ws/addressing/ContextUtils.java
@@ -67,6 +67,30 @@ public final class ContextUtils {
         "org.apache.cxf.ws.addressing.decoupled.allowedSchemes";
 
     /**
+     * System property that enables WS-Addressing decoupled destinations (non-anonymous
+     * wsa:ReplyTo / wsa:FaultTo). Defaults to {@code false} — decoupled destinations
+     * are <em>disabled</em> by default to prevent SSRF. Set to {@code true} only when
+     * your deployment legitimately requires decoupled WS-Addressing callbacks.
+     */
+    public static final String WS_ADDRESSING_DECOUPLED_ENABLED_PROPERTY =
+        "org.apache.cxf.ws.addressing.decoupled.enabled";
+
+    /**
+     * Exchange property key that higher-level protocols (e.g. WS-RM) may set to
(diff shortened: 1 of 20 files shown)


reta (approved):

reta:
Thanks @coheigea !

ppalaga:
Thanks for this fix, @coheigea. When testing with it I have found a small issue: 

When setting `decoupled.enabled=true` and reducing `allowedSchemes` to just `https://` via system property, the error message still says 

```
Decoupled WS-Addressing ReplyTo (http://localhost:8081/ws-addressing-target/replyTo) is not permitted by this server. Enable with system property org.apache.cxf.ws.addressing.decoupled.enabled=true
```

That's misleading, because `decoupled.enabled=true` is set already. It should recommend to expand the `allowedSchemes`.

Besides that, as usual, I would like to let Quarkus CXF users to use MicroProfile config to configure the decoupled permissions. Implementing it is easy, the only question is, whether we could find a way for Quarkus CXF to override the error and log messages to use `quarkus.cxf.endpoint.addressing.decoupled.enabled` and `quarkus.cxf.endpoint.addressing.decoupled.allowed-schemes` instead of `org.apache.cxf.ws.addressing.decoupled.enabled` and `org.apache.cxf.ws.addressing.decoupled.allowedSchemes` respectively? An SPI would perhaps work, but it sounds like an overkill for such a simple thing. Could we perhaps introduce some system properties for that?

I'd be happy to send a PR for both issues in case it sounds reasonable to you?

coheigea:
@ppalaga Thanks for taking a look, please submit a PR for the misleading error message. For the Quarkus proposal, do we have an existing template for this for other configuration properties?

ppalaga:
> @ppalaga Thanks for taking a look, please submit a PR for the misleading error message. For the Quarkus proposal, do we have an existing template for this for other configuration properties?

Thanks, @coheigea, this PR https://github.com/apache/cxf/pull/3300 covers both: 

* Adds `,or set ...allowedSchemes` to the log or fault messages where needed
* Puts the messages into a resource bundle, which allows us to override them using some class loading trickery on Quarkus CXF.

PR #3281, opened 2026-07-03
match saml sso issuer exactly instead of by prefix

dxbjavid:
validateIssuer in the SAML 2.0 SSO response validator checks the received issuer against the configured issuer IDP with issuerIDP.startsWith(issuer.getValue()), so any value that is only a prefix of the expected issuer satisfies the enforceKnownIssuer check, right down to a single character. That means the known-issuer control does not really pin the issuer, and where a service provider trusts more than one identity provider it weakens the binding between an assertion and the IdP it is meant to have come from. Compare the issuer to the configured value exactly instead, which is how SAML entityIDs are meant to be matched.

### DIFF
--- a/rt/rs/security/sso/saml/src/main/java/org/apache/cxf/rs/security/saml/sso/SAMLSSOResponseValidator.java
+++ b/rt/rs/security/sso/saml/src/main/java/org/apache/cxf/rs/security/saml/sso/SAMLSSOResponseValidator.java
@@ -169,8 +169,8 @@ public class SAMLSSOResponseValidator {
             return;
         }
 
-        // Issuer value must match (be contained in) Issuer IDP
-        if (enforceKnownIssuer && (issuer.getValue() == null || !issuerIDP.startsWith(issuer.getValue()))) {
+        // Issuer value must match the configured Issuer IDP
+        if (enforceKnownIssuer && !issuerIDP.equals(issuer.getValue())) {
             LOG.warning("Issuer value: " + issuer.getValue() + " does not match issuer IDP: "
                 + issuerIDP);
             throw new WSSecurityException(WSSecurityException.ErrorCode.FAILURE, "invalidSAMLsecurity");
--- a/rt/rs/security/sso/saml/src/main/java/org/apache/cxf/rs/security/saml/sso/SAMLSSOResponseValidator.java
+++ b/rt/rs/security/sso/saml/src/main/java/org/apache/cxf/rs/security/saml/sso/SAMLSSOResponseValidator.java
@@ -18,6 +18,7 @@
  */
 package org.apache.cxf.rs.security.saml.sso;
 
+import java.net.URI;
 import java.time.Instant;
 import java.util.List;
 import java.util.logging.Logger;
@@ -47,6 +48,7 @@ public class SAMLSSOResponseValidator {
     private boolean enforceResponseSigned;
     private boolean enforceAssertionsSigned = true;
     private boolean enforceKnownIssuer = true;
+    private boolean enforceStrictIssuerMatch;
     private TokenReplayCache<String> replayCache;
 
     /**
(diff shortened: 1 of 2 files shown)


rt/rs/security/sso/saml/src/main/java/org/apache/cxf/rs/security/saml/sso/SAMLSSOResponseValidator.java
```diff
@@ -169,8 +169,8 @@ private void validateIssuer(org.opensaml.saml.saml2.core.Issuer issuer) throws W
             return;
         }
 
-        // Issuer value must match (be contained in) Issuer IDP
-        if (enforceKnownIssuer && (issuer.getValue() == null || !issuerIDP.startsWith(issuer.getValue()))) {
+        // Issuer value must match the configured Issuer IDP
+        if (enforceKnownIssuer && !issuerIDP.equals(issuer.getValue())) {
```
reta: @coheigea could you please take a look? I believe you have changed exact much to prefix here [1], thank you

[1] https://github.com/apache/cxf/commit/9253b78fb7280439dd4e5d44f798566055d7a192

dxbjavid: That's the commit, yes. As far as I can tell the reason for the prefix check is that validateIssuer compares the response Issuer against requestState.getIdpServiceAddress() (set in AbstractRequestAssertionConsumerHandler), which is the IdP's SSO endpoint URL rather than its entityID, and the entityID is often a prefix of that address. The trouble is that a bare startsWith accepts any prefix at all, down to a single character, so with enforceKnownIssuer on the issuer isn't really pinned to anything. If exact matching against the service address is too strict for existing deployments, a separate expected issuer property might be the safer route, but happy to rework it whichever way you and Colm prefer.

coheigea: It'll be a week or so until I get to look at this, sorry for the delay

coheigea: How about something like this as a middle ground? We could make it configurable as well to accept strict matching only.
```
private boolean matchesKnownIssuer(String issuerValue) {
        if (issuerValue == null || issuerIDP == null || !issuerIDP.startsWith(issuerValue)) {
            return false;
        }

        if (issuerIDP.equals(issuerValue)) {
            return true;
        }

        // Keep prefix compatibility, but for URL-based IdP values require a URL-based issuer
        // on the same scheme/host/port to avoid accepting arbitrary short prefixes.
        URI issuerIdpUri = toUri(issuerIDP);
        if (isHierarchicalAbsoluteUri(issuerIdpUri)) {
            URI issuerUri = toUri(issuerValue);
            return isHierarchicalAbsoluteUri(issuerUri)
                && issuerIdpUri.getScheme().equalsIgnoreCase(issuerUri.getScheme())
                && issuerIdpUri.getHost().equalsIgnoreCase(issuerUri.getHost())
                && getEffectivePort(issuerIdpUri) == getEffectivePort(issuerUri);
        }

        return true;
    }

    private URI toUri(String value) {
        try {
            return URI.create(value);
        } catch (IllegalArgumentException ex) {
            return null;
        }
    }

    private boolean isHierarchicalAbsoluteUri(URI uri) {
        return uri != null && uri.isAbsolute() && uri.getHost() != null;
    }

    private int getEffectivePort(URI uri) {
        if (uri.getPort() != -1) {
            return uri.getPort();
        }
        if ("http".equalsIgnoreCase(uri.getScheme())) {
            return 80;
        }
        if ("https".equalsIgnoreCase(uri.getScheme())) {
            return 443;
        }
        return -1;
    }
```

dxbjavid: that works for me, thanks. i have pushed your matchesKnownIssuer version pretty much as written, so a same-origin prefix like the entityID being a prefix of the SSO endpoint still passes, but arbitrary short prefixes on a different scheme/host/port no longer do. i also added an enforceStrictIssuerMatch flag (default false to keep the current behaviour) for deployments that want exact matching only. the tests now cover a rejected short prefix, an accepted same-origin prefix, and the strict case rejecting the prefix.

coheigea (approved):

coheigea:
@reta It's good for me, let me know what you think

reta:
> @reta It's good for me, let me know what you think

LGTM, thanks a lot @coheigea and @dxbjavid

reta (approved):

dxbjavid:
@reta  @coheigea  Thank you for your time. Appreciated

PR #3288, opened 2026-07-07
Encode LDAP query parameter names as well

coheigea:


rt/rs/extensions/search/src/main/java/org/apache/cxf/jaxrs/ext/search/ldap/LdapQueryVisitor.java
```diff
@@ -60,8 +60,9 @@ public void visit(SearchCondition<T> sc) {
                 }
 
                 String ldapOperator = conditionTypeToLdapOperator(sc.getConditionType());
+                String encodedName = Util.doRFC2254Encoding(name);
```
reta: Should we also respect the `encodeQueryValues` here?
```suggestion
                String encodedName =  encodeQueryValues ? Util.doRFC2254Encoding(name)  : rvalStr
```

reta (approved):

rt/rs/extensions/search/src/main/java/org/apache/cxf/jaxrs/ext/search/ldap/LdapQueryVisitor.java
coheigea: Yes I suppose so, I'll include this

PR #3294, opened 2026-07-08
Enforce that Access tokens only are accepted in JwtAccessTokenValidator

coheigea:


rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/filters/JwtAccessTokenValidator.java
```diff
@@ -76,6 +81,18 @@ protected void validateToken(JwtToken jwt) {
         JwtUtils.validateTokenClaims(jwt.getClaims(), getTtl(), getClockOffset(), isValidateAudience());
     }
 
+    private void validateTokenType(JwtToken jwt) {
+        Object tokenType = jwt.getJwsHeader(JoseConstants.HEADER_TYPE);
+        if (tokenType != null && !JoseConstants.TYPE_AT_JWT.equals(tokenType.toString())) {
```
reta, line 86: I am wondering if token type has to be present all the time? (== required) Or it is optional?


```suggestion
        if (tokenType == null || !JoseConstants.TYPE_AT_JWT.equals(tokenType.toString())) {
```

reta (approved):

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/filters/JwtAccessTokenValidator.java
coheigea: @reta Copilot tells me that:

```
token_use is not a standard JWT access-token claim across issuers, so making it mandatory by default would break interoperability.
typ=at+jwt is the better protocol-level discriminator, but many deployments still omit it.
```

So I think we should keep the current logic, which is intended to make sure that an IdToken can't be processed by the validator.

PR #3297, opened 2026-07-09
Disallow a dynamic registration client from requesting scopes without…

coheigea:
… validation

### DIFF
--- a/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/services/DynamicRegistrationService.java
+++ b/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/services/DynamicRegistrationService.java
@@ -19,6 +19,7 @@
 package org.apache.cxf.rs.security.oauth2.services;
 
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 
 import jakarta.ws.rs.Consumes;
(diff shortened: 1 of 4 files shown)


rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/services/DynamicRegistrationService.java
```diff
@@ -417,6 +423,28 @@ public void setUserRole(String userRole) {
...
+            return;
+        }
+
+        List<String> allowedScopes = allowedClientScopes;
+        if (allowedScopes == null && clientProvider instanceof AbstractOAuthDataProvider) {
+            allowedScopes = new java.util.ArrayList<>(
+                ((AbstractOAuthDataProvider)clientProvider).getPermissionMap().keySet());
+        }
+
+        if (allowedScopes != null && !new HashSet<>(allowedScopes).containsAll(requestedScopes)) {
```
reta, line 441: Do we care about string case sensitivity? (lower case <-> upper case fe)

reta (approved):

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/services/DynamicRegistrationService.java
coheigea: RFC 6749, Section 3.3 says the scope parameter is a space-delimited list of case-sensitive strings defined by the authorization server. So I think we are OK to require case sensitivity.

PR #3299, opened 2026-07-09
[CXF-9229] Add PQC TLS 1.3 support via X25519MLKEM768 hybrid KEM

ffang:
- TLSParameterBase: new namedGroups API (getNamedGroups/setNamedGroups) to configure the ordered list of TLS named groups offered during the handshake
- security.xsd: new NamedGroups complex type and <sec:namedGroups>/<sec:namedGroup>
  elements on both TLSClientParametersType and TLSServerParametersType
- TLSClientParametersConfig / TLSServerParametersConfig: wire namedGroups from XML config into SSLParameters via SSLUtils.applyNamedGroups() (JDK 20+)
- SSLUtils.applyNamedGroupsBC(): reflection-based helper to set named groups on BC JSSE's ProvSSLEngine without a compile-time bctls-jdk18on dependency
- JettyHTTPServerEngine: override customize(SSLEngine) to call applyNamedGroupsBC() so each accepted engine gets named groups from tlsServerParameters configuration
- BCJsseServerALPNProcessor: ALPNProcessor.Server for BC JSSE's ProvSSLEngine; bridges it into Jetty's ALPN negotiation (JDK9ServerALPNProcessor only accepts java.base engines); registered via META-INF/services; enables HTTP/2 + BC JSSE
- BCJssePQCTest + BCJsseJettyServer: system test using BC JSSE on JDK 17-26; sets jdk.tls.namedGroups=X25519MLKEM768 before SSLContext creation so the BC JSSE context-level named-group map includes the PQC group; verifies X25519MLKEM768 was negotiated by restricting a raw BCSSLSocket to that group only
- PQCTLSTest: system test for SunJSSE on JDK 27+ (JEP 527); skips automatically when X25519MLKEM768 is absent from the configured named groups

### DIFF
--- a/rt/transports/http/src/main/java/org/apache/cxf/transport/https/SSLUtils.java
+++ b/rt/transports/http/src/main/java/org/apache/cxf/transport/https/SSLUtils.java
@@ -18,6 +18,7 @@
  */
 package org.apache.cxf.transport.https;
 
+import java.lang.reflect.Method;
 import java.net.Socket;
 import java.nio.ByteBuffer;
 import java.security.GeneralSecurityException;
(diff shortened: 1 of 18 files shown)


rt/transports/http/src/main/java/org/apache/cxf/transport/https/SSLUtils.java
```diff
@@ -531,4 +532,43 @@ public javax.security.cert.X509Certificate[] getPeerCertificateChain() throws SS
         }
     };
 
+    /**
+     * Applies the given named groups to {@code SSLParameters} via reflection.
+     * {@code SSLParameters.setNamedGroups()} requires JDK 20+; this method
+     * silently skips the call on older JDKs.
+     */
+    public static void applyNamedGroups(SSLParameters params, List<String> namedGroups) {
```
reta, line 540: Wondering if we should consider multi-release jar and have clean JDK-20+ implementation?

reta (approved):

reta:
Thanks @ffang !

rt/transports/http/src/main/java/org/apache/cxf/transport/https/SSLUtils.java
ffang: Thanks @reta !

IMO it's too much to maintain multi-release jars for it.  Probably I think it's time to consider bump the minimum JDK version to JDK21 for CXF 4.3.

Best Regards
Freeman

PR #3301, opened 2026-07-09
Add `maxSize` to BinaryDataProvider so the payload size could be constrained (to 1Gb by default)

reta:
Add `maxSize` to BinaryDataProvider so the payload size could be constrained (to 1Gb by default). Adding a new system property `org.apache.cxf.binary-max-size` to configure the default.

### DIFF
--- a/core/src/main/java/org/apache/cxf/helpers/IOUtils.java
+++ b/core/src/main/java/org/apache/cxf/helpers/IOUtils.java
@@ -34,10 +34,14 @@ import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.util.Objects;
 
+import org.apache.cxf.common.util.SystemPropertyAction;
 import org.apache.cxf.io.CopyingOutputStream;
 import org.apache.cxf.io.Transferable;
 
 public final class IOUtils {
+    public static final int DEFAULT_BINARY_MAX_SIZE =
+        SystemPropertyAction.getInteger("org.apache.cxf.binary-max-size", 1073741824 /* 1Gb */);
+
     public static final Charset UTF8_CHARSET = java.nio.charset.StandardCharsets.UTF_8;
     public static final int DEFAULT_BUFFER_SIZE = 1024 * 4;
 
@@ -176,7 +180,7 @@ public final class IOUtils {
         if (output instanceof CopyingOutputStream) {
             return ((CopyingOutputStream)output).copyFrom(input);
         }
-        return copy(input, output, DEFAULT_BUFFER_SIZE);
+        return copy(input, output, DEFAULT_BUFFER_SIZE, -1);
     }
 
     public static int copyAndCloseInput(final InputStream input,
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/cxf/helpers/IOUtils.java
```diff
@@ -228,6 +232,10 @@ public static int copy(final InputStream input, final OutputStream output,
             output.write(buffer, 0, n);
             total += n;
             n = input.read(buffer);
+
+            if (maxSize > 0 /* -1 sets to unlimited */ && total > maxSize) {
```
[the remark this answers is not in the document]
reta: It will allocate buffer at max

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/provider/BinaryDataProvider.java
```diff
@@ -105,7 +110,7 @@ public T readFrom(Class<T> clazz, Type genericType, Annotation[] annotations, Me
                 return clazz.cast(new InputStreamReader(is, getEncoding(type)));
             }
             if (byte[].class.isAssignableFrom(clazz)) {
-                return clazz.cast(IOUtils.readBytesFromStream(is));
+                return clazz.cast(IOUtils.readBytesFromStream(is, maxSize));
             }
             if (File.class.isAssignableFrom(clazz)) {
```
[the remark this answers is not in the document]
reta: This is desired, we should not OOM

core/src/main/java/org/apache/cxf/helpers/IOUtils.java
```diff
@@ -34,10 +34,14 @@
 import java.nio.file.Files;
 import java.util.Objects;
 
+import org.apache.cxf.common.util.SystemPropertyAction;
 import org.apache.cxf.io.CopyingOutputStream;
 import org.apache.cxf.io.Transferable;
 
 public final class IOUtils {
+    public static final int DEFAULT_BINARY_MAX_SIZE =
```
coheigea, line 42: Just a nit that it's probably not accurate to call this "DEFAULT" if it can be modified by the user via a system property.

coheigea (approved):

core/src/main/java/org/apache/cxf/helpers/IOUtils.java
reta: Thanks @coheigea , happy to change, but followed our conventions from https://github.com/apache/cxf/blob/main/core/src/main/java/org/apache/cxf/attachment/AttachmentDeserializer.java#L64 (and other places). Probably to clarify why default is applicable: that would be a value picked up by default if `maxSize` is not specified by the user. Yes, the value could be changed, but it will still be semantically the default unless specified. Thanks!

coheigea: OK no problem, please ignore then.

PR #3304, opened 2026-07-10
Filter claims by granted scopes for the IdToken case

coheigea:


rt/rs/security/sso/oidc/src/main/java/org/apache/cxf/rs/security/oidc/idp/UserInfoService.java
```diff
@@ -98,35 +99,40 @@ protected Object convertUserInfoToResponseEntity(UserInfo userInfo) {
...
+    protected UserInfo createFromIdToken(IdToken idToken, List<String> scopes) {
         UserInfo userInfo = new UserInfo();
         userInfo.setSubject(idToken.getSubject());
 
         if (super.isJwsRequired()) {
             userInfo.setIssuer(idToken.getIssuer());
             userInfo.setAudience(idToken.getAudience());
         }
-        if (idToken.getPreferredUserName() != null) {
+        if (scopes.contains(OidcUtils.PROFILE_SCOPE) && idToken.getPreferredUserName() != null) {
```
reta: Nit, we could probably move `scopes.contains(OidcUtils.PROFILE_SCOPE)` to own `if`:
```
if (scopes.contains(OidcUtils.PROFILE_SCOPE)) {
  if ( idToken.getPreferredUserName() != null) {
     userInfo.setPreferredUserName(idToken.getPreferredUserName());
  }
  ...
}
```

rt/rs/security/sso/oidc/src/main/java/org/apache/cxf/rs/security/oidc/idp/UserInfoService.java
```diff
@@ -136,6 +142,22 @@ protected UserInfo createFromIdToken(IdToken idToken) {
...
+        if (OidcUtils.EMAIL_CLAIMS.contains(claimName)) {
+            return scopes.contains(OidcUtils.EMAIL_SCOPE);
+        }
+        if (OidcUtils.ADDRESS_CLAIMS.contains(claimName)) {
+            return scopes.contains(OidcUtils.ADDRESS_SCOPE);
+        }
+        if (OidcUtils.PHONE_CLAIMS.contains(claimName)) {
+            return scopes.contains(OidcUtils.PHONE_SCOPE);
+        }
+        return true;
```
reta, line 160: Shouldn't it be `false` by default?  Or we really don't care what those claims are

reta (approved):

rt/rs/security/sso/oidc/src/main/java/org/apache/cxf/rs/security/oidc/idp/UserInfoService.java
coheigea: Good point, I'll fix this.

rt/rs/security/sso/oidc/src/main/java/org/apache/cxf/rs/security/oidc/idp/UserInfoService.java
coheigea: I think we're just checking the standard profile/email/etc claims here so it can default to true.

PR #3311, opened 2026-07-14
Make sure MessageContextImpl respects the limit to the maximum number of attachment to be collected

reta:
Make sure MessageContextImpl respects the limit to the maximum number of attachments to be collected

### DIFF
--- a/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/ext/MessageContextImpl.java
+++ b/rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/ext/MessageContextImpl.java
@@ -273,6 +273,8 @@ public class MessageContextImpl implements MessageContext {
                 m.getExchange().getInMessage().get(AttachmentDeserializer.ATTACHMENT_MAX_SIZE));
             inMessage.put(AttachmentDeserializer.ATTACHMENT_MAX_HEADER_SIZE,
                 m.getExchange().getInMessage().get(AttachmentDeserializer.ATTACHMENT_MAX_HEADER_SIZE));
+            inMessage.put(AttachmentDeserializer.ATTACHMENT_MAX_COUNT,
+                    m.getExchange().getInMessage().get(AttachmentDeserializer.ATTACHMENT_MAX_COUNT));
             inMessage.setContent(InputStream.class,
                 m.getExchange().getInMessage().get("org.apache.cxf.multipart.embedded.input"));
             inMessage.put(Message.CONTENT_TYPE,
@@ -281,6 +283,16 @@ public class MessageContextImpl implements MessageContext {
 
 
         new AttachmentInputInterceptor().handleMessage(inMessage);
+
+        final Object maxCountProperty = inMessage.getContextualProperty(AttachmentDeserializer.ATTACHMENT_MAX_COUNT);
+        int maxAttachmentCount = AttachmentDeserializer.DEFAULT_ATTACHMENT_MAX_COUNT;
+        if (maxCountProperty != null) {
+            if (maxCountProperty instanceof Integer) {
+                maxAttachmentCount = (Integer)maxCountProperty;
+            } else {
+                maxAttachmentCount = Integer.parseInt((String)maxCountProperty);
+            }
+        }
 
         List<Attachment> newAttachments = new LinkedList<>();
         try {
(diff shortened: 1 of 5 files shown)


coheigea (approved):

coheigea:
Just a reminder please @reta to backport to 3.6

reta:
> Just a reminder please @reta to backport to 3.6

Will do shortly, the CI was broken over the weekend, I didn't stackup changes

rt/frontend/jaxrs/src/test/java/org/apache/cxf/jaxrs/provider/EntityPartProviderTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+        // Add many attachments
+        IntStream.range(0, 40).forEach(i -> {
+            sb.append("------=_Part_34950_1098328613.1263781527359\n")
+                .append("Content-Type: text/xml\n")
+                .append("Content-Transfer-Encoding: binary\n")
+                .append("Content-Id: <b86a5f2d-e7af-4e5e-b71a-9f6f2307cab0>\n")
+                .append('\n')
+                .append("<message>\n")
+                .append("------=_Part_34950_1098328613.1263781527359--\n");
```
rnbwdsh, line 99: A `--` AFTER a boundary should be the end-of-messageboundary, everything after that should be considered the epilogue that should be ignored according to https://datatracker.ietf.org/doc/html/rfc2046 - so this should actually have 2 parts?

rt/frontend/jaxrs/src/test/java/org/apache/cxf/jaxrs/provider/EntityPartProviderTest.java
```diff
@@ -0,0 +1,170 @@
...
+            .append("------=_Part_34950_1098328613.1263781527359\n")
+            .append("Content-Type: text/xml; charset=UTF-8\n")
+            .append("Content-Transfer-Encoding: binary\n")
+            .append("Content-Id: <318731183421.1263781527359.IBM.WEBSERVICES@auhpap02>\n")
+            .append('\n')
+            .append("<envelope/>\n");
+
+        // Add many attachments
+        IntStream.range(0, 40).forEach(i -> {
+            sb.append("------=_Part_34950_1098328613.1263781527359\n")
```
rnbwdsh, line 93: Generally, the standard suggests CRLF as terminators, and optionally allows `\r` here, but this tests would break if cxf would ever implement strict parsing. I think the more leniant parsing is an active choice, I still think the tests should produce RFC complaint inputs, otherwise fixing this in the future will break things.

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/ext/MessageContextImpl.java
```diff
@@ -282,6 +284,16 @@ private MultipartBody createAttachments(String propertyName) {
 
         new AttachmentInputInterceptor().handleMessage(inMessage);
 
+        final Object maxCountProperty = inMessage.getContextualProperty(AttachmentDeserializer.ATTACHMENT_MAX_COUNT);
+        int maxAttachmentCount = AttachmentDeserializer.DEFAULT_ATTACHMENT_MAX_COUNT;
+        if (maxCountProperty != null) {
+            if (maxCountProperty instanceof Integer) {
```
rnbwdsh, line 290: instanceof is also a null check -> i'd just do
            if (maxCountProperty instanceof Integer) {
...
} else if (maxCountProperty instanceof String) {
...
} and remove the outer != null.

rt/frontend/jaxrs/src/test/java/org/apache/cxf/jaxrs/provider/EntityPartProviderTest.java
reta: At least for now, CXF treats `--` as the end of boundary (not the end-of-messageboundary), which is indeed delimits as last part (as per rfc2046), thank you would pointing it out,

rt/frontend/jaxrs/src/main/java/org/apache/cxf/jaxrs/ext/MessageContextImpl.java
reta: This is above our JDK baseline

PR #3331, opened 2026-07-23
CXF-9223 - Improve prefix matching for OAuth audience URIs

coheigea:


rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/filters/OAuthRequestFilter.java
```diff
@@ -295,6 +297,28 @@ protected String validateAudiences(List<String> audiences) {
...
+     * or is followed by a URL boundary character ('/', '?', '#').
+     */
+    protected boolean matchesAudiencePrefix(String requestPath, String configuredAudience) {
+        if (requestPath == null || configuredAudience == null) {
+            return false;
+        }
+        if (!requestPath.startsWith(configuredAudience)) {
+            return false;
+        }
+        if (requestPath.length() == configuredAudience.length() || configuredAudience.endsWith("/")) {
```
reta, line 315: I believe this specific condition is redundant;
```
        if (requestPath.length() == configuredAudience.length()) {
```

it would match `requestPath.startsWith(configuredAudience)` exactly, right?

reta, line 315:
```suggestion
        if (configuredAudience.endsWith("/")) {
```

coheigea: I don't think it is redundant. The startsWith check makes sure that the audience matches the start of the request URI. Then the next check is to make sure either they're equal or if it ends with a forward slash then we can accept the match.

reta: Oh I see, basically `== equals` part

reta (approved):

PR #3343, opened 2026-07-27
[CXF-9233] Ghost RESP_OUT log

vp340:
Pull Request related to Jira: [CXF-9233] AbstractLoggingInterceptor.LIVE_LOGGING_PROP already set in the message properties, so RESP_OUT log is disable :(

### DIFF
--- a/rt/features/logging/src/main/java/org/apache/cxf/ext/logging/AbstractLoggingInterceptor.java
+++ b/rt/features/logging/src/main/java/org/apache/cxf/ext/logging/AbstractLoggingInterceptor.java
@@ -37,7 +37,7 @@ public abstract class AbstractLoggingInterceptor extends AbstractPhaseIntercepto
     public static final int DEFAULT_LIMIT = 48 * 1024;
     public static final int DEFAULT_THRESHOLD = -1;
     public static final String CONTENT_SUPPRESSED = "--- Content suppressed ---";
-    protected static final String  LIVE_LOGGING_PROP = "org.apache.cxf.logging.enable."; // the EventType (flow) will be concatenated
+    protected static final String  LIVE_LOGGING_PROP = "org.apache.cxf.logging.enable."; // the EventType (flow) and ExchangeId will be concatenated
     private static final Pattern BOUNDARY_PATTERN =
         Pattern.compile("^--(\\S*)$", Pattern.MULTILINE);
     private static final Pattern CONTENT_TYPE_PATTERN =
@@ -63,15 +63,21 @@ public abstract class AbstractLoggingInterceptor extends AbstractPhaseIntercepto
 
     // The concatenated flow is added in order to enhance resilience against misuse and underlying framework
     // (Reuse of the same Message object with properties still there)
-    // The message will be logged once per flow
-    // If previous properties (ex. LIVE_LOGGING_PROP + REQ_IN) are still there... this will search
-    // only for the right properties (ex. LIVE_LOGGING_PROP + RESP_OUT)
+    // The message will be logged once per flow per ExchangeId
+    // If previous properties (ex. LIVE_LOGGING_PROP + REQ_IN + ExchangeId) are still there... this will search
+    // only for the right properties (ex. LIVE_LOGGING_PROP + RESP_OUT + ExchangeId)
     protected boolean isLoggingDisabledForThisFlow(Message message) throws Fault {
-        Object liveLoggingProp = message.getContextualProperty(LIVE_LOGGING_PROP + normalizeFlow(eventMapper.getEventType(message))); //idempotency per Flow
+        Object liveLoggingProp = message.getContextualProperty(getDisableLogKey(message)); //idempotency per Flow per ExchangeId
         return liveLoggingProp != null && PropertyUtils.isFalse(liveLoggingProp);
     }
     protected void disableFutureLoggingForThisFlow(Message message) throws Fault {
-        message.put(LIVE_LOGGING_PROP + normalizeFlow(eventMapper.getEventType(message)), Boolean.FALSE);
+        message.put(getDisableLogKey(message), Boolean.FALSE);
+    }
+
+    // LIVE_LOGGING_PROP + FLOW + ExchangeId
+    protected String getDisableLogKey(Message message){
+        createExchangeId(message); //Redundant
+        return LIVE_LOGGING_PROP + normalizeFlow(eventMapper.getEventType(message)) + '.' + message.getExchange().get(LogEvent.KEY_EXCHANGE_ID);
     }
 
     public void addBinaryContentMediaTypes(String mediaTypes) {
--- a/rt/features/logging/src/main/java/org/apache/cxf/ext/logging/AbstractLoggingInterceptor.java
+++ b/rt/features/logging/src/main/java/org/apache/cxf/ext/logging/AbstractLoggingInterceptor.java
@@ -25,20 +25,19 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.apache.cxf.common.util.PropertyUtils;
-import org.apache.cxf.ext.logging.event.DefaultLogEventMapper;
-import org.apache.cxf.ext.logging.event.LogEvent;
-import org.apache.cxf.ext.logging.event.LogEventSender;
-import org.apache.cxf.ext.logging.event.PrettyLoggingFilter;
+import org.apache.cxf.ext.logging.event.*;
 import org.apache.cxf.interceptor.Fault;
 import org.apache.cxf.message.Exchange;
 import org.apache.cxf.message.Message;
 import org.apache.cxf.phase.AbstractPhaseInterceptor;
 
+import static org.apache.cxf.ext.logging.event.DefaultLogEventMapper.normalizeFlow;
+
 public abstract class AbstractLoggingInterceptor extends AbstractPhaseInterceptor<Message> {
     public static final int DEFAULT_LIMIT = 48 * 1024;
     public static final int DEFAULT_THRESHOLD = -1;
     public static final String CONTENT_SUPPRESSED = "--- Content suppressed ---";
-    protected static final String  LIVE_LOGGING_PROP = "org.apache.cxf.logging.enable";
+    protected static final String  LIVE_LOGGING_PROP = "org.apache.cxf.logging.enable."; // the EventType (flow) will be concatenated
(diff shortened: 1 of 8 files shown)


rt/features/logging/src/main/java/org/apache/cxf/ext/logging/AbstractLoggingInterceptor.java
```diff
@@ -25,20 +25,21 @@
...
 import org.apache.cxf.phase.AbstractPhaseInterceptor;
 
+import static org.apache.cxf.ext.logging.event.DefaultLogEventMapper.normalizeFlow;
+
 public abstract class AbstractLoggingInterceptor extends AbstractPhaseInterceptor<Message> {
     public static final int DEFAULT_LIMIT = 48 * 1024;
     public static final int DEFAULT_THRESHOLD = -1;
     public static final String CONTENT_SUPPRESSED = "--- Content suppressed ---";
     protected static final String  LIVE_LOGGING_PROP = "org.apache.cxf.logging.enable";
+    protected static final String IDEMPOTENT_LOGGING_PROP = "org.apache.cxf.idempotent.logging."; // the EventType (flow) and ExchangeId will be concatenated
```
reta, line 41: Thanks for the pull request @vp340 , I would advice against introducing yet another property, not only it becomes very confusing, it also difficult to figured out where all these different properties are coming from. I will try to spend some time looking into the problem, if you could attach a simple reproducer to the JIRA ticket, that would be great to understand the issue in context. Thank you.

vp340: Hi @reta ,
thank you for the reply!

I added as you requested an example project https://github.com/vp340/cxf-log-example to the JIRA ticket where I develop a simple ExampleService that simulate the error-prone situation.
I added in the last JIRA comment a more detailed explanation. :) 
If you have any problem to run it locally let me know and I'll try to help you. (I'm currently on vacation, but I will try to reply asap :D )

I prepared wiremock configuration and a soapUI project (or if U prefer the endpoint and the raw request) .

If U go to src/main/resources/spring/example/v1/route-context.xml  ... and uncomment the processor U can make the RESP_OUT log reapper as I described in the JIRA ticket.

As i wrote in the comment, I undestand that adding  the IDEMPOTENT_LOGGING_PROP can be "confusing", but so is not finding the RESP_OUT log  because a generic property has already been set somewhere else and the framework propagates it, if U don't manually intervene .
My goal with the IDEMPOTENT_LOGGING_PROP  was to fullfill the use case "not log twice" without using the same property used to disable completely the log from the Bus (and that can lead to these sneaky situations ) .
In my project it worked fine without adding manual processor. 

Hope it helps. Keep me updated :)

reta: Thanks a lot @vp340, I am off this week, will surely pick it up when I am back. My apologies,  thank you

reta: @vp340 just to let you know - was able to reproduce the issue , thanks to the sample project and instructions, haven't figured out the flow / cause yet, but working on it.

vp340: Hi @reta,
Thanks for keeping me updated.
As I wrote somewhere in the jira ticket or in some comment in the demo project I presume that the "ResponseContext" header is  the culprit along with the Camel CxfProducer/CxfConsumer that set/reset It. If I were U I would look that way first. 

Now in my company I have bypassed the issue using the solution that I posted in the PR and It seems to work fine, but clearly there Is something more Camel related stuff that Is interfering :/ ...and which requires further exploration.

Let me know if U need anything. If I have time I'll try to investigate something myself.

Let's keep each other updated. :) 

Have a good job!

Valentino Porta

reta: @vp340 prototyped a fix here https://github.com/apache/cxf/pull/3372 , I still have to add test case(s) for it, I am wondering if you could give it a shot, thanks

vp340: Hi thanks. 
I was updating the Jira ticket right now.  (sorry I live in Italy and here is almost 2 a.m right now and tomorrow I have work) .
Yeah the #3372 could be an easier and more understandable solution. 
If I understand well the isRequestor() method is the same use to retrieve the EventType for logging ..so isRequestor ? EventType.RESP_IN : EventType.REQ_IN 
It will set 'LIVE_LOGGING_PROP + true' if it's a client and a 'LIVE_LOGGING_PROP + false' if is a server. 

The only doubt situation could be 2 backend call in a row... 
So the flow would be:
REQ_IN (set LIVE_LOGGING_PROP + false but not propagated)
REQ_OUT (set LIVE_LOGGING_PROP + true but not propagated)
RESP_IN (set LIVE_LOGGING_PROP + true AND propagated!)
REQ_OUT (find LIVE_LOGGING_PROP + true ...so ghost REQ_OUT logging)
RESP_IN 
RESP_OUT

(This is a situation that I thought right know... it should be tested)...

Another possible problem is that if we modify the property name itself, someone who had already set that property on the Bus to completely disable the logging would no longer be able to do so. This could therefore break backward compatibility. (I read something related to this in https://issues.apache.org/jira/browse/CXF-7000 )


If U are interested in reviewing another possible solution, going deeply in debug in the sample project I believe I found the very place where the LIVE_LOGGING_PROP is set and propagated. 
I open a new https://github.com/apache/cxf/pull/3373 

In the class org.apache.cxf.endpoint.ClientImpl ... processResult method
Message inMsg = exchange.getInMessage();
if (inMsg != null) {
if (null != resContext)
{ 
resContext.putAll(inMsg); 
// remove the recursive reference if present 
resContext.remove(Message.INVOCATION_CONTEXT); 
// remove the logging disable property //ADDED 
resContext.remove(Message.LIVE_LOGGING_PROP); //ADDED 
setResponseContext(resContext); 
}

Here someone already remove a Message.INVOCATION_CONTEXT property from the Response Context ... so I think It could be a good place to prevent that the logging properties is propagated in the ResponseContext at all!

To do so I had to transfer the String Constant  from the AbstractLoggingInterceptor to the Message. 
This way we won't change the behaviour at all. 
(I'm very confident that  it works... at least in debug I removed it manually and the RESP_OUT log reapperead... so the property was not propagated).

Let me know what U think about it. 
And if you have any suggestions, especially regarding design patterns or the overall architecture, please feel free to share them. If you think there’s a better way to approach it, I’d be more than happy to hear it...I have a lot to learn from you!

Have a great work!

Valentino Porta

reta: > I was updating the Jira ticket right now. (sorry I live in Italy and here is almost 2 a.m right now and tomorrow I have work) .

Thanks @vp340 , np at all

> If I understand well the isRequestor() method is the same use to retrieve the EventType for logging ..so isRequestor ? EventType.RESP_IN : EventType.REQ_IN
It will set 'LIVE_LOGGING_PROP + true' if it's a client and a 'LIVE_LOGGING_PROP + false' if is a server.

This is correct (in the nutshell) just benefiting from CXF message handling logic

> The only doubt situation could be 2 backend call in a row...

This should have different message instance created per call, the message should not be reused 

> I open a new https://github.com/apache/cxf/pull/3373

This is possible but not the best option: the framework core (Client / Message) knows nothing about properties that are specific to custom interceptors. The tracking and decision making has to be done within in/out logging interceptors. 

> Another possible problem is that if we modify the property name itself, someone who had already set that property on the Bus to completely disable the logging would no longer be able to do so. This could therefore break backward compatibility. (I read something related to this in https://issues.apache.org/jira/browse/CXF-7000 )

There is risk of that but the property is intentionally not public (it is protected), so if someone relies on implementation details - we could not guarantee that it will always work for non-public APIs.

vp340: Hi @reta,
Thanks again for your precious time.
I understand... The core module should in fact be discern from the logging one.
Now I have some days off that I Will spend in a chalet (a poor version :/ ...not the Rich one ) and I won't be assure to have internet access all the time. 
If U don't mind waiting for me, next weekend when I came back I could test your solution against some of my Company projects or simulate some strange situations in the sample project and share the results.
If I have time I would like also to improve my solution as well and try to decouple the core module from the logging one. (As a personal exercise).
Let me know what U think!
Good weekend!

reta: Thank you @vp340 , please take as much time as you need, no pressure here, Have a great weekend and days off

vp340: Hi @reta , I'm back .
I tested the case with more cxf clients in a row (here the example https://github.com/vp340/cxf-log-example/tree/test_more_cxf_client ) . The properties from ResponseContext set in RESP_IN are propagated only in RESP_OUT and not to other REQ_OUT/RESP_IN (As expected, but I wanted to be sure) . 
So your solution to add MessageUtils.isRequestor(message) to the LIVE_LOGGING_PROP should do the job!  :) 

The only inaccuracy remains that the properties is still propagated across the RESP_IN / RESP_OUT Messages (even if it's ignored). 

If U are interested I also tried to fix my solution decoupling the core module from the logging module. 
Now I used in the core module a Set of property field to remove from the ResponseContext Map when the ClientImpl populate the map from the InMessage. Then I exposed a method to add to that set other properties to exclude. 
So now the logging module can say to the core module: "I put the LIVE_LOGGING_PROP into the Message... but be careful not to propagate it into the ResponseContext". (this could be apply also for other modules in the future if someone call Message.put(...) but doesn't want to propagate the property into the ResponseContext). 
As from the previous analisys it's only an IN problem... I set the adding of LIVE_LOGGING_PROP only in the LoggingInInterceptor constructor. 
I would appreciate if U could take a look and maybe give me your opinion (If U see other problem or suggestion about the stile of the code). (https://github.com/apache/cxf/pull/3373/ ) 

Thanks in advance.
Have a great day!

Valentino Porta

PR #3348, opened 2026-07-29
Add a lock when revoking tokens

coheigea:


rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/provider/AbstractOAuthDataProvider.java
```diff
@@ -271,6 +271,17 @@ public void revokeToken(Client client, String tokenKey, String tokenTypeHint) th
     @Override
     public void revokeToken(Client client, UserSubject callerSubject, String tokenKey, String tokenTypeHint)
         throws OAuthServiceException {
+        if (!recycleRefreshTokens) {
```
reta, line 274: Nit, may be a bit better to read the flow, wdyt? 

```suggestion
        if (!recycleRefreshTokens) {
            synchronized (refreshTokenLock) {
                doRevokeToken(client, callerSubject, tokenKey, tokenTypeHint);
            }
        } else {
            doRevokeToken(client, callerSubject, tokenKey, tokenTypeHint);
        }
```

reta (approved):

PR #3349, opened 2026-07-31
Fix flaky SSE system tests

coheigea:
The Tomcat SSE tests almost always fail on my machine, the Undertow tests sporadically. There are two changes:

- moved the close-state check to happen after the server actually closes the connection.
- fixed the last event counter so it only increments when an event was unexpectedly delivered, not when failure was expected after close.


reta (commented):
Thanks @coheigea !

PR #3350, opened 2026-07-31
[CXF-9234]Concurrent DynamicClientFactory.createClient for the same W…

ffang:
…SDL url corrupts the cached schema DOM and spins forever at 100% CPU

### DIFF
--- a/rt/frontend/simple/src/main/java/org/apache/cxf/endpoint/dynamic/DynamicClientFactory.java
+++ b/rt/frontend/simple/src/main/java/org/apache/cxf/endpoint/dynamic/DynamicClientFactory.java
@@ -979,7 +979,15 @@ public class DynamicClientFactory {
         int type = node.getNodeType();
 
         if (node.getOwnerDocument() == document) {
-            return node.cloneNode(deep);
+            // The schema/WSDL Document backing this node may be cached and shared across
+            // concurrent createClient() calls (see WSDLManagerImpl). The native cloneNode(true)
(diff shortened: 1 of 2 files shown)


rt/frontend/simple/src/main/java/org/apache/cxf/endpoint/dynamic/DynamicClientFactory.java
```diff
@@ -979,7 +979,15 @@ public Node cloneNode(Document document, Node node, boolean deep) throws DOMExce
...
 
         if (node.getOwnerDocument() == document) {
-            return node.cloneNode(deep);
+            // The schema/WSDL Document backing this node may be cached and shared across
+            // concurrent createClient() calls (see WSDLManagerImpl). The native cloneNode(true)
+            // walk triggers UserDataHandler callbacks (e.g. StaxUtils$LocationUserDataHandler)
+            // that mutate the source document's internal userData table, which is not
+            // thread-safe (WeakHashMap in Xerces-J, HashMap in the JDK DOM impl). Concurrent
+            // clones of the same shared document must therefore be serialized (CXF-9234).
+            synchronized (document) {
```
reta, line 988: Thanks @ffang , I am wondering - could we make this method  (`cloneNode`) private? It seems to be used internally only, so the `synchronized` will be enforced properly,

ffang: Hi @reta ，

Yes, revised accordingly.

Thanks!
Freeman

reta (approved):

PR #3358, opened 2026-08-07
Fix some atomicity issues in removeCodeGrant

coheigea:


rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/grants/code/JPACMTCodeDataProvider.java
```diff
@@ -129,12 +132,29 @@ protected RefreshToken updateExistingRefreshToken(RefreshToken rt, ServerAccessT
         return super.updateExistingRefreshToken(rt, at);
     }
 
+    @Override
+    protected ServerAuthorizationCodeGrant removeCodeGrant(String code, EntityManager em,
```
reta, line 141: I believe the check + removal should happen in the same transaction scope (`executeInTransaction`), fe doRemoveClient / removeClientCodeGrants are good references

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/grants/code/JPACMTCodeDataProvider.java
```diff
@@ -129,12 +132,37 @@ protected RefreshToken updateExistingRefreshToken(RefreshToken rt, ServerAccessT
         return super.updateExistingRefreshToken(rt, at);
     }
 
+    @Override
+    public ServerAuthorizationCodeGrant removeCodeGrant(final String code) throws OAuthServiceException {
+        return executeInTransaction(em -> findAndRemoveWithLock(code, em));
```
reta, line 137: 👍

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/grants/code/JPACMTCodeDataProvider.java
```diff
@@ -129,12 +132,37 @@ protected RefreshToken updateExistingRefreshToken(RefreshToken rt, ServerAccessT
...
+    }
+
+    @Override
+    protected ServerAuthorizationCodeGrant removeCodeGrant(String code, EntityManager em,
+                                                           LockModeType lockModeType) throws OAuthServiceException {
+        return executeInTransaction(innerEm -> findAndRemoveWithLock(code, innerEm));
```
reta: This one could be problematic, could we introduce `executeInTransaction(executeInTransaction(EntityManager em, EntityManagerOperation<T> operation)` here so the same entity manager is picked up? The issue of using another one is that the operation could be a part of larger transaction, and if the instance of the   entity manager is not preserved, it will have unintended effects

coheigea: Yes I removed that one, the method is only ever called by the subclass which is already in a transaction.

reta (approved):

PR #3370, opened 2026-08-13
CXF-9237 - Add possibility to configure JCache expiry time

coheigea:
By default there is no eviction for backwards compatibility

### DIFF
--- a/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/provider/JCacheOAuthDataProvider.java
+++ b/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/provider/JCacheOAuthDataProvider.java
@@ -24,11 +24,14 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 
 import javax.cache.Cache;
 import javax.cache.CacheManager;
 import javax.cache.Caching;
 import javax.cache.configuration.MutableConfiguration;
+import javax.cache.expiry.CreatedExpiryPolicy;
+import javax.cache.expiry.Duration;
 import javax.cache.spi.CachingProvider;
 
 import org.apache.cxf.Bus;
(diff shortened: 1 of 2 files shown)


rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/provider/JCacheOAuthDataProvider.java
```diff
@@ -86,18 +96,65 @@ public JCacheOAuthDataProvider(String configFileURL,
...
                                    String refreshTokenCacheKey,
                                    boolean storeJwtTokenKeyOnly) {
+        this(configFileURL, bus, clientCacheKey, accessTokenCacheKey, refreshTokenCacheKey,
+             storeJwtTokenKeyOnly, CacheTTLs.ETERNAL);
+    }
+
+    // cacheTTLs lets the access/refresh token caches be evicted by the JCache infrastructure
+    // itself once entries age out, independently of any application-level expiry check;
+    // a TTL <= 0 (CacheTTLs.ETERNAL by default) leaves the corresponding cache eternal
+    public JCacheOAuthDataProvider(String configFileURL,
```
reta, line 106: @coheigea thanks for the change, what concerns me is that `cacheTTLs` gives a wrong impression that TTL will be configured for each cache, but that is not the case: if the configuration has dedicated cache preset (like below)

```xml
  <cache alias="cxf.oauth2.accesstoken.cache">
    <key-type>java.lang.String</key-type>
    <value-type>org.apache.cxf.rs.security.oauth2.common.ServerAccessToken</value-type>
    <heap unit="entries">100</heap>
    <jsr107:mbeans enable-management="false" enable-statistics="false"/>
  </cache>
```

the cacheTTL will be effectively ignored. I don't want to complicate it but also looking for intuitive API usage, may be instead of `CacheTTLs cacheTTLs` we introduce something like `JCacheCreator` (or alike) that will be used to create a cache when the is no one in configuration? Wdyt?

reta: Something along these lines:

```java
interface JCacheCreator  {
    <K, V> Cache<K, V> createCache(CacheManager cacheManager, String cacheKey, MutableConfiguration<K, V> config);
}
```

coheigea: Thanks @reta , maybe in that case we don't need this improvement at all if the user can configure the TTL in the config?

reta: Thanks @coheigea , that is also a valid point, we may not even need it

PR #3372, opened 2026-08-13
CXF-9233: AbstractLoggingInterceptor.LIVE_LOGGING_PROP already set in the message properties, so RESP_OUT is not logged

reta:
It turned out that `ClientImpl::onMessage` may create in-message (response) from the out message (request), copying the contextual properties but changing the message requestor role. That would lead to logging interceptor to mistakenly mark that the message as already logged.


coheigea:
@reta One compatibility concern is that the change no longer checks the original unsuffixed org.apache.cxf.logging.enable property, so existing applications using that property to disable logging may be affected. Is there any way we can preserve backwards compatibility, i.e. by allowing the old property if it's specified by users, but if not use the new approach?

reta:
> @reta One compatibility concern is that the change no longer checks the original unsuffixed org.apache.cxf.logging.enable property, so existing applications using that property to disable logging may be affected. Is there any way we can preserve backwards compatibility, i.e. by allowing the old property if it's specified by users, but if not use the new approach?

Thanks @coheigea , yeah, sure, we could do that, will push the changes shortly

coheigea (approved):

PR #3375, opened 2026-08-14
fix Invocation.Builder.property() not visible to HTTP transport

neenapj:
Fixes [CXF-9235](https://issues.apache.org/jira/browse/CXF-9235)

Made changes to` InvocationBuilderImpl.property()` to write directly into contextProps, matching how Client/WebTarget already behave.

**Tests added (InvocationBuilderImplTest)**

`testPropertyWrittenToBothFilterPropsAndFlatContext` sets `SET_EMPTY_REQUEST_CT_PROPERTY = false` via `Invocation.Builder.property()` and asserts two storage locations on `ClientConfiguration.getRequestContext()`:

(a) filterProps sub-map (`requestContext.get("jaxrs.filter.properties")`): consumed by `ClientRequestFilter.getProperty()` via `MessagePropertyHolder`. This path was always written; the test guards it stays working.
(b) flat top-level map (`requestContext.get("set.content.type.for.empty.request")`): consumed by the HTTP transport via `Message.getContextualProperty()` in `Headers.setProtocolHeadersInConnection()`. This was the CXF-9235 bug — only (a) was written, so the transport silently ignored the property. The test fails on (b) when the fix is reverted (expected:<false> but was:<null>), and passes once the flat write is restored. The test also verifies that passing null removes the property from both locations.
No network call is made — the test inspects the ClientConfiguration map directly, since `InvocationBuilderImpl.property() `populates it synchronously before any dispatch.

### DIFF
--- a/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/spec/InvocationBuilderImpl.java
+++ b/rt/rs/client/src/main/java/org/apache/cxf/jaxrs/client/spec/InvocationBuilderImpl.java
@@ -288,8 +288,12 @@ public class InvocationBuilderImpl implements Invocation.Builder {
         }
         if (value == null) {
             filterProps.remove(name);
+            contextProps.remove(name);
         } else {
             filterProps.put(name, value);
+            // CXF-9235: also write to the top-level context so that the transport layer
+            // can find the property via Message.getContextualProperty()
+            contextProps.put(name, value);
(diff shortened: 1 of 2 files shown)


neenapj:
@ffang Could you please take a look at this PR when you get a chance?

ffang (approved):
Thanks for the contribution @neenapj !

PR #3382, opened 2026-08-20
Introduce a TTL in the AbstractAccessTokenValidator cache

coheigea:


rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/services/AbstractAccessTokenValidator.java
```diff
@@ -105,8 +116,16 @@ protected AccessTokenValidation getAccessTokenValidation(String authScheme, Stri
         }
 
         AccessTokenValidation accessTokenV = null;
+        AccessTokenValidationCacheEntry cacheEntry = null;
         if (maxValidationDataCacheSize > 0) {
-            accessTokenV = accessTokenValidations.get(authSchemeData);
+            cacheEntry = accessTokenValidations.get(authSchemeData);
+            if (cacheEntry != null) {
+                if (isValidationDataCacheEntryExpired(cacheEntry)) {
```
reta, line 123: Are we concerned about the race here: new token gets added, replacing the existing one, while we are checking the expiration?

reta (approved):

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/services/AbstractAccessTokenValidator.java
coheigea: Good catch - I replaced the put with putIfAbsent. I think on the reading side it's not something we need to worry about.

PR #3394, opened 2026-08-24
Fix deflate bomb in compression utils

coheigea:


core/src/main/java/org/apache/cxf/common/util/CompressionUtils.java
```diff
@@ -26,6 +26,10 @@
 import java.util.zip.Inflater;
 
 public final class CompressionUtils {
+    // Guards against decompression-bomb style inputs that expand to an unbounded size on the heap
+    public static final long DEFAULT_MAX_INFLATED_SIZE =
```
reta, line 30: Wondering if 10Mb is a bit small, may be we could align with AttachmentDeserializer#DEFAULT_ATTACHMENT_MAX_SIZE, which is 50Mb now?

coheigea: I think it's only used by JWT and SAML though, I don't see any scenarios where it would be more than 10 mb?

reta (approved):

PR #3405, opened 2026-08-27
Use a free port not 8080 for SeBootstrapTest

coheigea:
I noticed this test was failing when I had Tomcat running locally.


reta (commented):
Thanks @coheigea !

reta (approved):

PR #3408, opened 2026-08-28
Default Oauth2 encryption providers to use AES-GCM

coheigea:
Note: Only for application on main.

### DIFF
--- a/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/utils/crypto/ModelEncryptionSupport.java
+++ b/rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/utils/crypto/ModelEncryptionSupport.java
@@ -261,10 +261,15 @@ public final class ModelEncryptionSupport {
         byte[] iv = CryptoUtils.generateSecureRandomBytes(DEFAULT_IV_SIZE);
         byte[] cipherText = CryptoUtils.encryptBytes(sequence.getBytes(StandardCharsets.UTF_8),
                                                      secretKey, gcmProperties(iv));
-        byte[] out = new byte[iv.length + cipherText.length];
-        System.arraycopy(iv, 0, out, 0, iv.length);
-        System.arraycopy(cipherText, 0, out, iv.length, cipherText.length);
-        return CryptoUtils.encodeBytes(out);
+        try {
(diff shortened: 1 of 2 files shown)


rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/utils/crypto/ModelEncryptionSupport.java
```diff
@@ -229,10 +242,55 @@ public static ServerAuthorizationCodeGrant decryptCodeGrant(OAuthDataProvider pr
...
+     * a client holding several tokens could cut and splice block-by-block to
+     * forge token state; pass explicit KeyProperties only if a legacy
+     * transformation must be retained.
+     */
+    private static String encryptSequence(String sequence, Key secretKey,
+                                          KeyProperties props) throws SecurityException {
+        if (props != null) {
+            return CryptoUtils.encryptSequence(sequence, secretKey, props);
+        }
+        byte[] iv = new byte[DEFAULT_IV_SIZE];
```
reta:
```suggestion
        byte[] iv = CryptoUtils.generateSecureRandomBytes(DEFAULT_IV_SIZE);
```

rt/rs/security/oauth-parent/oauth2/src/main/java/org/apache/cxf/rs/security/oauth2/utils/crypto/ModelEncryptionSupport.java
```diff
@@ -46,6 +49,12 @@
  */
 public final class ModelEncryptionSupport {
     public static final String SEP = "|";
+
+    private static final String DEFAULT_TRANSFORMATION = "AES/GCM/NoPadding";
+    private static final int DEFAULT_IV_SIZE = 12;
+    private static final int DEFAULT_AUTH_TAG_LENGTH = 128;
+    private static final SecureRandom RANDOM = new SecureRandom();
```
reta: We probably don't need it and could use CryptoUtils?

coheigea: Thanks @reta , I applied that change and did a little cleanup

reta (approved):

PR #3410, opened 2026-08-28
Fix Tika 4 detector context handling

skrcode:
Companion repair for #3409.

Tika 4 adds ParseContext to Detector.detect. This initializes or reuses the extraction context before media-type detection and passes the same context through detector, parser selection, and parsing. The focused regression test verifies that detection receives a context.

Verification (JDK 17):
- mvn -pl rt/rs/extensions/search -DskipTests compile — passed; Checkstyle and PMD included
- mvn -pl rt/rs/extensions/search -Dtest=TikaContentExtractorTest test — 7 tests passed; 0 failures, errors, or skips
- git diff --check — passed

JAIPilot Remote was also attempted on the exact source archive, but its environment could not resolve CXF build-utils 4.1.5-SNAPSHOT after repository.apache.org reset the connection, so no remote pass is claimed.

### DIFF
--- a/rt/rs/extensions/search/src/test/java/org/apache/cxf/jaxrs/ext/search/tika/TikaContentExtractorTest.java
+++ b/rt/rs/extensions/search/src/test/java/org/apache/cxf/jaxrs/ext/search/tika/TikaContentExtractorTest.java
@@ -18,12 +18,16 @@
  */
 package org.apache.cxf.jaxrs.ext.search.tika;
 
+import java.io.ByteArrayInputStream;
 import java.io.InputStream;
+import java.util.Collections;
 
 import org.apache.cxf.jaxrs.ext.search.SearchBean;
 import org.apache.cxf.jaxrs.ext.search.SearchCondition;
 import org.apache.cxf.jaxrs.ext.search.SearchConditionParser;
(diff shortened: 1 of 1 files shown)


reta:
@skrcode thanks for the change, there are few major issues to highlight:
 - in CXF, we try not to bring major dependency updates within minor releases 
 - however, if we have to, like in this case, we need to do that in a way old / new dependency work

The change in this direction landed here bc543bb989e4c881d54825f65c55846eae016501

skrcode:
Thanks, Andriy. #3409 was recreated after bc543bb989e4c881d54825f65c55846eae016501, so its current head already contains the TikaInputStream change. The remaining deterministic Tika 4 compilation failure is Detector.detect now requiring a ParseContext. This companion only initializes the existing ParseContext before detection, passes that same context through detection and parsing, and adds a focused regression test. That keeps the compatible TikaInputStream preparation on main while supplying the small signature adaptation on the Tika 4 upgrade branch. Happy to adjust if you prefer a different shape for the staged compatibility.

reta:
@skrcode I pushed a change into https://github.com/apache/cxf/pull/3409, we'll definitely benefit from the test, if you could rebase / merge., thank you

PR #3423, opened 2026-08-31
Use a properly random source to generate digest client nonces

coheigea:


reta (approved):

ppalaga:
Hi @coheigea, this is a bit problematic for Quarkus CXF in native mode, because we let GraalVM initialize as many classes as possible at build time for fast application boot. Doing this for a class wanting to init `java.util.Random` (or its subclasses) at build time is not a good idea for obvious reasons. Could I please send a change similar to https://github.com/apache/cxf/pull/2706/changes ?

ppalaga:
@coheigea I dared to send https://github.com/apache/cxf/pull/3452 and https://github.com/apache/cxf/pull/3451

PR #3435, opened 2026-09-02
Add a fallback to prevent attacks on RSA 1.5 in the Jose code

coheigea:


rt/rs/security/jose-parent/jose/src/main/java/org/apache/cxf/rs/security/jose/jwe/WrappedKeyDecryptionAlgorithm.java
```diff
@@ -41,22 +42,54 @@ public WrappedKeyDecryptionAlgorithm(Key cekDecryptionKey, KeyAlgorithm supporte
...
         KeyProperties keyProps = new KeyProperties(getKeyEncryptionAlgorithm(jweDecryptionInput));
         AlgorithmParameterSpec spec = getAlgorithmParameterSpec(jweDecryptionInput);
         if (spec != null) {
             keyProps.setAlgoSpec(spec);
         }
-        if (!unwrap) {
-            keyProps.setBlockSize(getKeyCipherBlockSize());
-            return CryptoUtils.decryptBytes(getEncryptedContentEncryptionKey(jweDecryptionInput),
-                                            getCekDecryptionKey(), keyProps);
+        byte[] fallbackCek = supportedAlgo == KeyAlgorithm.RSA1_5
```
reta, line 56: Sorry @coheigea , may be missing something here, but if `fallbackCek` is always random, how the decryption is going to work? The content is guaranteed to be encrypted with different key, right?

coheigea: @reta It's to prevent https://en.wikipedia.org/wiki/Adaptive_chosen-ciphertext_attack#Practical_attacks. By generating a random key and forcing it to be used for decryption we deny that knowledge back to the attacker.

reta: Makes perfect sense in this case, thanks @coheigea !

reta (approved):

PR #3443, opened 2026-09-04
Put a max length on the Fiql search term

coheigea:
…issue


rt/rs/extensions/search/src/main/java/org/apache/cxf/jaxrs/ext/search/fiql/FiqlParser.java
```diff
@@ -305,6 +338,40 @@ protected ASTNode<T> parseComparison(String expr) throws SearchParseException {
         throw new SearchParseException("Not a comparison expression: " + expr);
     }
 
+    private int[] findComparator(String expr) {
```
reta, line 341: @coheigea I do understand the intent here, but does it actually make sense to go such far (basically rewriting what regex is doing)? We've limited the expression length, I think the issue should be largely mitigated, right?

coheigea: @reta It's a fair point. I limited the search term even further to 4K characters by default and removed this custom code.

reta (approved):

PR #3458, opened 2026-09-09
Fix encrypted properties tests to work with WSS4J 4.0.2

coheigea:
The Jasypt algorithm changed, for now we'll just use the old encrypted values in the CXF tests until we pick up WSS4J 4.0.2 properly.

### DIFF
--- a/systests/ws-security/src/test/java/org/apache/cxf/systest/ws/action/ActionTest.java
+++ b/systests/ws-security/src/test/java/org/apache/cxf/systest/ws/action/ActionTest.java
@@ -246,25 +246,32 @@ public class ActionTest extends AbstractBusClientServerTestBase {
             return;
         }
 
-        SpringBusFactory bf = new SpringBusFactory();
-        URL busFile = ActionTest.class.getResource(
-                      JavaUtils.isFIPSEnabled() 
-                      ? "client-fips.xml" : "client.xml");
+        try {
+            if (!JavaUtils.isFIPSEnabled()) {
+                System.setProperty("org.apache.wss4j.crypto.jasypt.useLegacyDefaultAlgorithm", "true");
+            }
+            SpringBusFactory bf = new SpringBusFactory();
+            URL busFile = ActionTest.class.getResource(
+                        JavaUtils.isFIPSEnabled() 
+                        ? "client-fips.xml" : "client.xml");
 
-        Bus bus = bf.createBus(busFile.toString());
-        BusFactory.setDefaultBus(bus);
-        BusFactory.setThreadDefaultBus(bus);
+            Bus bus = bf.createBus(busFile.toString());
+            BusFactory.setDefaultBus(bus);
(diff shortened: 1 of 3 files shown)


systests/ws-security/src/test/java/org/apache/cxf/systest/ws/action/ActionTest.java
```diff
@@ -246,25 +246,32 @@ public void testEncryptedPassword() throws Exception {
...
-        QName portQName = new QName(NAMESPACE, "DoubleItEncryptedPasswordPort");
-        DoubleItPortType port =
-                service.getPort(portQName, DoubleItPortType.class);
-        updateAddressPort(port, PORT);
-        assertEquals(50, port.doubleIt(25));
-
-        ((java.io.Closeable)port).close();
-        bus.shutdown(true);
+        try {
+            if (!JavaUtils.isFIPSEnabled()) {
```
reta, line 250: @coheigea may I suggest please to have two independent test runs in Maven, non-FIPs as defaut:

```
<plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-surefire-plugin</artifactId>
                <executions>
                    <execution>
                        <id>default-test</id>
                        <configuration>
                            <systemPropertyVariables>
<org.apache.wss4j.crypto.jasypt.useLegacyDefaultAlgorithm>true</org.apache.wss4j.crypto.jasypt.useLegacyDefaultAlgorithm>
                            </systemPropertyVariables>
                        </configuration>
                    </execution>
</plugin>
```

And FIPS one:

```
<plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-surefire-plugin</artifactId>

                    <execution>
                        <id>fips-test</id>
                        <goals>
                            <goal>test</goal>
                        </goals>
                        <configuration>
                            <systemPropertyVariables>
                                <org.apache.wss4j.crypto.jasypt.useLegacyDefaultAlgorithm>false</org.apache.wss4j.crypto.jasypt.useLegacyDefaultAlgorithm>
                            </systemPropertyVariables>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
```

More than happy to help with that, it is very difficult to eliminate side effects of system properties in tests, thank you.

coheigea: @reta This is pretty complicated though for something I plan to remove once we pick up WSS4J 4.0.2 - I will convert the non-fIPS cases to the same algorithm that the FIPS case currently uses. The current code is just to tide us over so that I can compile the CXF main code with both versions of WSS4J for testing.

reta: Oh I see, thanks @coheigea , if that is temporary - non-issue

reta (approved):

PR #3463, opened 2026-09-13
Add sensible defaults to StaxUtils's MAX_ELEMENT_COUNT and MAX_XML_CHARACTERS

reta:
Add sensible defaults to StaxUtils's MAX_ELEMENT_COUNT and MAX_XML_CHARACTERS


core/src/main/java/org/apache/cxf/staxutils/StaxUtils.java
```diff
@@ -147,10 +147,10 @@ public final class StaxUtils {
             getInteger(MAX_TEXT_LENGTH, 128 * 1024 * 1024);  //128M - more than this should DEFINITELY use MTOM
     private static final int MIN_TEXT_SEGMENT_VAL =
             getInteger(MIN_TEXT_SEGMENT, 64); // Same default as woodstox
-    private static final long MAX_ELEMENT_COUNT_VAL =
-            getLong(MAX_ELEMENT_COUNT, Long.MAX_VALUE);
+    // The MAX_XML_CHARACTERS should be aligned with MAX_TEXT_LENGTH (maximum contiguous length of any character
```
reta, line 150: @coheigea it is difficult to come up with the right defaults here, but:
 - MAX_TEXT_LENGTH is set to 128Mb, so MAX_XML_CHARACTERS is set to at least 2x of that
 - MAX_ELEMENT_COUNT should be larger than MAX_CHILD_ELEMENTS (it is included), so setting it to be at least 10x of MAX_CHILD_ELEMENTS

coheigea:
@reta Some comments from Claude:

### 1. `maxElementCount` default overflows (blocking)

```java
getLong(MAX_ELEMENT_COUNT, Math.max(100 * MAX_CHILD_ELEMENTS_VAL, 50000));
```

`MAX_CHILD_ELEMENTS_VAL` is an `int`, so `100 * ...` is int arithmetic — the widening to `long` only happens after `Math.max(int,int)` has returned. It overflows once `maxChildElements > 21_474_836`:

| `maxChildElements` | resulting `MAX_ELEMENT_COUNT_VAL` |
|---|---|
| 50 000 (default) | 5 000 000 |
| 30 000 000 | 50 000 |
| `Integer.MAX_VALUE` | 50 000 (`100 * Integer.MAX_VALUE` wraps to `-100`, so the floor wins) |

That inverts the invariant the comment states — the total element count ends up well *below* the per-parent child count. It also hits precisely the users most likely to notice: `getInteger`/`getLong` coerce negatives back to the default, so there is no `-1` escape hatch and raising `maxChildElements` is the only way to relax that limit today. Anyone doing so would silently pick up a 50 000 total-element cap on upgrade.

One character:

```java
getLong(MAX_ELEMENT_COUNT, Math.max(100L * MAX_CHILD_ELEMENTS_VAL, 50000L));
```

### 2. Derive `maxXMLCharacters` from `MAX_TEXT_LENGTH_VAL`

Your rationale above is "2x of MAX_TEXT_LENGTH" and the new code comment says the two "should be aligned", but the code hardcodes 256Mb — so the alignment breaks as soon as anyone sets `org.apache.cxf.stax.maxTextLength`. Raise it to 512Mb and the document-wide character cap is *smaller* than a single permitted text segment: a contradictory config that worked fine before this change.

```java
private static final long MAX_XML_CHARS_VAL =
        getLong(MAX_XML_CHARACTERS, 2L * MAX_TEXT_LENGTH_VAL); // 256Mb by default
```

Same default (2 x 128Mb), stays correct under reconfiguration, and makes the comment true rather than aspirational. If you'd rather keep the literal, at least `256L * 1024 * 1024` so a future bump doesn't overflow.

reta:
Thanks @coheigea 

> 1. `maxElementCount` default overflows (blocking)

I will widen it to long, thanks!

> 2. Derive maxXMLCharacters from MAX_TEXT_LENGTH_VAL

The straight derivation does not make sense here to be fair. MAX_TEXT_LENGTH_VAL is really large by default, but we should not use multipliers there (ideally, I would think of MAX_TEXT_LENGTH_VAL as 8k and MAX_XML_CHARACTERS as 32Mb fe), so 256Mb is just reasonably large value that is not in conflict with MAX_TEXT_LENGTH_VAL default.

core/src/main/java/org/apache/cxf/staxutils/StaxUtils.java
```diff
@@ -161,6 +161,9 @@ public final class StaxUtils {
             getInteger(MAX_ELEMENT_DEPTH, getInteger(INNER_ELEMENT_LEVEL_SYSTEM_PROP, 100));
     private static final int MAX_CHILD_ELEMENTS_VAL =
             getInteger(MAX_CHILD_ELEMENTS, getInteger(INNER_ELEMENT_COUNT_SYSTEM_PROP, 50000));
+    // The MAX_ELEMENT_COUNT should be aligned with MAX_CHILD_ELEMENTS
+    private static final long MAX_ELEMENT_COUNT_VAL =
+            getLong(MAX_ELEMENT_COUNT, Math.max(100 * MAX_CHILD_ELEMENTS_VAL, 50000));
```
reta:
```suggestion
            getLong(MAX_ELEMENT_COUNT, Math.max(100 * MAX_CHILD_ELEMENTS_VAL, 50000L));
```

core/src/main/java/org/apache/cxf/staxutils/StaxUtils.java
```diff
@@ -161,6 +161,9 @@ public final class StaxUtils {
             getInteger(MAX_ELEMENT_DEPTH, getInteger(INNER_ELEMENT_LEVEL_SYSTEM_PROP, 100));
     private static final int MAX_CHILD_ELEMENTS_VAL =
             getInteger(MAX_CHILD_ELEMENTS, getInteger(INNER_ELEMENT_COUNT_SYSTEM_PROP, 50000));
+    // The MAX_ELEMENT_COUNT should be aligned with MAX_CHILD_ELEMENTS
+    private static final long MAX_ELEMENT_COUNT_VAL =
+            getLong(MAX_ELEMENT_COUNT, Math.max(100 * MAX_CHILD_ELEMENTS_VAL, 50000L));
```
coheigea: 50000L only widens the second argument, which picks the Math.max(long,long) overload. But 100 * MAX_CHILD_ELEMENTS_VAL is still int * int — it overflows and wraps before anything gets widened. The L needs to go on the 100 (as well)

reta:
```suggestion
            getLong(MAX_ELEMENT_COUNT, Math.max(100L * MAX_CHILD_ELEMENTS_VAL, 50000L));
```

coheigea (approved):

PR #3475, opened 2026-09-16
CXF-9245: Race Condition in ServerLifeCycleManagerImpl#stopServer()

reta:
Race Condition in ServerLifeCycleManagerImpl#stopServer()

```
java.lang.IndexOutOfBoundsException: Index: 3, Size: 2
 at java.base/java.util.concurrent.CopyOnWriteArrayList.listIterator(CopyOnWriteArrayList.java:1151)
 at org.apache.cxf.bus.managers.ServerLifeCycleManagerImpl.stopServer(ServerLifeCycleManagerImpl.java:66)
 at org.apache.cxf.endpoint.ServerImpl.stop(ServerImpl.java:177)
```

So two things at play here:
 - `listeners.listIterator` is safe, no need to synchronize (coming from Javadocs)
 - but because we iterate in reverse, there is a race between  `listeners.listIterator` and `listeners.size`
 - fixed that by cloning (== snapshoting)


core/src/main/java/org/apache/cxf/bus/managers/ServerLifeCycleManagerImpl.java
```diff
@@ -63,7 +63,10 @@ public void startServer(Server server) {
...
 
     public void stopServer(Server server) {
-        ListIterator<ServerLifeCycleListener> li = listeners.listIterator(listeners.size());
+        @SuppressWarnings("unchecked")
+        final CopyOnWriteArrayList<ServerLifeCycleListener> cloned = 
+            (CopyOnWriteArrayList<ServerLifeCycleListener>) listeners.clone();
+        ListIterator<ServerLifeCycleListener> li = cloned.listIterator(cloned.size());
         while (li.hasPrevious()) {
             li.previous().stopServer(server);
         }
```
coheigea, line 69: @reta Claude suggests this instead as it doesn't require the checkstyle suppresswarnings and doesn't constrain the field type:
```
final ServerLifeCycleListener[] snapshot = listeners.toArray(new ServerLifeCycleListener[0]);
for (int i = snapshot.length - 1; i >= 0; i--) {
    snapshot[i].stopServer(server);
}
```

coheigea (commented):
Is it also worth fixing the same pattern in:

https://github.com/apache/cxf/blob/1529dd69ab2eda78185e603c876ef2845463288b/core/src/main/java/org/apache/cxf/bus/managers/ClientLifeCycleManagerImpl.java#L66-L71
https://github.com/apache/cxf/blob/1529dd69ab2eda78185e603c876ef2845463288b/core/src/main/java/org/apache/cxf/bus/managers/CXFBusLifeCycleManager.java#L94-L97
https://github.com/apache/cxf/blob/1529dd69ab2eda78185e603c876ef2845463288b/core/src/main/java/org/apache/cxf/bus/managers/CXFBusLifeCycleManager.java#L107-L110

?

Finally: The synchronized on registerListener/unRegisterListener is decorative and always was: addIfAbsent/remove are already atomic, and neither startServer nor stopServer takes the lock

core/src/main/java/org/apache/cxf/bus/managers/ServerLifeCycleManagerImpl.java
reta: thanks @coheigea , yeah the clone is ugly, but there is an important difference: `clone` returns a shallow copy of this list.  (The elements themselves are not copied). With array, we actually allocate and copy all elements into a new array.

reta:
> Is it also worth fixing the same pattern in:

Addressed, thanks a lot @coheigea !

core/src/main/java/org/apache/cxf/bus/managers/ServerLifeCycleManagerImpl.java
```diff
@@ -63,7 +63,10 @@ public void startServer(Server server) {
     }
 
     public void stopServer(Server server) {
-        ListIterator<ServerLifeCycleListener> li = listeners.listIterator(listeners.size());
+        @SuppressWarnings("unchecked")
+        final CopyOnWriteArrayList<ServerLifeCycleListener> cloned = 
+            (CopyOnWriteArrayList<ServerLifeCycleListener>) listeners.clone();
+        ListIterator<ServerLifeCycleListener> li = cloned.listIterator(cloned.size());
```
coheigea: @reta What do you think about putting this into a common place like [CollectionUtils](https://github.com/apache/cxf/blob/main/core/src/main/java/org/apache/cxf/common/util/CollectionUtils.java) to avoid duplicating it?

coheigea (approved):

core/src/main/java/org/apache/cxf/bus/managers/ServerLifeCycleManagerImpl.java
reta: > @reta What do you think about putting this into a common place like [CollectionUtils](https://github.com/apache/cxf/blob/main/core/src/main/java/org/apache/cxf/common/util/CollectionUtils.java) to avoid duplicating it?

Sure @coheigea , I think we could keep it closer to XxxLifecycleManagers, let me push a change shortly, thanks!

core/src/main/java/org/apache/cxf/bus/managers/AbstractLifeCycleManager.java
```diff
@@ -0,0 +1,42 @@
...
+package org.apache.cxf.bus.managers;
+
+import java.util.ListIterator;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+/**
+ * AbstractLifeCycleManager implementation
+ * @param <T> listener class
+ */
+abstract class AbstractLifeCycleManager<T> {
```
reta, line 29: @coheigea how about that? I didn't push it too far, the register/unregsiter is very different between impl classes, so I just did the iteration part

core/src/main/java/org/apache/cxf/bus/managers/ServerLifeCycleManagerImpl.java
coheigea: LGTM thanks

